PR #1162, opened 2026-04-16
Filter OCR metadata from within-book search results

JDerekLomas:
## Summary
- Within-book search was returning pages where the only match was in OCR metadata tags (`<warning>`, `<image-desc>`, `<meta>`, `<page-type>`, etc.) — scan descriptions, not book content
- Now strips these tags before matching, drastically reducing noise
- Also gates semantic results: only adds them when keyword found <10 pages (avoids redundancy when keyword already has good coverage)

## Before/after
"can humans create life" in *Three Books on Gold*:
- Before: 55 pages (mostly blank pages and scan descriptions)
- After: only pages with actual content matches + semantic conceptual matches

## Test plan
- [ ] Search for common words ("life", "gold") — verify blank/metadata pages are filtered out
- [ ] Search for rare phrases — verify semantic results appear when keyword has few hits
- [ ] Verify no real content is accidentally stripped

### DIFF
--- a/src/app/api/books/[id]/search/route.ts
+++ b/src/app/api/books/[id]/search/route.ts
@@ -1,6 +1,7 @@
 import { NextRequest, NextResponse } from 'next/server';
 import { getDb } from '@/lib/mongodb';
 import { buildPageSearchStage } from '@/lib/atlas-search';
+import { semanticPageSearchScoped } from '@/lib/semantic-search';
 
 interface SearchMatch {
   field: 'ocr' | 'translation';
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Superseded: main's within-book search now routes all page text through the canonical stripEditorialWrappers() (PRs #2232/#2233 family) and merges parallel semantic results via semanticPageSearchScoped — both halves of this PR are implemented in better form. (The homegrown regex here also stripped <note> content, which we now deliberately keep per the quote-integrity rules.)

PR #1813, opened 2026-05-16
MCP: signal truncation and document pagination in get_book_text

### LINKED ISSUE #1772: Enhancement: signal /text truncation and add per-page fallback in get_book_text (happy to PR)
LS; Team,

This is a small enhancement request, not an issue per se, from someone who has been using Source Library recently. For context, I'm a doctoral researcher at Technion + Cornell (Architecture & AI) building a narrative-based occupant-behavior simulator and protocol, and I've been processing renaissance Hermetic + Architectural primary sources from your collection to anchor a 700-year design-genealogy argument (llull → bruno → maier). About a dozen books deep over the last two weeks via the REST API and the MCP, and i'd like to give a bit back. I have used the Ritman library's resources before and the  previous digitalised collection has been core of my curiosity.

Before the request: a few things that work exceptionally well and that I've been borrowing into my own project's open-sourcing roadmap.

- `/llms.txt` is the cleanest AI-consumer-facing developer doc i've seen anywhere
- DTS v1.0 + IIIF Search 2.0 implementation: standards-first instead of inventing custom surfaces is rare and right
- Per-page cost transparency on `/developers/pipeline` is a posture most teams don't take, and it earns trust faster than any pitch could
- syncing the OCR + translations to a public github repo is a much stronger commitment to legibility than open-sourcing only the code

So in short: thank to everyone for building this in the open. Its load-bearing for what i'm doing.

## The User Experience gap

Processing whole books via `/api/books/{id}/text` for unauthenticated callers, the endpoint returns 200 OK with `pages_returned: 0` after roughly the first 50–57 pages. there's no `truncated` flag, no `429`, no `Retry-After`, nothing in the response that distinguishes "You've hit a limit" from "This book has no more content." the `get_book_text` MCP tool (advertised in `llms.txt` as "Read 50+ pages in one call") wraps the same endpoint and so silently truncates at the same threshold.

An AI agent reading the response will reasonably infer "no content on these pages" rather than "you hit an undocumented cap," because that's what the response shape says. in two of my three deep-reads (bruno's *de umbris idearum* and maier's *atalanta fugiens*) I nearly produced analyses based only on the first 50 pages of 251 / 230 without noticing. I caught it because i was double-checking citation ranges, not because the API told me anything was wrong.

Example reproduction (tested may 2026, all three books in the lineage i'm researching):

```bash
# bruno, de umbris idearum (251 pages, 247 translated)
curl -s "https://sourcelibrary.org/api/books/699050d07d19f3f2aac1e6e4/text?from=1&to=50"
# 200 OK, full content

curl -s "https://sourcelibrary.org/api/books/699050d07d19f3f2aac1e6e4/text?from=51&to=100"
# 200 OK, "pages_returned": 0

# llull, ars magna generalis et ultima (lavinheta 1517, 286 pages)
# same pattern, accessible to ~p. 57, then pages_returned: 0

# maier, atalanta fugiens (230 pages)
# same pattern, accessible to ~p. 46, then 0
```

Worth flagging: the cutoff varies slightly per book (57 / 50 / 46), so this might be a rate-limit shape rather than a strict page count. either way, the silent-failure mode is the load-bearing issue from the consumer side.

## Suggested enhancement

Three small things, smallest first, all of which would make the anonymous tier much friendlier for AI consumers without changing the limit itself:

1. **return a `truncated: true` field plus a `reason` string** on the `/text` response when content is suppressed for any reason other than end-of-book. One extra field, no breaking change, gives consumers something to detect and switch on.

2. **Document the limit in `llms.txt`** under the `/text` section. roughly:

   > The `/text` endpoint is rate-limited for anonymous callers (~50 pages per book per session). for full-book reads, use the MCP server's `get_book_text` tool with the per-page fallback (see below), the DTS `/document` endpoint with a page range, or request an API key for higher limits.

3. **Fall back to per-page `/quote` calls in the `get_book_text` MCP tool** when `/text` returns truncation mid-book. so the MCP tool actually delivers what its description in `llms.txt` promises, without changing the underlying REST behaviour.

## What i'd happily PR

In this priority order, on whatever scope you'd prefer:

1. Docs PR to `llms.txt` — adds the truncation note for `/text`, plus the `--transport http` flag in the remote-MCP `claude mcp add` example (i lost ~30 min on that one too because the CLI defaults to stdio for the URL form, and the registered server appears connected at the CLI but never loads tools).
2. MCP server fallback in `get_book_text` — per-page `/quote` when `/text` returns truncation, with clear logging. would scope against your preferred error-handling pattern before writing any code.
3. Data-only PR with the catalogue metadata corrections noted below (lavinheta date, pal.lat.984 date, plus any others i hit through the ongoing reading project), if metadata lives in a versionable file in the repo.

If any of these are something you'd rather lead yourselves let me know, otherwise i'll send the docs PR. Happy to chat about the MCP one before writing.

## What i'm not asking for

Just to scope this politely: I'm not proposing changes to the pipeline (OCR, translation, image extraction), to the licensing model (which is exemplary), or to the IIIF / DTS / MCP transport layers (those are correctly implemented standard-compliant surfaces). This issue-late-night-enjoyment-back-to-open-source is purely about the AI-consumer-facing experience on the existing endpoints.

## Smaller related observations

Leaving these as a list rather than separate issues so you can decide whether you want them split into their own threads; happy to file individually on request.

- intermittent `403` on `/api/books/{id}/quote?page=N` for the same book where `/text` returns `200`
- `/api/search/images?book_id=…` returns `404` for image-rich books (e.g. atalanta fugiens, which carries 50 documented copperplate emblems as its primary feature)
- catalogue metadata: lavinheta *ars magna* (id `695592c17bd6d2cd1d61afec`) listed as 1523, but the colophon on p. 279 unambiguously gives 11 may 1517. pal.lat.984 (id `6953ab9577f38f6761bd786e`) listed as "1200" but the binding is inscribed 1725 and the hand is late 14th-c. cursive, so the date can't be earlier than late 14th c. happy to send a data-only PR with corrections if metadata lives in a versionable file.
- DTS `/document` endpoint isn't surfaced as the recommended path for bulk reads in `llms.txt`. one extra sentence under the `/text` section would help.
- "generous limits" in `llms.txt` instead of concrete rate-limit numbers, which combined with the silent-truncation behaviour above makes it impossible to distinguish rate-limit from page-cap from end-of-book.
- consider adding `companion_works` and `parent_collection` fields to book metadata. the 1651 strasbourg `raymundi lullii opera` printing llull bound *with* bruno commentaries is currently invisible in the catalogue, and it's the load-bearing primary evidence for the bruno-llull lineage in my work — without that linkage you can find each book separately but not the relationship.
- citation workflow gap: the `has_doi` filter is great, but most pre-modern texts in the catalogue currently don't have one because the DOI is awarded to the *Scholarly Edition*, not the *book*. for a researcher trying to identify citable editions, a `scholarly_edition_doi` field at the book level (resolving to the latest published edition) would be a meaningful improvement to citation flow. roadmap suggestion, not a bug.

Let me know if any of these should be split into their own issues, and which (if any) of the PRs above you'd like me to start on. Again, hartelijk dank for building source library in the open. The fact that i can read primary llullian text from a 1517 lyon print at 2026 from a CLI is remarkable and I (and my team of people, and agents) don't take it lightly, incredible time to be alive.

Daan van der Zwaag
_PhD Researcher at Technion, Faculty of Architecture and Town Planning, Haifa_

JDerekLomas:
## Summary

- Adds `truncated: true/false` to all `get_book_text` JSON responses (page-range mode) so LLMs can detect a page-budget cut rather than inferring end-of-book from `pages_returned: 0`
- When `truncated=true`, a `truncation_note` field provides the exact `from`/`to` values for the next call, plus total page count — designed to be machine-readable by LLM agents
- Updates the tool description in both the HTTP MCP server (`src/app/api/mcp/route.ts`) and the stdio MCP server (`mcp-server/src/`) to document the truncation signal, recommended 50-page chunk workflow, and budget limits for anonymous callers

## Background

Reported in #1772 by Daan van der Zwaag (PhD researcher, Technion + Cornell Faculty of Architecture and Town Planning), who caught mid-book silent truncation on Bruno's *De Umbris Idearum* (251 pp) and Maier's *Atalanta Fugiens* (230 pp) only because he was cross-checking citation ranges — a failure mode that would fool most AI agents without an explicit signal. His diagnosis: `pages_returned: 0` after page 50 is indistinguishable from "no content on these pages" without a `truncated` flag.

## What's not changed

- The underlying page budget limits are unchanged
- Chapter mode is unchanged (chapters already have clear structure)
- No breaking changes to the response shape — `truncated` and `truncation_note` are additive fields

## Test plan

- [ ] Call `get_book_text` with no `from`/`to` on a book with >50 pages as an anonymous caller — verify `truncated: true` and `truncation_note` appears with correct `from`/`to` values
- [ ] Call with explicit `from=1 to=50` on a 251-page book — verify `truncated: true`, `truncation_note` says `from=51 to=100`
- [ ] Call with `from=1 to=50` on a 30-page book — verify `truncated: false`, no `truncation_note`
- [ ] Call with `chapter=0` — verify no `truncated` field added (chapter mode unchanged)
- [ ] Verify stdio MCP server (`mcp-server/`) has same behavior

Closes #1772

### DIFF
--- a/mcp-server/src/api.ts
+++ b/mcp-server/src/api.ts
@@ -374,6 +374,28 @@ export async function getBookText(args: {
     }
   }
 
+  // Detect truncation: signal clearly when pages_returned < total_pages so LLMs
+  // don't infer "the book ends here" from an incomplete response.
+  const totalPages = Number(result.total_pages || 0);
+  const pagesReturned = Number(result.pages_returned || 0);
+  // Chapter mode returns different shape — only add truncation signal for page-range mode
+  if (args.chapter === undefined && totalPages > 0) {
+    const fromPage = args.from !== undefined ? args.from : 1;
+    const expectedUpTo = args.to !== undefined ? args.to : totalPages;
+    const lastPageReturned = pagesReturned > 0 ? fromPage + pagesReturned - 1 : fromPage - 1;
+    const isTruncated = lastPageReturned < expectedUpTo;
+    (result as Record<string, unknown>).truncated = isTruncated;
+    if (isTruncated) {
+      const nextFrom = lastPageReturned + 1;
+      const nextTo = Math.min(nextFrom + 49, totalPages);
+      (result as Record<string, unknown>).truncation_note =
+        `TRUNCATED — received ${pagesReturned} pages (up to p.${lastPageReturned}) ` +
+        `but requested up to p.${expectedUpTo} of ${totalPages} total. ` +
+        `This is NOT end-of-book — call get_book_text again with from=${nextFrom} to=${nextTo} ` +
+        `to read the next chunk. Repeat until you reach total_pages (${totalPages}).`;
+    }
+  }
+
   // Add quoting guidance when returning page text
   (result as Record<string, unknown>).tip =
     "When quoting from these pages, copy text verbatim from the translation field. Do not paraphrase or reconstruct from memory.";
--- a/mcp-server/src/index.ts
+++ b/mcp-server/src/index.ts
@@ -217,7 +217,7 @@ const TOOLS: Tool[] = [
   {
     name: "get_book_text",
     description:
-      "READ A BOOK — start here. Preferred: use 'chapter' param to read one chapter at a time (includes page markers like [Page 42] for citation). Or use page ranges (from/to) for focused reading. Call get_book first to see the chapter list.",
+      "READ A BOOK — start here. Preferred: use 'chapter' param to read one chapter at a time (includes page markers like [Page 42] for citation). Or use page ranges (from/to) for focused reading. Call get_book first to see the chapter list. TRUNCATION: the response always includes truncated: true/false. When truncated=true, the truncation_note field gives the exact next from/to values to call — this means content was cut short by a page-budget limit, NOT that the book ended. An AI agent MUST NOT infer end-of-book from pages_returned alone; always check truncated first. Budget limits apply to anonymous callers (~50 pages per 24h); sign in at sourcelibrary.org/auth/signin or get an API key at sourcelibrary.org/developers for higher limits.",
     inputSchema: {
       type: "object" as const,
       properties: {
@@ -240,11 +240,11 @@ const TOOLS: Tool[] = [
         },
         from: {
           type: "number",
-          description: "Start page number (inclusive). Use chapter param instead when possible.",
+          description: "Start page number (inclusive). Use with to for explicit page ranges. If the response has truncated=true, use the next from/to from truncation_note.",
         },
         to: {
           type: "number",
-          description: "End page number (inclusive). Use chapter param instead when possible.",
+          description: "End page number (inclusive). Recommended chunk size: 50 pages. If the response has truncated=true, use the next from/to from truncation_note.",
         },
         format: {
           type: "string",
--- a/src/app/api/mcp/route.ts
+++ b/src/app/api/mcp/route.ts
@@ -220,7 +220,30 @@ async function getBookText(args: Record<string, unknown>) {
   const slug = book?.slug || book?.id || args.book_id;
   const pages = result.pages as Array<Record<string, unknown>> | undefined;
   if (pages) for (const p of pages) p.url = `https://sourcelibrary.org/book/${slug}?page=${p.page_number}`;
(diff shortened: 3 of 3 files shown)


danoszz:
Thanks for the work!

Issue #1921, opened 2026-05-21
BPH catalog editor: expose full field set, fix card display, add author thesaurus (Paul feedback)


JDerekLomas:
## Context

Paul Dijstelberge logged into `bph.sourcelibrary.org/login` for the first time on **2026-05-21** and gave a first round of catalog feedback (more notes will follow). A related screenshot note arrived **2026-05-18** about collation formula rendering. This issue collates both into actionable work.

Source emails:
- *Re: Welcome to the inner circle — please sign in again* — 2026-05-21, p.dijstelberge@me.com
- *what I noticed* — 2026-05-18, pdijstelberge@efm.amsterdam

### 1. Login / sign-out
- **Paul says:** "I did not succeed in signing out and in."
- **Code says:** Sign-out is wired (`src/components/layout/UserMenu.tsx:173`, NextAuth + JWT, magic-link + Google). Works in our testing.
- **Likely cause:** UX, not code — the sign-out control isn't where Paul expects, or he was looking at the wrong account menu on the BPH subdomain.
- **Action:** Verify sign-out is visible on the BPH tenant subdomain (not just sourcelibrary.org). Make the user menu more obvious for first-time librarian logins.

### 2. Editor scope — \"only keywords pop up\"
- **Paul says:** Only keywords were editable.
- **Code says:** `BookEditModal` (`src/components/book/BookEditModal.tsx:41`) exposes 8 fields: title, display_title, author, language, published (year), place_published, publisher, ustc_id. Keywords render in `BibliographicInfo` but are display-only there. The Edit button is gated to `role=\"inner_circle\"` (`src/components/book/BibliographicInfo.tsx:194`) — library-scoped editors don't see it. None of: format/collation, binding, notes, shelfmark, original_language, collection — are editable.
- **Root cause:** BPH-rich fields live in **Supabase `bph_works`** (binding, shelfmark, bibliographic_format, variant_author, etc.), while the editor only writes Atlas `books`. The two stores are not unified.
- **Action:** Decide between (a) sync BPH fields into Atlas and extend `BookEditModal`, or (b) add a parallel \"BPH cataloguer\" panel that writes Supabase `bph_works`. Recommend (b) — keeps Atlas lean, gives Paul a tool that matches his mental model.

### 3. Default filter
- **Paul says:** Default is \"show digitised\"; should be \"show all\" + add \"show books actually in the library.\"
[...]

PR #1925, opened 2026-05-22
BPH catalog quick wins: card impressum, sign-out chrome, filter labels (#1921 P1)

JDerekLomas:
## Summary

Three of the four P1 quick wins from #1921 (Paul Dijstelberge's first-login feedback). Each is independent; bundled because they're all small UI fixes Paul will see immediately.

### 1. Grid card layout (`BphCatalogBrowser.tsx`)
Card now follows librarian convention: **author** → **title** → **impressum** (place, publisher, year). Place and publisher were missing from the card entirely. Falls back to printer when publisher is null (BPH data has many printer-only rows pre-1700).

### 2. Digitization filter labels
- `"All"` → `"All books"`
- `"On Source Library"` → `"Digitised on Source Library"`

Paul read \"All\" as filter-cleared rather than filter-set-to-all and \"On Source Library\" as ambiguous between \"hosted by SL\" vs \"present in SL's index\".

### 3. New `TenantUserChrome` floating widget
**Root cause for \"I did not succeed in signing out\":** the standard `SiteHeader` (with `UserMenu`) is hidden by CSS in embed mode (`body:has(.embed-mode) > header` in `globals.css`). Tenant subdomains rewrite to `/embed/[tenant]/*`, so on bph.sourcelibrary.org there is **no user menu rendered anywhere** — no avatar, no sign-out, no profile link. Paul could log in but had no way to log out without leaving the subdomain.

`TenantUserChrome` is a minimal floating widget (top-right, fixed) that:
- Renders **null** for anonymous visitors (public browsing stays uncluttered)
- Renders avatar + name + Sign out for authenticated users
- Intentionally omits `/account`, `/favorites`, admin links — these resolve off-subdomain via the `proxy.ts` rewrite and would violate the tenant subdomain lockdown rules from CLAUDE.md
- `signOut({ callbackUrl: '/' })` keeps the user on the same host post-logout

Mounted in `src/app/embed/[tenant]/layout.tsx` so it appears on every tenant page (homepage, catalog browse, catalog detail, book detail, search, gallery).

## Not included from #1921 P1

**Collation-formula asterisk stripping** — Paul cited *Polityke toet-steen* (UBN 11473), but that row has no `bibliographic_format` value in Supabase, so the bug isn't reproducible from that example. Need a specific UBN where the source has `*` characters that don't appear in display. Will follow up in a separate PR once Paul provides one.

## Test plan

- [ ] Anonymous on https://bph.sourcelibrary.org — no user chrome visible, no behavior change
- [ ] Signed in as Paul/Jose on bph.sourcelibrary.org — avatar + name visible top-right; click reveals dropdown; \"Sign out\" returns to bph.sourcelibrary.org root
- [ ] Grid view on /catalog — cards now show author / title / \"Place, Publisher, Year\" lines
- [ ] Filter dropdown labels read as expected
- [ ] `node scripts/audit-bph-leaks.mjs` still passes (no off-subdomain links from the new chrome)

### DIFF
--- a/src/app/embed/[tenant]/layout.tsx
+++ b/src/app/embed/[tenant]/layout.tsx
@@ -1,4 +1,5 @@
 import { TenantSessionUpdater } from '@/components/auth/TenantSessionUpdater';
+import TenantUserChrome from '@/components/tenant/TenantUserChrome';
 
 /**
  * Layout for /embed/[tenant]/* — the route group that tenant subdomains
@@ -8,6 +9,10 @@ import { TenantSessionUpdater } from '@/components/auth/TenantSessionUpdater';
  * the user's tenant-scoped role (e.g. BPH editor) is resolved into their
  * session. Without this, client gates like <AuthCheck role="inner_circle">
  * never see the membership and editor-only UI stays hidden.
+ *
+ * TenantUserChrome is the only sign-out affordance on tenant subdomains —
+ * the standard SiteHeader is hidden in embed mode. It renders null for
+ * anonymous visitors so public browsing stays uncluttered.
  */
 export default async function EmbedTenantLayout({
   children,
@@ -20,6 +25,7 @@ export default async function EmbedTenantLayout({
   return (
     <>
       <TenantSessionUpdater tenantSlug={tenant} />
+      <TenantUserChrome />
       {children}
     </>
   );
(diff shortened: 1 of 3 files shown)


JDerekLomas:
Closing as duplicate of #1932 (the other dev shipped equivalent #1921 P1 work). Will remove this branch's unique TenantUserChrome implementation; main now has EmbedUserMenu which covers the same ground.

PR #2058, opened 2026-05-26
seo: experimental segment not-found.tsx for /book/[id] — testing 404 status

JDerekLomas:
## Hypothesis under test
With \`revalidate=86400\` (ISR) on \`/book/[id]\`, \`notFound()\` thrown from the page body returns HTTP **200** in production — verified on \`https://sourcelibrary.org/book/<bad-slug>\` with brand-new URLs. The not-found UI renders (\`<template data-dgst="NEXT_HTTP_ERROR_FALLBACK;404">\` is in the body), but the response status commits as 200 and Vercel caches it for 24h via the ISR cache-control header.

A route-segment \`not-found.tsx\` is the App Router escape hatch documented for exactly this situation: when notFound() can fire inside a streaming boundary, the segment-local not-found file lets Next.js preserve the 404 status because it knows the boundary statically.

## Change
Add \`src/app/[tenant]/book/[id]/not-found.tsx\` rendering the existing \`NotFoundContent\` component. Smoke-test on the preview:

- \`curl -o /dev/null -w '%{http_code}\\n' <preview>/book/never-seen-before-$(date +%s)\` should return **404** (today: 200).

If the hypothesis holds, follow-up PR replicates the pattern in:
- \`src/app/[tenant]/book/[id]/page/[pageId]/\`
- \`src/app/[tenant]/topics/[slug]/\` and \`src/app/topics/[slug]/\`
- \`src/app/[tenant]/languages/[code]/\`

If it doesn't, fallback is \`dynamic = 'force-dynamic'\` on those routes (loses ISR but guarantees correct status).

### DIFF
--- a/src/app/[tenant]/book/[id]/not-found.tsx
+++ b/src/app/[tenant]/book/[id]/not-found.tsx
@@ -1,0 +1,18 @@
+import { Suspense } from 'react';
+import NotFoundContent from '@/components/layout/NotFoundContent';
+
+// Route-segment not-found.tsx for /book/[id]. When notFound() is called from
+// page.tsx or any of its children (BookInfo Suspense child included),
+// Next.js renders this UI AND sets HTTP status 404. The global
+// src/app/not-found.tsx renders the same body but, with ISR enabled at the
+// page level (revalidate = 86400), the status sometimes commits as 200
+// because streaming has already started — Vercel then caches that 200 for
+// 24h. A segment-local not-found.tsx is the documented App Router escape
+// hatch for that.
+export default function BookNotFound() {
+  return (
+    <Suspense>
+      <NotFoundContent />
+    </Suspense>
+  );
+}


JDerekLomas:
Closing as won't-fix for now. Findings for future revisit:

**Diagnosis.** `notFound()` thrown from the page body IS being called — the not-found UI renders (`<template data-dgst="NEXT_HTTP_ERROR_FALLBACK;404">` is in the response body, and `NotFoundContent`'s `min-h-screen bg-cream` wrapper renders, not the page's own wrapper). But the HTTP status commits as **200**, and Vercel caches that 200 for 24h via the ISR `s-maxage=86400` header.

**What I tried.** Added a route-segment `src/app/[tenant]/book/[id]/not-found.tsx`, hypothesizing that the App Router would use the segment-local boundary to preserve 404 status (per Next.js docs). Verified on the preview with a brand-new bad slug: still 200. Hypothesis falsified.

**Remaining options (all have real costs):**
1. `export const dynamic = 'force-dynamic'` on `/book/[id]`, `/topics/[slug]`, `/languages/[code]`, `/book/[id]/page/[pageId]`. Returns proper 404s but loses the 24h ISR HTML cache on every real book page (+200-500ms per request).
2. Middleware-level existence check in `src/proxy.ts` returning `new NextResponse(htmlBody, { status: 404 })`. Keeps ISR for real books but `NextResponse.rewrite` can't change status, so the response body has to be inlined HTML (degraded UX).
3. Wait for Next.js to fix the interaction between `notFound()` and ISR in a future release.

**Why deferring is OK.** The metadata already sets `robots: { index: false, follow: false }` on these branches (after #2046), so Google won't *index* the soft-404s. The only remaining cost is wasted crawl budget — a few percent at most. Re-evaluate if GSC's "Soft 404" report starts piling up materially.

**Related context.** `src/app/[tenant]/book/[id]/loading.tsx` exists and adds an implicit Suspense around the page, which may be relevant to why `notFound()` is committing the response before throwing. The `revalidate = 86400` export at the route level is the other ingredient.

PR #1916, opened 2026-05-21
proxy: demote source-provider 'tenants' to kind: 'source' (#1905, phase 1)

JDerekLomas:
## Summary

Phase 1 of #1905: add a `kind` discriminator to tenant rows and stop the proxy from resolving source-provider slugs (internet-archive, gallica, bodleian, …) as real tenants. Real tenants (`bph`, `kloss-collection`) and the route-scaffolding `default` row are untouched.

This PR ships the **mechanism** (code change) but leaves the **toggle** off — the migration script that tags source-provider rows with `kind: 'source'` is included but not run. Merging changes nothing observable until the script runs with `--apply`.

## What changes

**`src/proxy.ts`**
- `resolveActiveTenant()`, `resolveTenantSlugById()` now filter `kind: { $ne: 'source' }`. Rows without `kind` continue to match — backward-compatible.
- New `resolveSourceProvider()` + redirect block (placed just before the bot-enforcement section). Any `/{provider}/*` URL 301-redirects to the global equivalent:
  - `/{slug}/gallery[/...]` → `/gallery?library={slug}`
  - `/{slug}/browse[/...]` → `/browse?library={slug}`
  - `/{slug}/book/{id}` → `/book/{id}`
  - `/{slug}/collections[/...]` → `/collections[/...]`
  - `/{slug}` and anything else → `/libraries/{slug}`

**`src/lib/tenant-context.ts`** — `resolveTenantId()` also filters `kind: { $ne: 'source' }` for API-route defense in depth.

**`scripts/maintenance/tag-source-provider-tenants.mjs`** — dry-run by default. Lists 29 hardcoded source-provider slugs; refuses to touch `bph`, `kloss-collection`, `default` even if the list is tampered with. `--apply` sets `kind: 'source'`.

**`scripts/maintenance/measure-source-provider-scope.mjs`** — read-only. One `$group` per collection. Surfaces the real blast radius for the follow-up data migration.

## Measured scope (production DB)

| Collection | Total rows pointing to source-provider tenantIds |
|------------|--------------------------------------------------|
| books | 11,085 |
| gallery_images | 1,628 |
| collections | 151 |
| library_catalog_records | 0 |

Top providers by data: `internet-archive` (6,759 books / 734 gallery_images / 146 collections), `bavarian-state-library` (1,423 books), `e-rara` (1,094 books). 10 of the 29 source-provider rows have zero data and can be deleted outright in the follow-up.

[...]

PR #2126, opened 2026-05-28
Overhaul tenant architecture.

Mayank-PPL:
Shifted from tenant scoped routes to majorly global routes and only some of the tenant scoped routes.
Overhauled the directory structure & architecture.
This would solve all the band-aid fixes we needed to deal with non-tenant partner routes.


JDerekLomas:
Heads-up — follow-up cleanup available now that your architecture overhaul has landed.

In #2085 (yesterday) I cleaned up the data side of the source-provider tenants: ~11,054 books, 4M pages, 100K gallery items, 151 collections had `tenantId` unset, plus the legacy snake-case `tenant_id` field swept from ~6.4M docs. After the sweep, only the 3 real subdomain tenants (bph, kloss-collection, bhutan) have any docs pointing at them.

So the next step from #1916's "out of scope" list is unblocked: **delete the 29 source-provider tenant rows + the legacy `default` row from the `tenants` collection**. They're orphan metadata at this point — nothing references them.

I held off deleting them yesterday because I wanted you to decide the disposition (hard delete vs. `status: 'deleted'` tombstone vs. keep with `kind: 'source'`), since your architecture work may want them around for some audit/attribution path I don't have context on.

Verification queries if useful:
```
db.books.distinct('tenantId', {})
// → only [bph-uuid, kloss-uuid, bhutan-uuid]

db.tenants.find({ kind: { $ne: 'subdomain' } }).count()
// → 30 (29 provider + 1 default)
```

Let me know if you'd like me to do the delete pass, or if you've already folded it into your work.

Mayank-PPL:
I've not deleted or modified any data.
So, you can proceed with it.

Mayank-PPL:
I've deployed new changes. With that in place we can go ahead and hard delete 29 source-provider tenant rows from `tenants` collections.
With that we will be left with 3, and if we want we can even remove the field `kind` as well from the collections.
It is not used in runtime as such anymore.

PR #2136, opened 2026-05-28
feat(embeddings): tenant-aware Supabase RPCs + explicit CLIP degradation

JDerekLomas:
## Summary

**One concern: tenant isolation in semantic/visual retrieval.** Today the Librarian and gallery routes post-filter tenant scope in JS after over-fetching from Supabase. This starves results when tenant-tagged content dominates top-N. The Bhutan/BPH leak handoff is the bug this prevents. Plus a CLIP availability fix that's been bothering me — silent degradation made image search unreliable without anyone noticing.

### Tenant filtering at SQL

- Added \`tenant_id TEXT\` to all 5 embedding tables (NULL = main-site).
- Updated all 7 \`match_*\` RPCs to accept \`filter_tenant_id\` with strict semantics:
  - \`NULL\` → only \`tenant_id IS NULL\` rows (main-site)
  - \`'<uuid>'\` → only that tenant's rows
- Parameterized \`src/lib/semantic-search.ts\` end-to-end (every exported fn now takes \`tenantId?\`).
- \`src/lib/embassy/librarian.ts\` passes \`tenantId: null\` explicitly, with a comment marking the spot for future per-tenant Librarian instances. Dropped the 24→10 over-fetch since the RPC pre-filters in SQL.
- Gallery + visual + identify + unified search routes propagate resolved tenant context to the RPC layer.

### CLIP degradation

- Brief HEAD probe on \`/health\` (60s cached).
- If unhealthy: \`executeSearchImages\` returns \`{ images: [], degraded: 'clip_unavailable' }\`; caller surfaces "Image search is temporarily unavailable" in the Librarian's response stream instead of falling back silently to a misleading regex.

## Out of scope

- **Atlas Search tenant indexing.** The \`executeSearchCollection\` keyword path keeps its existing JS post-filter because the Atlas Search index has no tenant field. Adding it is a Mongo Atlas config change tracked separately.
- **Embedding store hygiene** — PR B (\`embedding_model\`, \`mongo_updated_at\`) is a sibling PR.
- **Retrieval changes** — PR A handles hybrid + rerank.

## Behavior change to flag

Callers that previously omitted \`filter_tenant_id\` got cross-tenant results. They now get main-site only (the desired default). All identified callers in this repo have been updated to pass explicit scope.

## Apply order

\`\`\`bash
psql \"\$SUPABASE_DB_URL\" -f scripts/migration/add-tenant-id-to-embeddings.sql
psql \"\$SUPABASE_DB_URL\" -f scripts/migration/update-match-rpcs-tenant-aware.sql
set -a; source .env.production.local; set +a
node scripts/migration/backfill-embedding-tenant-id.mjs --dry-run
node scripts/migration/backfill-embedding-tenant-id.mjs --apply
# then deploy code changes
\`\`\`

## Test plan

- [ ] Apply schema + RPC migrations (idempotent; can re-run safely)
- [ ] Dry-run + apply tenant_id backfill, verify counts (should be ~16K tenant-tagged across BPH/Bhutan/Kloss, rest NULL)
- [ ] \`node scripts/audit-bph-leaks.mjs\` still passes (no main-site→BPH leaks)
- [ ] Spot-check: query Librarian with terms matching BPH content; confirm main-site response excludes those books at the result level (not just visibility)
- [ ] Visit \`bph.sourcelibrary.org/gallery\` and main-site \`/gallery\` with the same query — confirm scope separation
- [ ] Bring down CLIP VM (or block the URL) — Librarian image search now says "temporarily unavailable" instead of returning regex matches

### DIFF
--- a/scripts/migration/add-tenant-id-to-embeddings.sql
+++ b/scripts/migration/add-tenant-id-to-embeddings.sql
@@ -1,0 +1,82 @@
+-- Add tenant_id to all five Supabase embedding tables.
+--
+-- Why: today the Librarian post-filters in JS (over-fetch from Supabase,
+-- look up books in Mongo with tenantId:{$in:[null,undefined]}, drop the rest).
+-- This starves results when tenant-tagged books dominate top-N — a query
+-- mapping strongly to BPH content can have its entire top-24 tenant-tagged,
+-- and the main-site user sees nothing even though relevant non-tenant books
+-- exist at rank 25+. The proper fix is tenant filtering at the SQL layer.
+--
+-- Source of truth: books.tenantId in Mongo (NULL = main-site).
+-- We DENORMALIZE that value onto each embedding row at write time. The
+-- "drift" risk (book is re-tenanted but embeddings stay on old value) is
+-- handled by the existing scripts/maintenance/lesson_tenant_promotion_backfill
+-- pattern — promotions explicitly backfill all derived stores.
+--
+-- Semantics for the RPC filter (see update-match-rpcs-tenant-aware.sql):
+--   filter_tenant_id IS NULL   → return ONLY rows where tenant_id IS NULL
+--                                (= main-site only; mirrors the existing
+--                                 Mongo convention tenantId:{$in:[null,undefined]})
+--   filter_tenant_id = '<uuid>' → return ONLY rows where tenant_id = '<uuid>'
+--                                 (= specific tenant only; for future per-tenant
+--                                  Librarian instances)
+--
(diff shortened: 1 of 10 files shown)


JDerekLomas:
Closing as **partly superseded** — page search (`match_semantic`) is already tenant-scoped inline on main (`src/lib/semantic-search.ts:276`). The residual work (scope the other 4 RPCs: books, artworks, gallery_text, clip) + reuse of this branch's migration SQL/backfill is captured in #2753. Branch `tenant-clip` preserved for the SQL reference.

PR #2142, opened 2026-05-28
fix(bph): stable React key for null-ubn catalog rows

JDerekLomas:
## Summary

- BPH catalog search "Helicone" returned `2 of 29,876 works` but rendered 4 rows — 2 Helicone matches plus 2 unrelated Chymische Hochzeit rows leaked from the initial 50-row fetch.
- Root cause: 2,012 of 29,876 `bph_works` rows have `ubn: null` (legacy pre-Memorix entries). The catalog used `<tr key={w.ubn}>`, so every null-ubn row got `key={null}`. With `reactCompiler: true`, the compiler's per-key memoization collides on duplicate null keys, and shrinking the works array leaves stale null-keyed siblings in the DOM. The counter (`total` scalar) updates correctly because it doesn't go through per-key memoization.
- Fix: fall back to `null-ubn-${idx}` so every row in the current page has a unique key. Applied to both list view and grid view. `setWorks` always replaces the array, so idx-based keys are safe here.

## Test plan
- [x] Reproduced with Playwright before fix: `Visible rows: 4, Counter: 2 of 29,876 works`
- [ ] After preview deploy, re-run the repro against the preview URL and confirm `Visible rows: 2`
- [ ] Manual: search "Helicone" on the preview tenant, confirm only the 2 Turbo entries render
- [ ] Manual: search "Seleniana" (per [the earlier handoff](.claude/handoffs/2026-05-28-bph-catalog-browser-bugs.md)) and confirm no Hochzeit leak
- [ ] Manual: spot-check a third search term and toggle list ↔ covers, no stale rows

## Out of scope
- Backfilling UBNs for the 2,012 null-ubn rows. They're real catalog entries (mostly "fot"-prefixed photographic copies); the right key is structural, not data cleanup.
- The deeper question of whether `reactCompiler: true` produces other duplicate-key collisions elsewhere in the codebase. This PR fixes the BPH catalog only.

### DIFF
--- a/src/components/libraries/BphCatalogBrowser.tsx
+++ b/src/components/libraries/BphCatalogBrowser.tsx
@@ -737,14 +737,22 @@ export default function BphCatalogBrowser({
                     </tr>
                   ))
                 )}
-                {works.map((w) => {
+                {works.map((w, idx) => {
                   const digitized = resolveDigitized(w);
                   const external = resolveExternal(w, !!digitized);
                   const displayTitle = w.title || w.parallel_title || w.uniform_title || '(untitled)';
                   const displayAuthor = w.author || w.variant_author || w.pseudonym;
                   return (
                     <tr
-                      key={w.ubn}
+                      // 2,012 of 29,876 catalog rows have ubn:null (legacy
+                      // pre-Memorix entries). With reactCompiler:true, multiple
+                      // siblings sharing key={null} collide in the per-key
+                      // memoization cache and previous-search rows persist in
+                      // the DOM after works[] shrinks — e.g. searching
+                      // "Helicone" returned 2 rows but rendered 4. Offset the
+                      // fallback key with the row index so each null-ubn row
+                      // is unique within the current results page.
+                      key={w.ubn ?? `null-ubn-${idx}`}
                       className="border-b border-border-light last:border-0 hover:bg-cream/50 transition-colors"
                     >
                       <td className="px-3 py-2 align-top text-secondary hidden sm:table-cell">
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Verified the fix against the preview deploy:

```
$ REPRO_URL='https://sourcelibrary-v2-git-worktree-fix-bp-19a3e0-dereklomas-projects.vercel.app/embed/bph?view=catalog' \
    node scripts/_tmp-repro-helicone.mjs

q=(none), total=29876, rows=50, first="The way of Muhammad"
q=Helicone, total=2, rows=2, first="Turbo, sive moleste et frustra per cuncta divagans"
Visible rows: 2     ← was 4 before the fix
Counter shows: 2 of 29,876 works
```

The `test` CI failure is pre-existing — `tests/unit/proxy-non-tenant-paths.test.ts` can't locate `NON_TENANT_PATHS` after commit `8e348991` refactored `src/proxy.ts`. Unrelated to this PR; will need a separate fix to either the regex or the proxy structure.

The `Vercel – sourcelibrary-v2` "Deployment failed" check is stale — the preview deployment built successfully (the embed page loads, the API responds, and the repro above ran against it). Re-querying the deployment may pick up the actual status.

JDerekLomas:
Confirmed against the Cowork-session findings in `.claude/handoffs/2026-05-28-bph-catalog-browser-bugs.md`:

**A2 (Seleniana ghost rows)** — same root cause. The 2 unrelated `Chymische Hochzeit [Calw]` + `Chymische Hochzeit C. Rosencreutz anno 1459` rows pinned at positions 0–1 are the same null-ubn collision (both have `ubn: null`, sit at indices 0 and 1 of the initial 50-row author-sort fetch).

Repro against preview confirms the fix:

```
$ REPRO_URL='…/embed/bph?view=catalog' REPRO_QUERY='Seleniana' node scripts/_tmp-repro-helicone.mjs

q=Seleniana, total=3, rows=3, first="Seleniana Augustalia"
Visible rows: 3        ← was 5 in handoff
Counter shows: 3 of 29,876 works
```

**A8 (Z→A descending puts no-place/no-year Chymische Hochzeit above real Z-titles)** — same bug, same fix. The Chymische Hochzeit rows weren't actually being sorted to the top; they were stuck at positions 0–1 from the prior null-keyed memoization slot, regardless of the new sort.

This PR resolves Items 2 and 3 in the handoff's "Items worth a follow-up PR" list.

Issue #2149, opened 2026-05-28
Bring hybrid RRF fusion to the user-facing /search (/api/search)


JDerekLomas:
## Context

PR #2137 shipped hybrid Reciprocal Rank Fusion (RRF) search to the **Librarian** (the AI assistant's retrieval) via `src/lib/search/librarian-search.ts` (`hybridSearch()`). On a 17-query golden set it beat every single-lane variant: **P@1 0.529** (vs 0.471 next-best), **MRR 0.632** (vs 0.525), ~2.5× faster than book-then-page, and nearly doubled niche-passage recall (the Boehme "seven fountain spirits" case).

This issue tracks bringing that same fusion to the **user-facing search** at `/search` → `/api/search` (`src/app/api/search/route.ts`), which today runs keyword and semantic lanes side-by-side without principled fusion.

## Why this isn't a copy-paste

`/api/search` is materially more complex than the Librarian path. Any fusion work must preserve:

- **Tenant scoping.** Unlike the Librarian (main-site only, `tenantId: null`), `/api/search` serves tenant subdomains. `hybridSearch()` already accepts `tenantId`, but its tenant path has **never run in production** — it must be exercised and audited against the Tenant Subdomain Lockdown invariant (run `node scripts/audit-bph-leaks.mjs`). The Mongo metadata join in `hybridSearch` drops tenant-foreign hits, but verify the semantic lanes don't leak before the join.
- **Filters.** language / category / collection / date-range / DOI / translation-status. RRF must fuse *within* the filtered candidate set, not after.
- **Pagination.** RRF produces one fused ranking; paginating a fused list (and keeping it stable across pages) needs design — likely fuse a deep candidate pool once, then page the merged list.
- **Work-level dedup** (keep best edition per `work_id`) — currently at `route.ts:638`.
- **The existing relevance ladder** (title-match → exact-phrase → original-language-beats-translation → older-edition → quality). Decide: does RRF *replace* this, or feed it (e.g. RRF for candidate generation, ladder as tiebreaker)?
- **Latency budget.** The route has 8s hard timeouts and Atlas-Search `Promise.race()` workarounds. RRF fans out 3 lanes; confirm it fits the budget under load.

## Known tradeoff to mitigate

[...]

PR #2154, opened 2026-05-29
feat(search): opt-in RRF ranking for /api/search (flag-gated)

JDerekLomas:
Implements the plumbing from #2149.

## What

`/api/search` runs four lanes (Supabase-trigram books, Atlas-Search pages, semantic books, semantic pages) but never fuses their relevance signal — it concatenates results then orders them by a heuristic ladder (books>pages → title-word hits → exact-phrase → original-language → older edition → quality). Semantic similarity only gates results; it never informs ranking.

This adds **Reciprocal Rank Fusion** as an **opt-in** ranking strategy:

- `src/lib/search/rrf.ts` — pure, key-agnostic `rrfScores(rankedLists, k=60)`. Unit-tested (`tests/unit/rrf.test.ts`, 5 cases).
- `route.ts` builds four ranked id-lists (keyed by the same `result.id` scheme the results use — `book.id` for books, `${book_id}-p${n}` for pages) and computes fused scores. When `?ranking=rrf`, the relevance sort uses fused score as the primary key, with the existing ladder as the deterministic tiebreaker.
- Default stays `ladder`. `SEARCH_RANKING_DEFAULT=rrf` can flip the default later without code change. Optional `?rrf_k=` for tuning.

## Why it's safe to merge now

**Default behavior is byte-for-byte unchanged.** With `ranking=ladder` the only added work is computing an unused score map; the ladder logic is extracted verbatim into a named comparator (`ladderCompare`). All four lanes, every filter, tenant scoping, work-id dedup, pagination, and the response shape are untouched. Response + search log now echo `ranking` so we can A/B.

Try it: `GET /api/search?q=<query>&ranking=rrf` vs the same with `&ranking=ladder`.

## Verification

- `npx tsc --noEmit`: clean in touched files (only pre-existing carbon-ledger d3 errors remain repo-wide).
- Unit suite: **219/219** (214 + 5 new RRF tests).
- Live preview sanity check comparing fused vs ladder ordering: see PR comment.

## Out of scope (tracked in #2149)

The **web-search eval** that measures fused-vs-ladder on a golden set, and the decision to flip the default, are the follow-up. Also deferred: extracting a single shared RRF core used by both this route and the Librarian's `librarian-search.ts` (they currently have separate, small implementations); phrase-query lane weighting to mitigate the verbatim-quote regression seen in the Librarian eval.

### DIFF
--- a/.claude/handoffs/2026-05-29-api-search-rrf-and-bph-compare.md
+++ b/.claude/handoffs/2026-05-29-api-search-rrf-and-bph-compare.md
@@ -17,6 +17,27 @@
 
 ---
 
+## ✅ SESSION COMPLETE — final state (supersedes the planning sections below)
+
+Everything below this line was the plan; here's what actually shipped.
+
+- **Found + fixed a CRITICAL tenant leak (#2159, MERGED):** `/api/search`'s semantic BOOK lane returned global/other-tenant books in a partner subdomain's results (up to 18/25 for "hermetic philosophy and the soul" on bph). `match_books_semantic` is global (no `tenant_id` column) and the Mongo materialization didn't re-apply `tenantId`. Fixed both semantic materializations + added `scripts/audit/search-tenant-purity.mjs` (re-runnable guard: pre-fix 43 leaked across 5 queries → post-fix 0).
+- **BPH compare page is LIVE (#2161, preview-only, NOT merged):** access decision resolved as a **tenant-explicit compare API** served on a **preview** (honors "hold #2154"). Built leak-safe (single tenantId-scoped book join). Features delivered:
+  - Two labeled columns: **Current (ladder)** vs **New (RRF)**, per-result 👍/👎 + overall winner + note → `search_compare_votes` (anonymous).
+  - **Scope checkbox** "Restrict to BPH books only" (default on); unchecked compares against the whole Source Library; scope recorded with each vote.
+  - **Clickable results** → BPH books open on `bph.sourcelibrary.org`, others on `sourcelibrary.org` (per-result `tenantOwned` flag; non-tenant rows tagged "main").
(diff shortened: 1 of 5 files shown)


JDerekLomas:
## Live preview sanity check

Query: **"alchemical transformation of the soul"** (a conceptual query, where lane fusion should matter most). Same deploy, `&ranking=ladder` vs `&ranking=rrf`.

Both return **total: 61** — identical candidate set, confirming filters / work-id dedup / tenant scoping / lanes are untouched and only the *ordering* differs.

**ladder (default, unchanged):**
1. [book] The Alchemical Seven-Stars
2. [book] Alchemical Process Drawings — Figura Divina Theosophia
3. [book] The Alchemical Seven Stars
4. [book] The Chemical Art
5. [book] The Sophic Nucleus: An Interpretation of the Tincture…
6. [book] On the Philosopher's Stone
7. [book] The Little Peasant: A Philosophical and Chemical Treatise
8. [book] The Atropic Phoenix Returned from Death

→ all books, ordered by literal title-keyword matching.

**rrf (opt-in):**
1. [book] The Blood of Nature
2. [page] The Hermetic Museum (Vol. 1) p317
3. [page] Aurei Velleris oder der güldin Schatz und Kunstkammer p463
4. [book] On the Philosopher's Stone
5. [page] The Ripley Scroll p1
6. [page] Wasserstein der Weisen / Via veritatis p109
7. [book] The Secret and Hidden Chemical Laboratory
8. [page] A Suggestive Inquiry into the Hermetic Mystery p43

→ blends conceptually-strong books with specific primary-source passages that the ladder buries below every title-keyword book.

Qualitatively the intended behavior. Quantitative fused-vs-ladder on a golden set is the #2149 follow-up.

PR #2283, opened 2026-06-01
feat(curator): "Translate Next" disposition dashboard

JDerekLomas:
## What

A platform-admin **disposition dashboard** over every `confirmed_first` first-translation candidate (~6,950 books). A curator filters the set and acts on each book.

**Page:** `/platform/admin/translate-queue` (superadmin via the protected layout; mutation API gated at editor/curator).

## Why a full dashboard, not a tight work queue

I probed production before designing. The premise ("7,908 books to triage for translation") doesn't match the data:

- **6,499** confirmed_first books are **already fully translated** by us (≥90% readable)
- **131** partial
- **322** untranslated

So the literal "translate next" opportunity is only ~450 books. The durable curatorial value is reviewing dispositions across the *whole* set — catching books where a prior English translation actually exists (mark already-translated), deferring, and queuing the genuine opportunities. That's the scope chosen in the design conversation.

Also: `read_count` is ~0 across the untranslated tail (median 0, max 9) — it can't carry the ranking, so the default sort is the existing deterministic **`processing_priority`** composite (subject + language + scan + scholarly + illustrations + efficiency, from `src/lib/processing-priority.ts`), which is already populated on 90% of these books.

## Changes

- **`GET/POST /api/admin/translate-queue`** (`withEditorAuth`)
  - filter by language / collection / translation status / read count / text search
  - sort by priority score (default), reads, page count, queue priority, recency
  - actions:
    - `queue` (P1–P3) — flags `translation_queue.status:'queued'` for a scheduled run to consume in priority order
    - `defer`
    - `already_translated` — flips `is_first_translation → false` and disposition → `translation_found`, **stashing prior values in `translation_queue.prior` for undo**
    - `clear` — removes the decision, restoring prior flags if it had been marked already-translated
  - global facets (status / queue / language / collection counts)
- **`/platform/admin/translate-queue`** page + client (dark platform theme): summary stat cards (click to filter), filter bar, sortable table with catalog/audit evidence chips, per-row actions, pagination
- **PlatformNav:** "Translate Next" link

## Verification

- `npx tsc --noEmit` clean on all new files (the 14 repo errors are pre-existing d3 type gaps in `carbon-ledger` blog interactives, unrelated)
- Aggregation logic verified against prod: status facet returns 322 / 131 / 6499; filtered+sorted queries return correct composite-ranked rows
- Pre-commit import scan passes

## Out of scope (deliberate)

Wiring the **scheduled run** that consumes `translation_queue.status:'queued'`. The field + P1–P3 priority contract are in place and queryable; the consumer (daily-sourcelibrary / a cron sorted by `translation_queue.priority`) is a focused follow-up.

### DIFF
--- a/src/app/api/admin/translate-queue/route.ts
+++ b/src/app/api/admin/translate-queue/route.ts
@@ -1,0 +1,311 @@
+/**
+ * Curator "Translate Next" queue — disposition dashboard over all
+ * `confirmed_first` first-translation candidates.
+ *
+ * Background: the translation-verification pipeline (PRs #2178/#2190/#2195/
+ * #2205/#2226/#2241) labelled ~6,950 books `translation_verification.disposition
+ * = 'confirmed_first'` + `is_first_translation: true` — "no prior English
+ * translation found, so translating this would be a first." ~93% of those are
+ * already fully translated by us; the actionable tail (~450) is untranslated or
+ * partial. This surface lets a curator review the whole set, filter it, and act:
+ *
+ *   - queue              → flag for translation; a scheduled run picks up
+ *                          `translation_queue.status: 'queued'` in priority order
+ *   - defer              → explicitly not now (keeps it out of the untriaged pile)
+ *   - already_translated → a prior English translation DOES exist; this is not a
+ *                          first-translation candidate. Flips is_first_translation
+ *                          → false and disposition → 'translation_found', stashing
+ *                          the prior values in translation_queue.prior for undo.
+ *   - clear              → remove the curator decision (restores prior flags if the
+ *                          book had been marked already_translated).
+ *
+ * Default ranking is the existing deterministic `processing_priority` composite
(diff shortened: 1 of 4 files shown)


JDerekLomas:
Closing in favor of a focused failure-triage view. Investigating *why* the confirmed_first books are untranslated showed the blocker isn't un-queued work — it's pipeline failures: of 322 untranslated, ~226 are `failed`/`needs_attention` (130 failed + 96 needs-attention) and 50 are already enrolled. A 'translate next' queue would mostly re-point at books that already ran and broke. Grouping by root cause: 103 import_failed_empty (need re-acquisition), 67 ocr_finalize_blocked, 27 OCR FAILED_PRECONDITION (gemini-preview deprecation, root cause already fixed → retryable), 26 image_download_failed. Pivoting to a triage view that groups by failure signature and batch-retries the genuinely-retryable buckets via the existing enroll-pipeline endpoint.

PR #2326, opened 2026-06-01
feat(maintenance): remediate AI recitation-loop translation pages

JDerekLomas:
## Problem

`detect-translation-loops.mjs` already *flags* AI recitation-loop pages but its header says it deliberately leaves the fix for "a future `--fix` pass." That pass never happened, so the loop pages persist.

**~2.4% of pages (~155,000 library-wide)** carry runaway translations: 50,000–65,000 "words" of one repeated phrase, overwhelmingly **non-Latin liturgical scripts** (Tibetan, Hebrew, Ge'ez, CJK). Confirmed example — the Thadrak Kanjur, a 1,613-word Tibetan page translated as *"They see the Dharma."* **×6,522** = 52,411 words. Source model is the **retired `gemini-3.1-flash-lite-preview`**.

These pollute search, embeddings, quotes, and corpus stats, and they inflate the translation word-count ~4× (this is what made [How Big Is the Library?](https://sourcelibrary.org/blog/how-big-is-the-library) need a winsorized estimate).

## Fix: `scripts/maintenance/fix-translation-loops.mjs`

- **DRY-RUN by default**, `--fix` gated, `--book=<id>` for single-book testing.
- **Two-gate safety:** length-flags candidates server-side, then a **precision gate** — the dominant line must be ≥50% of the text — so a long-but-*varied* page is never destroyed. Validated on the Thadrak Kanjur: 46 flagged, **97.8% confirmed** loops, ~2% spared.
- On `--fix`: clears `translation.data`/`_summary`/`_keywords`, stamps a `loop_quarantine` marker, decrements `books.pages_translated`, and resets the book to `ocr_complete` so orchestrator Phase 4 re-queues the now-untranslated pages.

## Why re-translation comes back clean

The current pipeline routes non-Latin scripts to `flash` (not the loop-prone `lite`) via `getModelForBook` in `translate-worker.mjs`. The loops are a legacy artifact of the old `flash-lite-preview`; clearing + re-queueing produces correct translations.

## In scope / out of scope

- **In scope:** the remediation tooling, dry-run-validated on one book.
- **Out of scope (gated):** the mass `--fix` run on ~155K pages is **not** executed here — it re-runs paid translation and mutates thousands of books, so it waits on review of the full dry-run (breakdown posted below). Re-embedding the corrected pages is downstream pipeline work.

### DIFF
--- a/scripts/maintenance/fix-translation-loops.mjs
+++ b/scripts/maintenance/fix-translation-loops.mjs
@@ -1,0 +1,201 @@
+#!/usr/bin/env node
+/**
+ * Remediate AI "recitation-loop" translation pages — the destructive half that
+ * detect-translation-loops.mjs deliberately left for "a future --fix pass".
+ *
+ * BACKGROUND. A page sends the translation model into a token-repetition loop
+ * ("They see the Dharma." × 6,522 → 52k words from a 1,600-word original). The
+ * culprit is the retired `gemini-3.1-flash-lite-preview` running on non-Latin
+ * scripts (Tibetan, Hebrew, Ge'ez, Syriac, CJK, Javanese). The current pipeline
+ * already routes those scripts to `flash` (see getModelForBook in
+ * translate-worker.mjs), which does not loop — so clearing the garbage and
+ * letting the orchestrator re-queue the page produces a correct translation.
+ *
+ * WHAT IT DOES (per flagged page):
+ *   - $unset translation.data / translation_summary / translation_keywords
+ *     (the loop garbage — it pollutes search, embeddings, quotes, and corpus
+ *     stats; it is never real source text).
+ *   - decrements books.pages_translated by the number cleared.
+ *   - resets the book's pipeline_auto.status to 'ocr_complete' and clears any
+ *     stale job, so orchestrator Phase 4 re-queues the now-untranslated pages
+ *     (which then route to flash and translate cleanly).
+ *
+ * TWO-GATE SAFETY. Length alone is a coarse flag (a dense folio can be long).
+ * Before clearing ANY page, this script fetches the translation and confirms it
+ * is genuinely repetitive: the most-frequent line/sentence must account for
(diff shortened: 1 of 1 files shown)


JDerekLomas:
## Dry-run breakdown (sampled)

The full-corpus single-pass scan (`$strLenCP` over ~5.5M translated pages) times out / hits a `$sample` random-cursor error from a Mac→Atlas connection — so the exact full scan and the mass `--fix` should both run **on the Hetzner box** (local Atlas connection, where pipeline maintenance normally runs). Numbers below are from a 40,000-page sample (`corpus-size.mjs --outliers`) plus single-book verification:

**Scope**
- **~2.4% of pages → ~155,000 library-wide** carry runaway translations (>15K words; most cluster at 50–65K, the model's output ceiling).
- **Precision gate confirms ~97–98%** of length-flagged pages as genuine loops (validated on the Thadrak Kanjur: 46 flagged, 97.8% confirmed). ~2% are long-but-varied and are spared at `--fix` time.

**Languages** — overwhelmingly non-Latin liturgical scripts, exactly the set the current `getModelForBook` allowlist routes to `flash`: Tibetan (Kanjur volumes dominate), Hebrew, Ge'ez, Syriac, CJK, Javanese. Latin-script pages are essentially absent.

**Source model** — the retired `gemini-3.1-flash-lite-preview`.

## Recommended rollout
1. Run `node scripts/maintenance/fix-translation-loops.mjs` (dry run) **on Hetzner** for the exact full-corpus count + by-book list.
2. `--book=<id> --fix` on one Tibetan Kanjur volume; confirm orchestrator Phase 4 re-queues it and `flash` returns clean text.
3. Full `--fix`, then let the pipeline re-translate + re-embed the cleared pages.

The `--fix` clears garbage and re-queues — it does not touch any legitimate translation — but it re-runs paid translation over ~150K pages, so it's worth eyeballing the dry run first.

Issue #2416, opened 2026-06-03
Persist IIIF manifests: store collection + bulk backfill of 13.5K imported books + capture-on-import


JDerekLomas:
## Why

Today a IIIF manifest is fetched **once, at import time** (`importBookFromIIIF()` in `src/lib/import-utils.ts`, and `/api/import/iiif/route.ts`), parsed for per-page image URLs + metadata, then **discarded**. We persist only:

- the manifest **URL** → `book.image_source.iiif_manifest`
- the extracted page image URLs → `pages.photo` / `photo_original` / `thumbnail`

The raw manifest JSON is never stored. Consequences:

1. **No provenance.** If a source takes its manifest offline or mutates it, we can't reconstruct the original structure, labels, rights, or canvas order. We keep page URLs but lose the source of truth.
2. **No generic re-archive.** PR #1674 re-archives Vatican books by re-fetching the manifest at archive time — provider-specific and re-hits the source. A stored manifest makes re-archive generic and offline.
3. **No drift detection.** Issue #1504 (1/3 of IA books have IIIF-vs-JP2 page mismatch) needs the manifest's canvas list to compare against stored pages.
4. **Slow imports.** Every import blocks on a live manifest fetch. Pre-extracting page URLs in bulk makes the import step instant.

Archiving is already fully decoupled from manifests (the archive workers read `pages.photo`, never touch manifests) — so this change is orthogonal to the slow per-page archiver.

## Scale (measured 2026-06-03, `bookstore`)

- **13,459** books carry a `iiif_manifest` URL (10,391 visible). Top providers: internet_archive 4931, bl 1460, mdz 1427, e-rara 1149, bsb 959, harvard 827, leiden 597, allard_pierson 489, umedia 393, …
- **943,396** `import_candidates` are `status: discovered` (each has a `manifest_url` but no fetched manifest).
- `iiif_manifests` collection does not exist yet.

## Design

New Mongo collection **`iiif_manifests`**, unique on `manifest_url`:

```
{
  manifest_url,            // unique key
  book_id,                 // nullable — set for imported books, null for candidate-only
  source / provider,       // e.g. 'harvard', 'mdz'
  iiif_version,            // 2 | 3
  label,                   // manifest label (display)
  page_count,              // canvas count
  image_urls: [{ photo, thumbnail }],  // pre-extracted, import-ready
  raw,                     // the manifest JSON (well under Mongo 16MB)
  http_status,
[...]

PR #2441, opened 2026-06-05
feat(iiif-discovery): Wikidata P6108 manifest source (#2416)

JDerekLomas:
Adds a **Wikidata** discovery source to the existing `scripts/iiif-discovery/` system (#2416).

## What
`scripts/iiif-discovery/sources/wikidata.mjs` harvests IIIF manifest URLs from Wikidata property **P6108** into `import_candidates`, matching the existing source conventions (`withMongo` + `insertCandidates`, `source: 'wikidata'`).

- `--manuscripts` restricts to `P31 = Q87167` — **~74K multipage codices** (median ~188 canvases, sampled).
- Full P6108 is ~306K items (the rest is mostly single-object art — paintings/prints/drawings — so `--manuscripts` is the useful default for a reading library).

## Why it's net-new
Wikidata P6108 is a **cross-institution** channel the current OAI-PMH/SRU/Collections sources don't cover. Validated: **~80% of the first 17K manuscripts are not already in `import_candidates`** — surfacing SMK Copenhagen, Yale, Penn, National Library of Wales, Toronto, Leiden, Ryukoku, Tübingen, Vanderbilt, Chicago, ISOS Dublin, and more. Small real run inserted 3,678 candidates as proof.

## Notes for reviewers
- **Does NOT call `candidate-store.ensureIndexes()`** — it builds a plain `unique` index on `manifest_url`, which now **fails**: `import_candidates` holds ~91K `manifest_url: null` docs and the live index is the **sparse, non-unique** `manifest_url_lookup`. Heads-up: because that index isn't unique, `insertCandidates`' "skip if exists" is **not DB-enforced** anymore — worth a look for the other sources too. This source therefore dedups **in-app** against existing `manifest_url`s (indexed `$in` lookup) + an in-run `seen` set.
- SPARQL pagination uses **no `ORDER BY`** (sorting 306K times out WDQS); unordered `LIMIT/OFFSET` overlaps slightly, absorbed by the dedup.

## Scope
- In: the source + label-free discovery (import resolves real titles from the manifest, per the existing flow).
- Out: the full ~74K run (kept manual pending review — run `--manuscripts` with no `--max-records`); any change to the shared `ensureIndexes`/index strategy (flagged, not touched).

### DIFF
--- a/scripts/iiif-discovery/lib/candidate-store.mjs
+++ b/scripts/iiif-discovery/lib/candidate-store.mjs
@@ -17,7 +17,12 @@
  */
 export async function ensureIndexes(db) {
   const col = db.collection('import_candidates');
-  await col.createIndex({ manifest_url: 1 }, { unique: true });
+  // Sparse, NON-unique — matches the deployed `manifest_url_lookup` index.
+  // manifest_url is intentionally not unique: the collection retains duplicate-
+  // manifest rows from overlapping sources and resolves them via dedupe-candidates.mjs
+  // (FRBR clustering → mark non-primary `skipped`), not by deletion. ~91K rows also
+  // have a null manifest_url, so a plain unique index can't build (it threw before).
+  await col.createIndex({ manifest_url: 1 }, { name: 'manifest_url_lookup', sparse: true });
   await col.createIndex({ source: 1, status: 1 });
   await col.createIndex({ status: 1, date_earliest: 1 });
   await col.createIndex({ source: 1, source_id: 1 });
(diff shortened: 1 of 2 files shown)


JDerekLomas:
Updated: also fixes the stale `ensureIndexes()` (it built a UNIQUE manifest_url index that threw on the ~91K null-manifest_url rows). After reading `dedupe-candidates.mjs`, the right fix is to align it with the deployed **sparse, non-unique** `manifest_url_lookup` — manifest_url is non-unique **by design** (dedup is FRBR clustering → mark non-primary `skipped`, not deletion). No rows deleted, no unique index forced.

Also ran the full source: **+47,900 net-new manuscript candidates** (61,427 seen, 13,527 already covered) across 52 institutions — `import_candidates` now 1,180,812. The ~1,319 exact-manifest dup groups are left for `dedupe-candidates.mjs` to cluster, per the existing design (426 are already `discovered+skipped`).

PR #2512, opened 2026-06-10
feat(cite): include original imprint (place, publisher, year) in citation output

JDerekLomas:
## Why

Footer feedback from Derek (2026-06-04, on /book/drebbel-tractatus-duo-1628-latin-laurembergius): **"put the bibliographic info in the cite info."**

Citations generated by the Cite button and the quote API credited only "Source Library" as publisher and silently dropped the original imprint (place of publication, printer, year) — the bibliographic data scholars actually need to identify the edition.

## What

- **New `src/lib/citations.ts`** — shared `generateCitations()` + `formatImprint()`. The global (`/api/books/[id]/quote`) and tenant (`/api/[tenant]/books/[id]/quote`) routes each had a byte-identical private copy of the citation generator — the same duplication pattern that caused the #2420 quote-surface drift. Both now import the shared module.
- **All citation formats** (inline APA, Chicago footnote, bibliography, Chicago author-date, MLA, BibTeX) now include `Place: Publisher, Year` when the book records an imprint. BibTeX emits proper `publisher`/`address` fields with Source Library kept as `translator`.
- **CiteButton** already received `publisher`/`placePublished` props from both call sites (book page, TranslationEditor) but never used them — now both APA and BibTeX outputs include them.
- Graceful degradation: books with no imprint data produce the exact same citation text as before.

## Out of scope

- Shelfmark / holding-institution attribution in citations (per-book provenance stays in BibliographicInfo).
- The quote API's `license.attribution` block — unchanged.

## Testing

- `npx tsc --noEmit` clean; all 206 unit tests pass.

### DIFF
--- a/src/app/api/[tenant]/books/[id]/quote/route.ts
+++ b/src/app/api/[tenant]/books/[id]/quote/route.ts
@@ -1,7 +1,8 @@
 import { NextRequest, NextResponse } from 'next/server';
 import { getReadDb } from '@/lib/mongodb';
 import { Book, Page, TranslationEdition } from '@/lib/types';
-import { getShortUrl, getRequestBaseUrl } from '@/lib/shortlinks';
+import { getRequestBaseUrl } from '@/lib/shortlinks';
+import { Citation, generateCitations } from '@/lib/citations';
 import { markForExport } from '@/lib/provenance';
 import { stripEditorialWrappers } from '@/lib/strip-editorial-wrappers';
 import { isBot, isTrustedBot, botMaxPage } from '@/lib/bot-gate';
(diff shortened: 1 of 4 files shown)


JDerekLomas:
Closing as **superseded** — the original-imprint feature this PR adds is already live on `main`, implemented inline and **more completely**.

Both quote routes (`/api/books/[id]/quote` and the `[tenant]` twin) and `CiteButton.tsx` already render the original imprint in footnote / bibliography / Chicago / MLA / APA / BibTeX. Main's version includes `place_published` + `publisher` **plus `format` and `ustc_id`** (e.g. BibTeX `address`/`publisher` + a USTC note); this PR only carried place + publisher.

Resolving the rebase toward this branch would either regress the live citations (dropping format/USTC) or require re-doing the shared-`citations.ts` refactor against main's richer logic — net negative on the quote-integrity path. The DRY extraction (shared `src/lib/citations.ts`) is the only unshipped piece and isn't worth a regression; if we want it later, it should be a fresh extraction *of main's current logic*, not this older copy.

PR #2522, opened 2026-06-15
fix(security): gate hidden (visible:false) books on reader + content APIs

JDerekLomas:
## Problem

`visible: false` only removed a book from listings/search/sitemap. The **reader page and every content API still served the full text at the direct URL**, so deliberately-hidden books — including in-copyright modern editions/translations — were fully readable by anyone with the link.

Found while checking the hidden 1995 Garfield translation of the *Mūlamadhyamakakārikā*: page `200`'d, `/page/1` `200`'d, and `/api/books/<id>/quote` returned the verbatim copyrighted translation, despite `visible: false`.

## What this does

Gates **every public read surface** so hidden books 404 for the public, while preserving access for **editors (logged-in browser)** and **Claude Code / the pipeline (`Authorization: Bearer $CRON_SECRET`)**. Direct Mongo/MCP access is unaffected.

### Scope boundary (important)
Gates **`visible === false` only** — not `visible !== true`. The ~6,498 legacy `visible: null`/missing books keep their current public behavior. (Counts at time of change: 30,681 visible:true · 18,364 visible:false · 6,865 of those with real OCR text = the exposure closed.)

### Surfaces
- **`src/lib/book-access.ts`** (new): `isHiddenBook` + `isBookReadable(book, request)`. Public books short-circuit before any session lookup.
- **Reader pages** `/book/[id]`, `/book/[id]/page/[pageId]`: hidden → `notFound()`, gated **statically** (no per-user session read) so the 24h ISR/CDN cache is preserved. `generateMetadata` noindexes hidden.
- **Editor browser access**: existing `/book/[id]/preview` (already `force-dynamic`) is now **editor-gated** + passes `allowHidden`; new matching `/book/[id]/page/[pageId]/preview` for per-page reading of hidden books — keeps the high-volume per-page ISR route static.
- **Content APIs** (global + tenant twins): `quote`, `text`, `search`, `download`, `index` (GET), `chat` (GET); **IIIF** `manifest`, `search`, `autocomplete`, `canvas/[n]/{ocr,translation}`. Each 404s a hidden book unless editor/CRON_SECRET.

## Why caching is safe
A naive session check on `/book/[id]` would opt the route into dynamic rendering and kill the 24h CDN cache the project depends on. Instead the public route gates on the data only (uniform 404), and editor access lives on separate `force-dynamic` preview routes.

## Verification
- `npx tsc --noEmit`: clean.
- After the Vercel preview builds: a hidden book should 404 on the reader + all listed APIs; a visible book unchanged; `/book/<id>/preview` should 404 logged-out and render logged-in as editor; `Authorization: Bearer $CRON_SECRET` should still fetch hidden text.

## Out of scope / notes
- MCP tools that call these APIs without `CRON_SECRET` will now 404 on hidden books — intended (don't serve in-copyright text to external LLMs). Derek's CC access uses CRON_SECRET/Mongo.
- Did not touch the ~6.5k null-visible books, image/asset routes, or admin/mutation routes (already editor-gated).

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -145,6 +145,17 @@ const nextConfig: NextConfig = {
         missing: [{ type: 'header', key: 'rsc' }],
         headers: [{ key: 'CDN-Cache-Control', value: 'public, max-age=86400, stale-while-revalidate=3600' }],
       },
+      // Editor-only hidden-book preview routes are force-dynamic + auth-gated and
+      // MUST NOT be edge-cached — a cached editor 200 would leak to anon users.
+      // Placed AFTER the broad /book rule so these specific paths override it.
+      {
+        source: '/book/:id/preview',
+        headers: [{ key: 'CDN-Cache-Control', value: 'private, no-store' }],
+      },
+      {
+        source: '/book/:id/page/:pageId/preview',
+        headers: [{ key: 'CDN-Cache-Control', value: 'private, no-store' }],
+      },
       {
         source: '/collections/:path*',
         missing: [{ type: 'header', key: 'rsc' }],
--- a/src/app/api/[tenant]/books/[id]/chat/route.ts
+++ b/src/app/api/[tenant]/books/[id]/chat/route.ts
@@ -5,6 +5,7 @@ import { withAuth } from '@/lib/auth-helpers';
 import { getChapterTexts } from '@/lib/chapter-text';
 import { z } from 'zod';
 import { resolveTenantId } from '@/lib/tenant-context';
+import { isBookReadable } from '@/lib/book-access';
 
 // Validation schema for chat messages
 const messageSchema = z.object({
@@ -363,6 +364,11 @@ export async function GET(
 
     const book = await db.collection('books').findOne({ id, tenantId });
     if (!book) {
+      return NextResponse.json({ error: 'Book not found' }, { status: 404 });
+    }
+
+    // Hidden (visible:false) books are not public — 404 unless editor / CRON_SECRET.
+    if (!(await isBookReadable(book, request))) {
       return NextResponse.json({ error: 'Book not found' }, { status: 404 });
     }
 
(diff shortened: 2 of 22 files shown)


JDerekLomas:
## Verified on the preview deploy (commit b6c434ab)

Tested against the hidden 1995 Garfield *Mūlamadhyamakakārikā* (`visible:false`) and a visible control book.

| Surface | Anonymous | CRON_SECRET / editor |
|---|---|---|
| `/api/books/:id/quote` | **404** ✅ | 200 ✅ (bearer) |
| `/api/books/:id/text` | **404** ✅ | 200 ✅ (bearer) |
| `/api/iiif/:id/manifest` | **404** ✅ | — |
| `/api/iiif/:id/canvas/:n/ocr` | **404** ✅ | — |
| `/api/books/:id/search` | **404** ✅ | — |
| `/api/books/:id/download` | 400 (no leak) ✅ | — |
| `/book/:slug` reader | "Book Not Found", noindex, **no content** ✅ | — |
| `/book/:slug/preview` | "Book Not Found" / unauthorized, **no content** ✅ | renders for editor (logic) |
| visible control `/book/:slug` | 200, **still edge-cached** (`max-age=86400`) ✅ | — |
| `/book/:slug/preview` cache header | `CDN-Cache-Control: private, no-store` ✅ (was leaking via the blanket `/book/*` 24h rule) | — |

**Confirmed the leak is live on prod:** `https://sourcelibrary.org/api/books/69b63ecba85a63b8ab194e60/quote` returns 200 with the verbatim copyrighted translation today.

**Note on status codes:** hidden books render the app's "Book Not Found" UI at HTTP 200 — identical to how a genuinely non-existent book behaves (verified a fake slug returns 200 on prod too). So this is consistent pre-existing behavior, not a regression; no book content is served and the page is `noindex`.

**One manual check for a human:** log in as an editor and confirm `/book/<slug>/preview` renders the hidden book in-browser (the anonymous-blocked path is verified; the logged-in path is logic-verified but I can't curl an authenticated session).

CI: `tsc` clean, 306 unit + 23 integration tests pass.

PR #2583, opened 2026-06-19
fix(archive): decouple archiving from the global spend-pause

JDerekLomas:
## Why
`archive-ocr.mjs` exited whenever `processing_control.paused` was set. That flag pauses Gemini **spend** (OCR/translation). Archiving only fetches source images into R2 — **zero Gemini cost** — so it was needlessly halted, leaving the draft backlog (~23k books, ~2.3k English) unarchived while the spend-pause was on.

## Change
Gate archiving on a dedicated `processing_control.archiving_paused` flag instead of the global `paused`. Archiving keeps running while OCR/translation are paused, with its own off-switch.

## Context / out of scope
- Pipeline core crons (archive, OCR submit/collect, orchestrator phases 1–8) are currently disabled on Hetzner pending a scheduler migration — this change removes the spend-pause as a blocker but doesn't re-schedule archiving (migration-owned).
- IA / Gallica / Harvard archive from the Mac side (block Hetzner IPs); `archive-ocr.mjs` covers Hetzner-reachable providers (IIIF/BSB/Cambridge/Vatican/etc.).

### DIFF
--- a/scripts/workers/archive-ocr.mjs
+++ b/scripts/workers/archive-ocr.mjs
@@ -230,10 +230,13 @@ async function main() {
   const client = await MongoClient.connect(process.env.MONGODB_URI, { maxPoolSize: 10 });
   const db = client.db('bookstore');
 
-  // Check processing_control pause
+  // Archiving is decoupled from the global `paused` (Gemini-spend) flag: archiving
+  // fetches source images into R2 and costs no Gemini, so it should keep running
+  // while OCR/translation are paused for spend. It has its own dedicated
+  // off-switch — `archiving_paused` — for when archiving specifically must stop.
   const control = await db.collection('system_config').findOne({ _id: 'processing_control' });
-  if (control?.paused) {
-    console.log(`[archive-ocr] Pipeline paused. Exiting.`);
+  if (control?.archiving_paused) {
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Closing as **superseded by the selective-unpause system (#2616 / #2650)**.

This PR (written 2026-06-19) decoupled archiving from the global `paused` flag, gating it instead on a dedicated `archiving_paused` field. Two days later (2026-06-21) #2650 landed a more precise mechanism: all four archive workers (`archive-ocr/bulk/harvard/gallica`) now call `shouldBypassPause(control)`, so during a global spend-pause archiving still runs **for selectively-unpaused scopes** (book ids / collections) — getting those books' images to R2 to feed the paid path — while holding back the rest of the backlog.

Live prod state confirms this is working: `paused: true`, no `archiving_paused`, and a scope of 47 books + 1 collection that archiving currently honors.

Merging this as-is would make `archive-ocr` ignore `paused` and archive the **entire** backlog to R2 (not just the scoped books) — the exact mass-archiving cost that #2616's scope-gating was built to prevent — and would make `archive-ocr` inconsistent with the other three archive workers. The underlying need (archiving runs while OCR/translation are paused) is already met, more precisely, by the scope mechanism.

If full archiving-exemption from the spend-pause is ever wanted, that's a deliberate change across all four archive workers, not a one-file swap.

Issue #2453, opened 2026-06-06
Universal works catalog (pre-1900): work-level union catalog across all traditions — translation status, scan sources, holdings


JDerekLomas:
## Why

Source Library's mission claim — *putting primary sources into people's hands, translated* — rests on a question we can currently only answer anecdotally: **of everything humanity wrote before 1900, what survives, where, and how much of it can an English reader actually read?**

We keep answering it one sample at a time: the USTC census (~2% of early modern Latin print translated), the Siku census (~1–2% of the Chinese imperial canon, #2234), the first-translation audits. Each census rebuilds the same machinery — work identity, source availability, translation evidence — and throws it away. Meanwhile every import session re-answers "do we hold this? where are scans?" by hand, and dedup gets harder at scale (#2318).

A **universal work-level catalog** (pre-1900, all traditions) makes all of these the same query. It is also, in itself, the flagship publishable resource: a map of the world's untranslated written heritage — the strongest version of the translation-gap story for scholars, press, and funders (cf. translation-gap-site).

## Schema (generalizes #2452 — drop the `cn_` prefix)

Supabase, following the `ustc_editions` / ILP reference-catalog pattern:

- **`works`** — one row per work: canonical title (original script + romanized + English), `tradition`, `original_language`, author (→ authors thesaurus where possible), date/century, `wikidata_qid`, per-tradition authority IDs (JSONB: `kanripo_id`, `openiti_uri`, `bdrc_id`, `ustc_work_cluster`, `pandit_id`, …), `translation_status` (`none | partial | full`) **+ evidence URL + method (`census-auto | hand-verified | claimed`)** — the `is_first_translation` lesson applies: a status is a *claim* with provenance, never a bare boolean.
- **`work_sources`** — availability: `work_id`, source slug, source ID / manifest URL, kind (`scan | transcription | translation`), volume/juan coverage, IIIF flag.
- **`work_holdings`** — our books: `work_id`, `book_id`, coverage (volume/juan range), pages, translated %.

Identity spine: **Wikidata QID where it exists**, per-tradition authority IDs everywhere, variant normalization per script (CJK variant glyphs are documented in `.claude/docs/chinese-iiif-sources.md`; Latin gets i/j u/v normalization; etc.), Gemini adjudication for ambiguous tails (pattern: `scripts/lib/author-reconcile.mjs`). Edition→work clustering for edition-level catalogs is exactly #2318.

[...]

Issue #2318, opened 2026-06-01
Work-identity layer: stop import dedup from getting harder at scale


JDerekLomas:
## Why
As we scale imports (recent Daoist/alchemy batches; Sefaria; ongoing curation), the recurring risk is **duplication** — and it gets harder as the catalog grows. We have good *manifestation*-level dedup but no *work*-level identity, so "do we already have this text, in any edition?" is currently a fuzzy guess that fails on CJK, romanization variants, and alternate scans.

This issue proposes the durable fix. Two cheap pieces already shipped (PR #2290): the hidden-backlog dedup fix and a reusable `enumerate→dedupe→source` tool. This issue is the third, larger piece.

## What exists today (`src/lib/dedup.ts`, wired into all 9 `/api/import/*` routes)
Three-tier `checkDuplicate()`:
1. **Source fingerprint** (exact) — `ia:…`, `gallica:…`, `iiif:…`. Reliable. This is the **manifestation** layer and it works.
2. **Normalized title+author** (exact match on normalized strings) — works for European text.
3. **IIIF manifest URL** (exact).

Plus `scanForDuplicates()` for periodic audit.

## Gaps
1. ✅ **FIXED (PR #2290):** tiers 2/3 and the audit scanner filtered `visible: true` → blind to the hidden import backlog (the regime we import into at volume).
2. **No fuzzy distance in tier 2.** Exact normalized-string equality only. `normalizeTitle` strips *European* volume words (vol/tomus/band) but not CJK 卷上/卷中/卷下, and doesn't fold variant chars (真/眞, 参/參) or romanization drift (Wuzhen/Wu-chen). CJK effectively dedupes on fingerprint alone.
3. **No work-level identity.** `work_id` (FRBR/WEMI "all editions of one work") exists in the schema and on `/work/[id]`, but **nothing assigns it at import**. Consequences seen this week:
   - Multi-volume works imported as N loose books (Zhen'gao ×5, Yunji qiqian ×10) instead of one work.
   - Alternate scans of a held work classify as NEW (e.g. IA `06076376-78` 抱朴子內篇 vs held `06049004-07`/`06076375`) — same *work*, different *manifestation*, invisible to fingerprint dedup.
   - An existing intra-library dup: Baopuzi Inner Chapters held twice (`06049004-07` Siku + `06076375`).

[...]

Issue #2264, opened 2026-05-31
Work-identity resolver: canonical works on the author thesaurus (FT + ILP recall) — step 3 of #2179


JDerekLomas:
## TL;DR

Build a **work-identity layer** — `(canonical_author_id, uniform_title) → work_id` — sitting on top of the canonical author thesaurus (#2250), then **join that work_id across the external reference catalogs** (`translation_catalogs`, USTC `catalog_coverage`, the ILP/Index catalog). This is the layer that actually moves the **first-translation** and **Index-Librorum-Prohibitorum** numbers, because both are *per-work* questions that author identity alone can't answer.

Step 3 of #2179 (authors = step 2/2.5; works = step 3). Precision-first, same as #2218/#2250: **under-cluster over mis-cluster** — wrongly merging two editions into one work can produce a false "first translation" claim.

## Why (start here)

Two flagship efforts are blocked on the same missing primitive:

- **First translation.** "Is this the first English translation of work X?" requires knowing *every* edition/expression of work X across title and language variants. The census already diagnosed the failure mode: residual misses are *title-divergent translations* (same work, different title) that never match. Grounded today: **6,369 FT-flagged books, but 0 have `original_title` and only 720 have `original_language`** — the FT flag is a bare claim with no work anchor.
- **Index Librorum Prohibitorum.** "Do we hold this banned work?" is a work-level join. The ILP author-held join (entity-join lifted Pico 0→12) is the author-axis version; **work-held is still pending** (#2156). It needs work identity to match an Index entry (Latin censor title) to our edition.

A reader/scholar wants *one* page per work showing all editions and translations, with an authoritative "first English translation: yes/no, here's the evidence." That's navigation + scholarly accountability — same thesis as the author thesaurus, one level up.

## Current state (measured 2026-05-31 — measure before building)

- **No `work_id` anywhere.** `books` has only an (unpopulated) `original_title` field. No uniform-title/OCLC/FRBR clustering.
- **Author axis is done** (#2250): `authors` thesaurus (4,765 canonical persons), `books.author_id` on the linked set. The hard disambiguation problem (same name → different people) is already solved — works inherit that spine.
[...]

Issue #76, opened 2026-02-25
Historical Timeline: lifespan visualization for 1,865 dated entities


JDerekLomas:
## Motivation

Wikidata alignment has given us **1,865 entities with birth/death dates** (or floruit periods). A timeline visualization lets users see who was alive when, how intellectual movements clustered in time, and which figures were contemporaries. This is especially valuable for understanding the transmission of esoteric knowledge across centuries.

## Data Available

- `entities` collection: 1,865 documents with `wikidata.birth_year` and/or `wikidata.death_year`
- Entity types: mostly `person`, some `concept` (for movements/schools with date ranges)
- Each entity has `book_count`, `total_mentions`, `type`, `description`
- Date range spans ~3000 BCE to ~1900 CE, with heavy concentration in 1400-1700 CE

### Library: Pure SVG (no heavy charting library)
- **Why SVG:** Lightweight, consistent with existing chart patterns in the codebase, full control over styling
- **Why not D3:** Overkill for a timeline; React + SVG is simpler to maintain
- **Interaction:** Native DOM events on SVG elements, no library needed

### API Endpoint

**`GET /api/explore/timeline`** — returns entities with date ranges

```typescript
{
  entities: Array<{
    id: string;
    name: string;
    type: 'person' | 'concept';
    birth_year: number | null;
    death_year: number | null;
    book_count: number;
    total_mentions: number;
    description?: string;
  }>;
  stats: {
    total: number;
    year_range: [number, number];
    by_century: Record<string, number>; // for density histogram
  };
}
```

### Components

| Component | Purpose |
|-----------|---------|
| `src/app/explore/timeline/page.tsx` | Page wrapper with metadata |
| `src/components/explore/Timeline.tsx` | Main SVG timeline component |
| `src/components/explore/TimelineBar.tsx` | Individual lifespan bar |
| `src/components/explore/DensityHistogram.tsx` | Century distribution at top |
| `src/components/explore/TimelineControls.tsx` | Zoom, filter, color controls |

### Key Behaviors

1. **Lifespan bars** — horizontal bars from birth_year to death_year, vertically stacked
   - Entities without death_year: use birth_year + 70 (estimated lifespan) with dashed end
   - Entities without birth_year: skip or show as point markers at death_year
2. **Density histogram** — fixed bar chart at top showing entity count per century, acts as navigation (click century to scroll)
[...]

PR #2615, opened 2026-06-20
feat(shwep): scale inline-linked reading lists to all episodes via process matcher

JDerekLomas:
## What

Scales the SHWEP Reading Room's inline-linked reading lists from **2 hand-curated episodes (#76, #323) to 132 episodes**, using the validated two-step **process matcher** in place of the old embedding + single-confirm matcher.

## Why

Earl Fontainelle's bibliographies cite hundreds of primary sources we hold and have translated, but readers had no way to jump from the citation to our facsimile+translation. The old matcher missed cross-language editions and risked same-author/wrong-work false positives. The process matcher fixes both — and the inline links are placed so they never imply an edition provenance we don't have.

## How

**Holdings (Stage 3) — the process matcher.** For each cited work:
1. **NORMALIZE** → canonical title forms in all languages + extant/lost status.
2. **embed the title forms → retrieve → per-work LLM confirmer** that reasons per candidate, keeping only editions/translations/commentaries of *that exact work by that author*. Lost works → explicit no-holding (never a same-author fallback).

Mirrors the eval that scored **10/10 on #76** (vs 6/10 single-shot — recovers City of God, *Theologia Platonica*, *in Politiam Platonis*) and **9/9 with 0 false positives** on the trap-laden Islamicate/Byzantine tail. Wikidata work-level coverage is ~45% (≈0% Islamicate, see #2318), so the LLM confirmer stays in the loop; canonical title forms are a precision overlay, not the sole mechanism.

**LinkBib (Stage 5, new).** Injects inline `/book/<slug>` "read here" links into each episode's *displayed* bibliography, only for works we publicly hold (visible + translated) via a new `bestEdition` helper. The LLM only *locates* each held work's mention; the link is **spliced into the original text in JS**, so the bibliography stays **byte-for-byte verbatim** apart from the added markup.

Two guards keep the links honest:
- **Never splice inside an existing bracket span** (existing links, abbreviation keys, editorial notes) — no broken/nested markdown.
- **Never link a mention inside a dated edition citation.** A title in "Author, Ed., *1997*. Title. Publisher." is a reference to a *specific* edition; our link points at a *different* edition we hold, so wrapping it there would misattribute provenance (the ep-13 *Théologie Platonicienne* → Saffrey-Westerink 1997 case). Years appear in formal citations, never in Earl's in-discussion shorthand, so this keeps every #76-style link and drops the edition-citation ones. Formal-citation episodes fall back to the honest work-level "Read in Source Library" list (`shwep-book-matches`), which makes a *work* claim, never an *edition* claim.

A span the LLM returns that isn't a clean, undated, verbatim slice is dropped — never fabricated. Hand-curated #76/#323 are preserved verbatim as the quality bar.

## Output (tracked data files)

- `src/data/shwep-cited-works.ts` — 464 works (230 held, 234 to acquire); `held[]` now carries `slug` + `translated`
- `src/data/shwep-book-matches.ts` — 204 episodes
- `src/data/shwep-linked-bibliographies.ts` — **132 episodes, 394 inline links** (up from 2 hand-curated)

## Verification

- ✅ `npx tsc --noEmit` clean; pre-commit import check clean
- ✅ **0 fidelity mismatches** (stripping injected links reproduces the source bibliography exactly), **0 broken markdown**, **0 links inside a dated entry**
- ✅ Trap precision intact: *De regressu animae* / Damascius *In Phaedonem* / *Book of the Eparch* / Psellos *Epistle on Chrysopoeia* correctly **not held**; Malalas *Chronographia* **held**
- ✅ Cross-tradition spot-check (Greek/Latin, Islamicate, Byzantine, Jewish, late-antique) — all sampled link targets correct
- ✅ ep 13 (formal-citation bib) no longer mislinks → falls back to work-list

## Known limitation / out of scope

- **Mixed episodes** (some discussion mentions + some dated citations) emit the inline bib and hide the separate work-list, so a held work mentioned only in a dated citation stays un-clickable (still visible in Earl's text). Honest (nothing misattributed) but a discoverability cost; a "show both" page change is a possible follow-up.
- The author guard favours precision: for traditional-attribution works (e.g. Song of Songs "by Solomon") it can drop Anonymous-cataloged editions, but always keeps a valid one when present.
- `acquire.md` (234 works) is a reference output, not part of this PR. Re-embedding / Wikidata QID overlay not added (tracked under #2318).

Refs #2318, #2453, #2264.


JDerekLomas:
Added a recall pass: holdings now retrieves per-title-form and unions candidates (vs a single blended query), recovering divergent-title/cross-language holdings the old retrieval missed (e.g. the McCown *Testament of Solomon*).

Final numbers: **236 works held** (was 230), **228 to acquire** (was 234), **134 episodes inline-linked / 399 links**, 205 episodes with a held source. Integrity unchanged: 0 fidelity mismatches, 0 broken markdown, 0 links inside a dated edition citation. Precision unaffected — the per-candidate confirmer + author guard still gate every hit.

JDerekLomas:
Closed the mixed-episode discoverability gap. A linked bibliography was hiding the whole held-works grid, so a work held but cited only via a dated edition (e.g. Eunapius' *Lives of the Sophists* via the 1922 Loeb in #250) had no reader path.

LinkBib now emits `src/data/shwep-supplementary-works.ts` — per linked episode, the held works (one representative edition each) not reachable via an inline link. The episode page renders them as an **"Also in Source Library"** grid beneath the reading list. **71 linked episodes carry 128 such works.** Hand-curated #76/#323 unaffected; work-level dedup avoids edition clutter; tsc clean.

Verified on preview: #250 now shows Zosimus inline + a 2-card grid including *Lives of the Sophists*.

Issue #2632, opened 2026-06-20
SHWEP Reading Room: acquisition + translation sweep for cited primary sources


JDerekLomas:
Make the historical works cited across SHWEP episodes readable on Source Library. Triage from the work-identity matcher (PR #2615): ~195 cited works are unheld, plus a handful held only in barely-translated original-language editions. This issue is the costed worklist to close the high-value gaps.

**Operational note:** the OCR→translation pipeline is deliberately paused (Gemini-spend hold). Even the public-domain *English* acquisitions need OCR to be readable/quotable on the platform. Execute by importing hidden, then **processing only these specific book ids** (selective OCR, not a global un-pause), then QA → visible, then re-run the SHWEP matcher (`scripts/enrichment/shwep-cited-works.mjs --holdings --emit --linkbib`) to pick them up. Estimated total Gemini spend: well under ~$10 (OCR of PD English + OCR/translate of the two Latin works).

## ACQUIRE — public-domain English (import via `/api/import/ia`, hidden)
- [ ] **Philo of Alexandria** — C.D. Yonge, *Works of Philo* (PD). Unlocks **12 cited works** (De Abrahamo, De Cherubim, De vita Mosis, De Decalogo, De Aeternitate Mundi, De Confusione Linguarum, De vita Contemplativa, De Agricultura, De Congressu, On the Posterity of Cain, Quod Deterius, Philonic corpus). Source: archive.org `worksofphilojuda01yonguoft` (+ vols 2-4); avoid the borrow-only `worksofphilocomp0000phil`. *Biggest single win.*
- [ ] **Gregory of Nyssa, Life of Macrina** (3 ep) — Lowther Clarke 1916 (`lifeofsaintmacri0000greg`) or NPNF ser.2 vol.5.
- [ ] **Acts of Peter** (2 ep) — Ante-Nicene Fathers vol.8 (`ante-nicene-fathers-vol-8`).
- [ ] **Aristotle, De generatione animalium** (2 ep) — Platt, *Works of Aristotle* vol.V 1910 (`worksofaristotle0005jand`).
- [ ] **Augustine, Sermones** (2 ep, partial) — NPNF ser.1 vol.6.
- [ ] **Clement of Alexandria, Adumbrationes in 1 John** (2 ep) — ANF (confirm vol/section).
[...]

Issue #2567, opened 2026-06-19
[Tracking] Translation & works knowledge layer — architecture map + cleanup


JDerekLomas:
**Tracking / architecture map.** The "first translation" claim has sprawled across ~15 issues + ~37 writer scripts + a doc, with repeated census attempts (now closed) and no single picture. This issue is the map: the layered architecture, which issue owns each layer, and the cleanup/merge actions. Not a deliverable itself — a coordination home so the cluster stops being re-derived.

## The layered architecture (bottom → top)

Each layer depends on the ones below. A "first translation" claim is a *per-work* question, so it bottoms out in identity.

### 1. Identity foundation
- **#2179** — author thesaurus (epic; canonical persons)
- **#2264** — work-identity resolver `(author_id, uniform_title) → work_id` (step 3 of #2179). *Precision-first: under-cluster over mis-cluster — a wrong merge fabricates a false "first translation."* The blocking primitive for every per-work question.
- **#2318** — work-identity layer so import dedup doesn't worsen at scale

### 2. Work-level catalog (the corpus frame + denominator)
- **#2453** — universal works catalog (pre-1900, all traditions): work identity + holdings + translation status. The *home* for the translation index; the reusable frame the one-off censuses kept rebuilding.
- **#2452** — Chinese mega-catalog → **subsumed by #2453** ("generalizes #2452"); merge-candidate.
- **#2447** — IIIF manifest census (Latin/Chinese/Sanskrit) — feeds the frame.

### 3. Translation knowledge (the growing positive registry + census)
- **#2352** — work-keyed translation index: flip first-translation from per-book *search* to a *membership test*. The growing, compounding registry (the flywheel); "denominator for free"; the automated setter #1974 lacks.
- **#2244** — backfill earlier published translations (`translations_found`) — feeds the index.
- **#2234** — Chinese census denominator + source-language provenance for `translation_catalogs`
- **#2555** — census expansion levers (Hebrew/NLI, Sanskrit/Chinese sources)
- **#2476** — `ustc_editions.has_english_translation` is provenance-free (evidence columns empty) — the *evidence-of-absence* bug in another table; fixed by the same attempt-log discipline.
- **#2395** — distinguish originals from translations (`text_role`); **#2184** — scanned translations mislabeled by language.

### 4. Flag mechanics: setter, cleanup, measurement, honesty
[...]

PR #2654, opened 2026-06-21
feat(shwep): chapter-aware deep links — cited works link to their part of a collected edition

JDerekLomas:
## What
Follow-up to #2615. A cited work that resolves to a **collected/omnibus edition** (Complete Works of Philo, Works of Plato, NPNF/ANF…) now deep-links to the **specific treatise's page** within that volume, instead of the 800-page front.

## How
The contained works are already in the TOC layer — `book.chapters[]` carries `{titleEn, pageId, pageNumber}` (generated by the enrich-worker chapter phase, served by the DTS navigation API). So this is a lookup, not a guess:
- For a collected-edition target, match the cited work to a chapter by **English-title token overlap** (conservative ≥0.8, no LLM — both sides carry English forms).
- On a confident match, link to `/book/<slug>/page/<chapter.pageId>`; otherwise fall back to the volume front (never a wrong-page link).

Applies to **inline links** and the **supplementary "Also in Source Library" cards** (which now carry an optional chapter `page`; the card URL deep-links when present).

## Result
17 page-precise links — e.g. *Cyropædia* → Xenophon *Opera*, Tertullian *Adv. Valentinianos*, Sextus *Contra mathematicos*, Porphyry *De antro*, Julian *Ep.* Each is a confident title match; the rest stay book-level. Verified: `book.chapters` "On the Cherubim and the Flaming Sword" (p65, pageId resolved) is exactly how *De Cherubim* resolves.

## Out of scope
The book-matches grid (non-inline episodes) stays book-level — it shows multiple editions per work, so per-edition page-precision is a separate increment.

## Verification
- `npx tsc --noEmit` clean
- Conservative matcher → no wrong-page links; fidelity of the linked bibliographies unchanged (links only gain a `/page/<id>` suffix).

Builds on the existing works/TOC layer — refs #2453, #2567, #2632, #2615.


JDerekLomas:
Superseded by a clean re-cut from main (the branch carried pre-squash history of #2615 → DIRTY). New PR incoming.

PR #2678, opened 2026-06-21
feat(email): Resend webhook + broadcast stats

JDerekLomas:
## What
Real delivery monitoring for the staged user broadcast (and all future Resend sends).

- **`/api/webhooks/resend`** — Svix-verified webhook that logs `email.{sent,delivered,bounced,complained,opened,clicked,failed}` events into a new `email_events` collection. Idempotent on `(email_id, type)` so Resend retries don't double-count.
- **`scripts/email-broadcast-stats.mjs`** — aggregates bounce / open / complaint / click rates per `broadcast_id`.

## Why
The user-welcome broadcast (`scripts/send-user-broadcast.mjs`) sends in 3 staged batches; we need Batch N's bounce/complaint rate before sending Batch N+1. Resend's broadcast API exposes no aggregate stats, so we capture events ourselves.

## Go-live (after merge + deploy)
1. `npm run deploy:prod` so the endpoint is live.
2. Set `RESEND_WEBHOOK_SECRET` in Vercel prod env (the secret Resend returns when the webhook is registered).
3. Register the webhook in Resend pointing at `https://sourcelibrary.org/api/webhooks/resend`.

## Scope
Out of scope: no UI; stats are CLI-only for now. Signature verification mirrors the existing Stripe webhook pattern.

### DIFF
--- a/scripts/email-broadcast-stats.mjs
+++ b/scripts/email-broadcast-stats.mjs
@@ -1,0 +1,52 @@
+/**
+ * Aggregate Resend email events (logged by /api/webhooks/resend) into
+ * bounce / open / complaint / click rates, per broadcast.
+ *
+ *   node scripts/email-broadcast-stats.mjs                 # all broadcasts, summary
+ *   node scripts/email-broadcast-stats.mjs <broadcast_id>  # one broadcast
+ *
+ * Source via: set -a; source .env.production.local; set +a
+ */
+
+import { getScriptClient } from './lib/mongo.mjs';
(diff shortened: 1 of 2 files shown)


JDerekLomas:
Closing — not needed. We set up the Resend tracking subdomain (ilove.sourcelibrary.org), which makes opens/clicks appear in the /emails API `last_event` directly, so the webhook + email_events store is redundant. Branch kept if we ever want real-time push.

Issue #2688, opened 2026-06-21
SEO: open demand-proven reader pages to indexing (read≥5 ∪ first-translations)


JDerekLomas:
## Why

Individual reader pages (`/book/<slug>/page/<pageId>`) are `noindex` today (`src/app/book/[id]/page/[pageId]/page.tsx:19`). That was a deliberate crawl-budget call — there are ~100× more page URLs than book URLs, and indexing millions of thin template pages risks soft-404/duplicate noise.

But the unique, irreplaceable content — our first English translations — lives on those page URLs. By noindexing them all and canonical-ing to the book, a reader Googling a phrase from page 89 of a Latin alchemical text never lands on our translation of it. We're hiding our most defensible content from search.

The fix is not a blanket flip. It's to **open up the demand-proven subset**: pages people actually read, plus the uncontested first-translations we uniquely own.

## Data (validates the approach)

From `analytics_pageviews` (human-only URL reads, ~90-day TTL), reader-page URLs:
- 93,222 distinct URLs read; 198,888 total reads
- read ≥3 → 26,925 pages · **read ≥5 → 3,452** · read ≥10 → 897
- Top pages are exactly the canon: Fludd *Macrocosm*, *Zohar*, *Key of Solomon*, etc.

## Eligibility (the union rule)

A page becomes indexable if it has substantial text (translation present, or meaningful OCR — not blank) **AND** either:
- **read ≥ 5** in the rolling 90-day window (proven demand — skews to famous works), **OR**
- the book is `is_first_translation` and the page is translated (`pages_translated > 0`) — uncontested long-tail we own outright.

Start at read ≥5 (~3.5K pages). Loosen to ≥3 if the readout is clean.

## Build

1. **`scripts/seo/flag-indexable-pages.mjs`** — aggregate `analytics_pageviews` (resolve `pageId` = last path segment, book-id-form agnostic), union with first-translation translated pages, gate on substantial text, write `seo_indexable: true` + `seo_url` (`/book/<slug>/page/<id>`) onto qualifying `pages` docs. Idempotent; `--clear` reverts; `--dry-run` sizes. Re-run weekly (Hetzner cron) so the rolling window stays fresh.
[...]

PR #2703, opened 2026-06-22
fix(seo): compound index so sitemap reader-page chunks aren't empty (#2688)

JDerekLomas:
Follow-up to #2691 — post-deploy verification caught the sitemap serving **empty page chunks** (`/sitemap/1000.xml` returned an empty `<urlset>`, and the index omitted the `1000+` chunks entirely).

## Root causes (both fixed)
1. **`getIndexablePages` timed out.** It sorts by `_id` for stable skip/limit pagination, but the single-field `{seo_indexable:1}` partial index made the planner full-scan the `pages` collection in `_id` order to satisfy the sort → `MaxTimeMSExpired` → `safeQuery` returns `[]` → empty chunk. Fixed with a **compound `{seo_indexable:1, _id:1}` partial index** (verified on prod: a skip-10k/limit-5k query went from timeout to **~7s**). Query timeout bumped to 60s for the deepest chunks.
2. **`generateSitemaps` silently dropped all page chunks.** Its page-count `safeQuery` falls back to `0` on timeout, which zeroes out the entire `1000+` chunk range. Atlas was slow during the last deploy (the build logs show the unrelated `entities` aggregate timing out too). Bumped its `maxTimeMS` 10s → 30s; the compound index makes the count ~50ms anyway.

## State
The compound index is **already created on prod**, so the runtime query is fixed now. This redeploys so the **static sitemap rebuilds** with the page chunks listed and populated.

The core robots gate (live since #2691) is unaffected — flagged pages already serve `index, follow`, unflagged serve `noindex, follow`, both verified in prod. This only fixes sitemap-driven discovery.

### DIFF
--- a/scripts/seo/flag-indexable-pages.mjs
+++ b/scripts/seo/flag-indexable-pages.mjs
@@ -61,6 +61,14 @@ const db = client.db('bookstore');
 const pages = db.collection('pages');
 const books = db.collection('books');
 
+// Partial index on the flag — without it, the reconcile `$nin` below and the
+// sitemap's count/find on seo_indexable full-scan the millions-row pages
+// collection. Idempotent; only indexes the ~few-thousand flagged docs.
+await pages.createIndex(
+  { seo_indexable: 1 },
+  { partialFilterExpression: { seo_indexable: true }, name: 'seo_indexable_partial' }
+);
+
 if (CLEAR) {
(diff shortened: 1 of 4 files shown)


JDerekLomas:
Superseded by a clean re-cut from main (the branch carried pre-squash feature commits → merge conflict).

Issue #2740, opened 2026-06-25
AI-native search: one recall-honest tool for agents (engineered against real diagnostic failure modes)


JDerekLomas:
## Why / who

`/api/search/unified` is built for **a human at a typeahead**: 8–12 results, UI buckets, thumbnails, blobs, display hints. The fast-growing *other* consumer — the AI librarian, external MCP clients, in-app AI (`/api/learn`, search-expand, voice), and Claude itself — wants none of that. We've been feeding agents a human's search, and it's the wrong meal.

The "user" of an AI-native search is **an agent acting on behalf of a researcher**. Its cardinal sin is the **confident false negative**: reporting "we don't have X" / "we hold 1" when the library actually holds the work, fully processed, under a different catalog name. That single failure mode strikes directly at the mission (*put primary sources in people's hands*) — an agent that hides what we hold is worse than no agent.

## The spec is the set of errors a real agent (me) made diagnosing this

This issue's requirements are grounded in actual mistakes made while auditing search in one session. Every one is a failure mode the AI search must engineer against. They share a single root: **concluding *absent / broken* from a partial or mis-shaped signal — one narrow lane, one ambiguous empty array, one unrepresentative probe.**

| # | Error made | Root | Requirement it implies |
|---|---|---|---|
| 1 | Saw `books:1` for "corpus hermeticum", nearly accepted the corpus was shallow — we hold ~27 Pimander/De-mysteriis editions under another name | Catalog-name sparsity read as collection sparsity | **Work-cluster + synonym bridge.** Never show a lonely `1`; show the work + its editions + *why* it matched ("via Pimander, semantic 0.78") |
| 2 | Claimed "the anon gate silently eats results" from my parser printing `[]` | Empty array treated as a fact, not a state | **No empty may look like a fact.** Every lane returns explicit `status: ok/empty/timeout/rate_limited/degraded` |
| 3 | Then *retracted* it ("it never silently empties, the gate shows a wall") | Checked only the gate path on `/unified`; missed that the **separate** semantic fetch on `/search` swallows errors (`no r.ok check`, `.catch(()=>[])`) → conceptual editions DO silently vanish | **A correction needs a representative probe too.** "It silently empties" must be checked at *every* lane that can empty |
[...]

Issue #2735, opened 2026-06-25
Unified search: artwork bucket is semantic-only + confirm the search UI is wired to /api/search/unified


JDerekLomas:
## Context
`/api/search/unified` already exists and returns **books + index + gallery + visual + semantic + artworks + collections** in one request — so cross-surface discovery is largely built. But verifying it on "tarot" surfaced two gaps worth closing.

## Gap 1 — artwork recall is semantic-only
The `artworks` bucket is fed by `semanticArtworkSearch` (embedding similarity), so an exact/term query under-returns: live `?q=tarot` → **artworks: 4**, while the library holds **~115 `content_type:'artwork'` records with "tarot" in the title** (all with real images). A user searching the literal word should see those cards.
- Fix: add a **title/keyword recall path for artworks** (substring/lexical match on artwork title/author), fused with the semantic results — same shape as the book lexical+semantic blend. So "tarot", "emblem", "ouroboros" surface the matching artwork, not just the 4 nearest by vector.

## Gap 2 — confirm the front-end is wired to /unified
Verify the main search box actually calls `/api/search/unified` (and renders the `gallery` + `artworks` buckets), **not** the books-only `/api/search`. If it's books-only, users searching "tarot" see ~6 books and never the ~115 card images — the content exists but is invisible to them.

## Acceptance
- A term that matches artwork by title (tarot, emblem, …) returns those artworks in the main search UI, alongside books.
- The search box demonstrably hits `/unified` and shows all relevant buckets.

## Not in scope
- Book OCR backlog / processing (separate; pipeline paused).
- Field-convention confusion that started this — already documented in CLAUDE.md (use `getBookThumbnailUrl()`, artworks legitimately have `pages_count:0`).

🤖 Drafted with Claude Code (from a search-discovery audit, 2026-06)

JDerekLomas:
## Refined diagnosis + motivating evidence — it's a *surfacing* problem, not a recall problem to build

While auditing, three cases where **we already own the work, fully processed, but it doesn't surface under the canonical name a user types:**

| User searches | We hold (processed) | Why missed |
|---|---|---|
| "corpus hermeticum" (#1 query) | 1516 + 1497 Aldine + ~27 readable editions | Catalogued as *Pimander / Pymander / De mysteriis* |
[...]

PR #2742, opened 2026-06-25
fix(search): artwork lexical recall + stop semantic lane silently reading as 'no results' (#2735)

JDerekLomas:
Closes part of #2735 (the artwork-recall gap + the silent-semantic findability bug). Companion issue for the agent-facing work: #2740.

### 1. Lexical recall for artworks in `/api/search/unified`
The `artworks` bucket was fed only by `semanticArtworkSearch` (vector similarity), so a literal term massively under-returned: `?q=tarot` surfaced **4** artworks while **~158** `content_type:'artwork'` records have "tarot" in the title (all with live thumbnails). The cards existed but were invisible.

- New `lexicalArtworkSearch` — title/author substring recall on `content_type:'artwork'` books, **fused ahead of** the semantic neighbours, deduped by `book_id`, tenant-scoped, sorted by `quality_score`.
- Raised the `/search` artwork display cap 3 → 8 so the matches actually surface alongside books.
- Verified on the preview: `?q=tarot` now returns **24 artworks** (was 4); thumbnails resolve 200.

### 2. Stop the semantic lane silently reading as "no results"
The `/search` page renders "Conceptual matches" from a **separate** `fetch('/api/search/semantic')` that had **no `r.ok` check** and `.catch(() => setSemanticResults([]))`. That endpoint has **no anon gate** and a **6 s** lane — so on any timeout/5xx the held editions catalogued under another name (Pimander ≈ Corpus Hermeticum) vanished with no signal. This is the actual cause of the "corpus hermeticum looks like 1 result" symptom in #2735 — a bug, not a missing-unified-wiring gap (the box already calls `/unified`).

- Throw on non-200 instead of parsing an error body into `[]`.
- Track `semanticDegraded`; keep the full "no results" empty-state from firing when the lane errored (an empty set from a failure is not a fact).
- Show a subtle "you may be seeing fewer matches than we hold" notice.

## Scope
- **In:** the two findability fixes above. `npx tsc --noEmit` clean.
- **Out:** the agent-facing search rework (MCP `search_library` is books-only, more limited than `/unified`) — specced separately as **#2740**. Re-OCR/processing backlog. The `/catalog/scholar` bare-`/api/search` surface.

## Verification
- Preview deploy: `/api/search/unified?q=tarot` → `artworks.total: 24`, `books.total: 6`.
- Lexical lane returns 24 tarot artworks, all thumbnails 200.
- No ranking-ladder change to the book lane; artwork lane is additive. (Eval-guard recommended before any book-ranking change — none here.)

### DIFF
--- a/src/app/api/search/unified/route.ts
+++ b/src/app/api/search/unified/route.ts
@@ -11,6 +11,7 @@ import { checkRateLimit, getClientIp } from '@/lib/rate-limit';
 import { anonSearchGate, SIGNIN_URL } from '@/lib/anon-gate';
 import { getTenantContextFromRequest } from '@/lib/tenant-context';
 import { CLIP_URL } from '@/lib/clip';
+import { getBookThumbnailUrl } from '@/lib/utils';
 
 const ENTITIES_SEARCH_INDEX = 'entities_search';
 const GALLERY_SEARCH_INDEX = 'gallery_search';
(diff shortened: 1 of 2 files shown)


JDerekLomas:
## Verification (preview `79g1pbwx4`)

**Artwork recall (`/api/search/unified`):**
| query | artworks (prod → preview) |
|---|---|
| tarot | 4 → **24** |
| emblem | — → **22** |
| ouroboros | — → **9** |

**No-flood / regression:**
- `quantum mechanics` → 3 artworks (pure semantic; no lexical title match — expected).
- `xyznonexistent` → 4 artworks / 6 semantic — **identical on prod**, i.e. pre-existing lowered-floor vector noise when there are zero keyword hits, *not* introduced here.

**Search eval (`scripts/eval/search-quality-eval.mjs`):**
- Prod baseline: **17/18**.
- This branch (preview): **17/18** — same single pre-existing failure ("Language filter: Sanskrit … found an English book"), unrelated to this PR. **No new regressions; book-ranking ladder unmoved.**

**Still owed (the semantic-degraded fix):** a failure-injection check — force `/api/search/semantic` to 500/timeout and assert the page shows the "fewer matches than we hold" notice and does *not* collapse to a "no results" empty state. Endpoint-level per-lane `status:timeout` is tracked separately in #2740. The peer is committing a canonical-held-works acceptance harness; I'll wire it in pre/post-merge when it lands.

PR #2766, opened 2026-06-26
fix(seo): hard 404 (not soft-404 200) for missing/hidden book + page URLs

JDerekLomas:
## Why
Follow-up to #2758. Nonexistent or hidden `/book/<id>` and `/book/<id>/page/<id>` URLs render the not-found UI but return **HTTP 200** — a soft-404 — because both routes have a `loading.tsx`: Next streams a 200 shell before the page body's `notFound()` runs, so the status is already committed. GSC flags these in the **Soft 404** bucket (~1.2k).

## Fix
Move the existence/hidden check into `generateMetadata`, which is awaited **before** the shell streams, so `notFound()` there sets a real 404 status.

- **`book/[id]/page.tsx`** — `notFound()` for genuinely-missing and hidden books (was 200 noindex metadata). `getBookForMetadata` already retries and rethrows transient DB errors (caught → generic 200 metadata), so this never 404s on a blip — only a real miss.
- **`book/[id]/page/[pageId]/layout.tsx`** — same for missing/wrong-tenant/hidden pages. `getPageData` now retries once and **propagates** a persistent DB error (→ generic 200 metadata) instead of swallowing it to `{null,null}`, preserving the blip-vs-miss distinction. Added `visible`/`hidden` to the metadata projection.

Pairs with the page-number redirect (#2758) to clear the Soft-404 bucket: numeric page URLs 301 to the real page; genuinely-dead URLs now hard-404.

## Verification (must confirm on preview before merge)
The `notFound()`-in-generateMetadata → 404 behavior needs a live check under Next 16 + Turbopack. Will verify on the preview that: missing book → 404, missing page → 404, hidden book → 404, valid book/page → 200, before merging.

### DIFF
--- a/src/app/book/[id]/page.tsx
+++ b/src/app/book/[id]/page.tsx
@@ -195,24 +195,21 @@ export async function generateMetadata({ params }: PageProps): Promise<Metadata>
   }
 
   if (!book) {
-    // Self-referential canonical (not the inherited root '/') so Google doesn't
-    // see this as a duplicate of the homepage while it's still 200/noindex.
-    return {
-      title: 'Book Not Found - Source Library',
-      robots: { index: false, follow: false },
-      alternates: { canonical: `/book/${id}` },
-    };
+    // Genuinely missing book → hard 404. Calling notFound() here in
+    // generateMetadata (awaited BEFORE the response shell streams) sets a real
+    // 404 status; the same call from the page body comes too late — loading.tsx
+    // has already flushed a 200 shell — and only soft-404s (200 + not-found UI).
+    // getBookForMetadata retries once and rethrows transient DB errors, which the
+    // catch above turns into generic 200 metadata, so we only reach here for a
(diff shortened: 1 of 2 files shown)


JDerekLomas:
Closing: the approach renders the not-found UI but can't set a 404 status. Both /book/[id] and /book/[id]/page/[pageId] have a loading.tsx, so Next streams a 200 shell before notFound() resolves (even from generateMetadata) — verified on the preview (missing book renders not-found UI but returns HTTP 200). The only reliable fix is removing those two loading.tsx files, which costs the loading skeleton on the two highest-traffic route families. Not worth it: the meaningful soft-404s (numeric /page/<n> URLs and the 29 no-slug books) are already fixed and live via #2758; the residual is genuinely-dead URLs where a soft-404 is benign (Google correctly declines to index them). See handoff 2026-06-26-seo-404-fixes.md.

PR #2767, opened 2026-06-26
fix(reading-history): record global (non-tenant) books (#2759)

### LINKED ISSUE #2759: P0: Reading history records nothing for global (non-tenant) books since ~2026-05-14
## Problem
Reading history records **nothing for any global (non-tenant) book** — i.e. the entire public catalog — and has since ~2026-05-14. Surfaced by founder QA: "I feel like not all of the books I've read are in my history, I was reading some Chinese book."

## Root cause
`src/app/api/reading-history/route.ts` bails out for any book without a `tenantId`:
```
let tenantId = getTenantContextFromRequest(request).id;
...
if (!tenantId) return; // unknown book/tenant — silent fail
```
Only tenant-promoted books (bph, etc.) have a `tenantId`. Every normal global book → `tenantId === null` → the write is silently skipped. The session-collapse `findOne` and all upserts are gated on `tenantId`, so global reads are never persisted.

## Fix
Mirror the likes fix (#2407): do **not** bail when `tenantId` is null. Resolve the book; if it exists, store the row with `tenantId: null` (and query existing rows with `tenantId: null`). Only bail if the book genuinely can't be resolved.

## Acceptance
- Reading a global book creates/updates a `reading_history` row for the user.
- `/reading-history` shows global books read after the fix.
- Tenant books still scope correctly.

_Reported via feedback widget 2026-06-13. Confirmed by feedback-triage swarm + direct code read._

JDerekLomas:
Closes #2759

## Problem
`POST /api/reading-history` bailed out for any book without a `tenantId` — which is every normal global book, i.e. the entire public catalog. Reading history has recorded **nothing** for global books since ~2026-05-14. Surfaced by founder QA ("I was reading some Chinese book… not in my history").

## Fix
Mirror the likes fix (#2407): resolve the book and store the row with `tenantId: null` when it's a global book. Only bail when the book genuinely can't be resolved. The session-collapse `findOne` now matches `tenantId: null` explicitly, and the `GET` path already shows global books on the canonical `/reading-history` page (no tenant header ⇒ no `tenantId` constraint).

## Scope
- In: the POST write path only.
- Out: no schema/index changes; existing tenant rows keep scoping correctly (tenant subdomains still filter by `tenantId`).

## Acceptance
- Reading a global book creates/updates a `reading_history` row.
- `/reading-history` shows global books read after the fix.
- Tenant books still scope correctly.

### DIFF
--- a/src/app/api/reading-history/route.ts
+++ b/src/app/api/reading-history/route.ts
@@ -38,6 +38,9 @@ export async function POST(request: NextRequest) {
       const db = await getDb();
 
       // Tenant from proxy header if available; otherwise look up the book.
+      // Global (non-tenant) books resolve to tenantId: null — they must still
+      // be recorded. Only bail when the book genuinely can't be resolved
+      // (mirrors the likes fix, #2407).
       let tenantId = getTenantContextFromRequest(request).id;
       let resolvedBookId = book_id;
       if (!tenantId) {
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Superseded: `main` already carries a functionally identical reading-history fix (resolve the book, store with `tenantId: null`, only bail when unresolvable) — landed inline via a parallel session. Closing to avoid a duplicate. The conflict on merge was just a comment difference.

PR #2768, opened 2026-06-26
fix(search): language filter no longer fails open; hidden artworks no longer 404 (#2760)

### LINKED ISSUE #2760: P0: Search correctness — language filter fails open + hidden artworks leak into 404s
### 1. Language filter fails OPEN (P0)
`/api/search` (and the MCP `search_translations` / `search_concept` tools) accept `languages` / `exclude_languages`, but when the computed allow-list is empty the `book_id` constraint is left unset, so the query returns the **whole corpus** instead of zero results. Reported by LLM/MCP users: "the languages and exclude_languages params appear silently ignored or applied very loosely."
- Root cause: `if (allowedBookIds.length > 0) { ...apply book_id $in... }` — empty list ⇒ no filter applied (fail-open).
- Fix: apply `book_id: { $in: allowedBookIds }` unconditionally (empty list ⇒ empty result), or short-circuit the lane to `[]`. Add a smoke test for a zero-match language.
- File: `src/app/api/search/route.ts` (~L311) + MCP search paths.

### 2. Search surfaces hidden artworks → click 404s (P0)
Semantic search returns artwork records that are `visible:false / hidden:true`; clicking them 404s. Reported: "i searched toltec. got one hit. when i clicked on it, error message" (book `art-eagle-relief-toltec-temple`).
- Root cause: the semantic artwork lane (`artwork_embeddings`) has no visibility column, and slug-enrichment doesn't re-check `visible:true` before returning hits.
- Fix: add `visible:true` to the enrichment `find`, drop ids not returned by it.
- File: `src/app/api/search/unified/route.ts` (~L353).

## Acceptance
- A zero-match language filter returns 0 results, not the corpus.
- Hidden/invisible artworks never appear in search results.

_Reported via feedback widget + MCP feedback. Confirmed by feedback-triage swarm._

JDerekLomas:
Closes #2760

Two confirmed search-correctness bugs from the feedback-triage swarm.

## 1. Language filter failed OPEN (P0)
The page-content lane in `/api/search` built an allow-list of book IDs from the language/category/date filters, then applied it only `if (allowedBookIds.length > 0)`. A zero-match language (or any filter resolving to no books) left the `book_id` constraint **unset**, so the query scanned the **whole corpus** instead of returning nothing — "languages and exclude_languages appear silently ignored."

**Fix:** short-circuit the page lane to `[]` on an empty allow-list, and harden the shared `buildPageSearchStage` so an empty array matches nothing (impossible `book_id` sentinel) instead of dropping the filter — this defends every caller, not just this route.

- `search_translations` (MCP) routes through `/api/search` → covered.
- `search_concept` (MCP) → `/api/search/semantic`, whose language filter is applied server-side in the Supabase RPC (language strings passed directly, `expandLanguages` always keeps the original term) — already correct, no fail-open.

## 2. Hidden artworks leaked into results → click 404s (P0)
The artwork slug-enrichment in `/api/search/unified` didn't re-check visibility, so `visible:false / hidden:true` artworks surfaced and 404'd on click (`art-eagle-relief-toltec-temple`).

**Fix:** gate the enrichment `find` on `visible:true` and drop any artwork not returned. Fail **closed** — if the lookup errors we can't confirm visibility, so the lane is dropped rather than risk a leak.

## Tests
`tests/unit/page-search-filter.test.ts` pins the empty-allow-list invariant on `buildPageSearchStage` (undefined ⇒ no filter; empty array ⇒ matches nothing; never fails open). 4 passing.

## Scope
- In: the keyword page lane (`/api/search`), the shared Atlas stage builder, the unified artwork lane.
- Out: no change to the semantic RPC (already correct); no re-indexing.

## Acceptance
- A zero-match language filter returns 0 results, not the corpus.
- Hidden/invisible artworks never appear in search results.


JDerekLomas:
Superseded in part: `main` already has the hidden-artwork visibility gate, and an inline fail-open fix (unconditional `$in: allowedBookIds`). BUT that inline fix is incomplete for the empty-allow-list case (`$in: []` is dropped by buildPageSearchStage → still leaks the whole corpus). I split that genuine residual out as #2836 rather than merge this now-conflicting branch. Closing in favor of #2836.

PR #2769, opened 2026-06-26
fix(ocr): read-time safety net — collapse lacuna dot-walls + strip LaTeX leakage (#2764)

### LINKED ISSUE #2764: OCR read-time safety net: collapse lacuna dot-walls + strip LaTeX leakage
## Problem
Two classes of OCR artifact reach readers/search/quotes as garbage text:
1. **Runaway dot/lacuna loops** — degenerate repetition where the model emits thousands of "." (papyrus/critical-edition lacunae). OCR prompt **v14** now prevents most at write-time, but legacy pages and any residual loop still need a read-time guard.
2. **LaTeX math leakage** — literal `$\frac{a}{b}$` etc. rendered raw to readers (~thousands of pages).

## Fix (read-time, zero re-OCR cost)
In the page text renderer (`NotesRenderer.tsx` / shared text-cleaning) add preprocessing, sequenced **after** `stripEditorialWrappers`:
- Collapse any run of >~12 repeated dot/dash/underscore chars (incl. spaced ". . . .") to a single `[…]` marker.
- Convert/clean stray `$...$` / `\frac{a}{b}` to readable form (or render via a math formatter).
- Apply to the snippet/quote surfaces too (see CLAUDE.md "Quote & snippet integrity" — every surface reads its own copy).

## Related (separate, owned elsewhere)
- Write-time prompt fix: shipped as OCR **v14** (lacuna + anti-repetition + margin rules; v12 archived).
- Page-level re-OCR of truly-broken legacy pages: owned by the OCR-reprocessing session (agent-comms).

## Acceptance
- No multi-hundred-dot wall renders in the reader.
- No raw `$\frac{}{}$` renders in the reader.
- Quote/snippet APIs return cleaned text.

_Derived from founder feedback ("ocr went nuts on the dots") + Treatise on Universal Harmony p.506 LaTeX leak. Confirmed by feedback-triage swarm._

JDerekLomas:
Closes #2764

Two classes of OCR artifact reached readers / search / quotes as garbage text. This adds a **read-time** safety net (zero re-OCR cost) plus the matching write-time **prompt-seed** sync.

## Read-time cleaners — new `src/lib/clean-ocr-artifacts.ts`
- `collapseLacunaWalls` — collapses a run of ≥12 repeated dot/ellipsis/middot/dash/underscore "units" (solid `............` **and** spaced `. . . .`) to a single `[…]` marker. Threshold 12 so ordinary ellipses/short rules are untouched; uses `[ \t]` between units so a wall never merges two lines.
- `latexToReadable` — `$\frac{a}{b}$` → `a/b`, `\sqrt{x}` → `√(x)`, common symbol/Greek commands → Unicode, strips inline-math `$…$` delimiters **only when the inner looks like math** (so historical currency `$5 and $10` survives). Unknown `\commands` are left intact rather than risk mangling text.

## Wiring (every surface reads its own copy — CLAUDE.md)
- **Reader** (`NotesRenderer`): lacuna collapse before parsing; LaTeX mop-up *after* the existing Greek/superscript/overline handlers so their nicer HTML output stands.
- **Snippet/quote/search/embedding**: folded into `stripEditorialWrappers` (after wrapper + markdown stripping), so all of `/api/search`, `/api/books/[id]/{search,quote}`, IIIF, likes, embeddings, and the tenant twins are covered in one place.

## Write-time seed sync (per issue comment)
Synced the OCR prompt **seed** `DEFAULT_PROMPTS.ocr` to the live DB **v14** — some batch OCR paths read the seed, not the DB: added the lacunae/anti-repetition block and the margin-transcription rule; bumped `PROMPT_VERSION`.

## Out of scope
Re-OCR of truly-broken legacy pages (owned by the OCR-reprocessing session); rendering math via a full formatter (KaTeX) — readable plain-text form is sufficient here.

## Tests
`tests/unit/clean-ocr-artifacts.test.ts` (14 passing) covers dot-walls (solid/spaced/dash/underscore), ellipsis/short-rule preservation, newline safety, frac/sqrt/symbol conversion, currency preservation, and the `stripEditorialWrappers` integration. Existing `strip-editorial-wrappers` suite still green (16). `tsc --noEmit` clean.

## Acceptance
- No multi-hundred-dot wall renders in the reader. ✓
- No raw `$\frac{}{}$` renders in the reader. ✓
- Quote/snippet APIs return cleaned text. ✓

### DIFF
--- a/src/components/reader/NotesRenderer.tsx
+++ b/src/components/reader/NotesRenderer.tsx
@@ -8,6 +8,7 @@ import rehypeRaw from 'rehype-raw';
 import { ChevronDown, ChevronRight, Info } from 'lucide-react';
 import { isRTLLanguage } from '@/lib/types';
 import { NOTE_TAG_STYLES } from '@/lib/style-constants';
+import { collapseLacunaWalls, latexToReadable } from '@/lib/clean-ocr-artifacts';
 
 // Page types where the entire "translation" is AI-generated description (no original text)
 const DESCRIPTION_ONLY_PAGE_TYPES = new Set(['blank', 'frontispiece', 'illustration', 'cover', 'map', 'diagram', 'musical-score', 'table']);
@@ -783,7 +784,9 @@ export default function NotesRenderer({ text, className = '', showMetadata = tru
   const isDescriptionOnly = DESCRIPTION_ONLY_PAGE_TYPES.has(pageType ?? '') ||
     DESCRIPTION_ONLY_PAGE_TYPES.has(metadata.pageType ?? '');
 
-  const withBracketTags = useMemo(() => preprocessBracketTags(cleanText, showNotes), [cleanText, showNotes]);
+  // Collapse runaway dot/lacuna walls before anything else parses the text (#2764).
+  const withLacuna = useMemo(() => collapseLacunaWalls(cleanText), [cleanText]);
+  const withBracketTags = useMemo(() => preprocessBracketTags(withLacuna, showNotes), [withLacuna, showNotes]);
   // Drop dangling vocabulary chips when notes are off (see preprocessTerms).
   const withTerms = useMemo(() => preprocessTerms(withBracketTags, showNotes), [withBracketTags, showNotes]);
   // On description-only pages, render the whole AI description uniformly (no half-highlighting).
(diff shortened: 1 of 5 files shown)


JDerekLomas:
Superseded: `main` already has the OCR read-time net — `cleanOcrArtifacts` in strip-editorial-wrappers (snippet/quote path) AND wired into NotesRenderer (reader), plus the OCR seed lacunae block — landed inline via a parallel session. Closing to avoid a duplicate. (Minor: main left `PROMPT_VERSION` at v6.1 despite the seed change; not worth a PR.)

PR #2787, opened 2026-06-27
docs(ft): a history of the first-translation attempts

JDerekLomas:
Narrative retrospective the team asked for — how first-translation ID went from a one-line heuristic to a per-book cross-model census, and the seven months of wrong turns between. Six eras, the recurring failure modes, and the Arithmologia near-miss. Doc-only; companion to the SoT and the census methodology doc. Refs #2564 · #2567.


JDerekLomas:
Superseded by #2792 (sourced rewrite with provenance + uncertainty section).

Issue #2800, opened 2026-06-27
First-translation: inventory of SAVED verification evidence + the reuse gap (matching, not discovery)


JDerekLomas:
Maps everything we've **already saved** from prior first-translation verification work, so workers stop re-discovering (and re-paying) for it. Part of #2780 / cluster #2567. Confirms the history doc's finding #4: *"matching/reconcile failure, not a seeding gap — we hold the answer and don't match/reconcile it."*

## What is saved (live counts, bookstore, 2026-06-27)

**Durable registries (Mongo):**
- `translation_catalogs` — **24,040** scholarly translation records, with normalized match fields (`canonical_author_normalized`, `author_surname`, `canonical_work_normalized`, `english_title_normalized`, `translator`, `pub_year_int`, `completeness`, `source_language`). Sources: unesco_master 7,542 · loc_marc 5,689 · openlibrary 4,442 · extended_unesco 3,191 · hathitrust 1,297 · validated_additions 401 · **ft_verification_discovery 302** (Sink C harvest) · loeb 184 · cambridge 62 · cua_paulist 94.
- `first_translation_attempts` — **2,831** append-only rigorous attempts over **2,581 distinct books** (2,397 with `queries[]`, 1,756 with real `sources_checked[]` URLs, 205 with structured `priors[]`). gemini_verifier 2,321 · tier2_agent 496 · claude pilot 14.
- `latin_translations_census` — 7,542 · `translation_classification` — 9,906 · `external_catalog` — ~250K · `catalog_coverage` — **~2.4M** · `catalog_coverage_snapshots` — 79 · `ft_reverify_proposal` — 2,716 · `kloss_catalog` — 1,614 · `translation_backup_683a_2026_05_31` — 1,036.

**On `books`:**
- **2,983** books carry saved prior-translation citations (`translation_verification.translations_found[]`).
- 12,557 have a `disposition`; 4,244 have a real `tools_called[]` trail; **15,285** have an `ai_metadata.first_translation` content opinion.
- 545 carry a graded `first_translation.verdict` (237 first_no_prior, 131 not_applicable, 125 not_first, 52 first_complete).

**On disk:** `scripts/eval/results/ft-*.json` (tier2-verdicts-final 332K, recall worklists, sample manifests); `scripts/output/ft-evidence-2026-06-{21,26}/` (promote-346 + demote-145 results, backups, signoff diffs).

## The reuse gap (where the saved work isn't connected)

[...]

Issue #2780, opened 2026-06-26
First-translation census: verify all 14,106 eligible books (tiered, per-book)


JDerekLomas:
Standing plan to convert the first-translation flag from a ~53%-reliable legacy disposition into a **per-book, evidence-backed census** across the whole corpus. Methodology doc: `.claude/docs/first-translation-census-methodology.md` (PR #2779). Part of #2564 / cluster #2567.

## The gap
- **15,809** readable books; **14,106** FT-eligible (non-English original); **5,570** currently badged; **only 181 (~1.3%) rigorously verified.**
- Estimated true first English translations: **~5,000 (range 4,000–6,500)** — close to today's badge count but *different books* (~1,900 over-claims + ~1,900 missed firsts).

## Why tiered + cross-model (proven)
- Single-pass Gemini fabricates ~63% of "a prior exists" claims → can't drive demotes alone.
- A blanket reconcile `--apply` would have wrongly stripped **39%** of its demotion candidates (measured on the 145 run) — a census, not a sample, surfaced that.

## Execution checklist
- [x] Tier-2 instrument validated (ft-verify skill, Claude Sonnet subagents)
- [x] Demote census — 145 candidates (56 kept / 89 demoted), applied + provenance
- [ ] Promote census — 346 high-prior unbadged (in progress) → apply survivors
- [ ] Stage A/B: structural pre-filter + Tier-0 work_id/catalog linkage (free) — needs #2318 progress
- [ ] Stage C: Gemini Tier-1 sweep over remaining eligible
- [ ] Stage D: Tier-2 census on the **consequential set** (~7,500: badged 5,570 + high-prior unbadged ~2,000), in signed background batches
- [ ] Stage E: schedule the single-writer reconcile (Hetzner cron) so the flag stays materialized
- [ ] Stage E: retire the ~8 active legacy setters of `is_first_translation`
- [ ] Stage F: statistical closure (Wilson/FPC CI) for non-censused strata
- [ ] Product decision: switch the public headline to `isPublicFirst` (excludes weak) — lowers the number to the defensible set

## Cost
Tier-2 ≈ 58k tokens/book. Consequential set (~7,500) ≈ ~440M tokens / ~1–2 weeks of background batches; full eligible (14,106) ≈ ~820M tokens. Each paid/large run gated on cost sign-off.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JDerekLomas:
[...]

PR #2801, opened 2026-06-27
fix(ft): honest evidence panel for unverified legacy badges (#2780)

JDerekLomas:
~5,534 books carry a public **First Translation** badge whose only evidence is a legacy disposition with no graded verdict and no recorded search trail. The evidence panel rendered **nothing** in that case — so the reader saw a bald conclusion (often with a reasoning line like *"No results from any of the 3 library catalogs queried"*) that **implied a thorough search had happened**. Example: Comparetti, *Papiro ercolanese inedito* (id `69aea4993555bdbfd93e4026`).

This makes the panel's honesty match the verification — principle #1 of the system (*a claim's strength must equal its verification*) made visible to the reader.

**Changes** (`FirstTranslationEvidence.tsx`, render-only, no data writes):
- `EvidenceFooter`: when there is no grounded attempt **and** no tool trail, render *"Preliminary determination — automated catalogue check only; detailed verification pending"* instead of rendering nothing.
- `StrengthChip`: surface confidence at a glance. `strong`/`moderate` now carry tooltips distinguishing cross-model from single-search; weak/legacy claims show a muted **preliminary** chip so they don't borrow the authority of a verified one.

**Why ship now:** this stops the over-claim immediately for all ~5,534 books at zero cost, while the census (#2780) backfills real attempts over time. It does **not** touch the headline count (`isPublicFirst` switch is a separate product decision) — purely the per-book panel.

Refs #2780 · #2564. Pairs with the saved-evidence reuse work in #2800.

### DIFF
--- a/src/components/book/FirstTranslationEvidence.tsx
+++ b/src/components/book/FirstTranslationEvidence.tsx
@@ -22,6 +22,10 @@ import {
   type FirstTranslationAttempt,
 } from '@/lib/first-translation/attempt-log';
 import {
+  summarizePriorEvidence,
+  type PriorEvidenceSummary,
+} from '@/lib/first-translation/prior-evidence';
+import {
   FIRST_FAMILY,
   DISPOSITION_TO_VERDICT,
   VERDICT_TO_DISPOSITION,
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Still valuable — directly serves the 'reader sees the evidence; claim strength must equal verification' principle, and the legacy-badge honesty it adds is consistent with the three-number reality (claimed ~5,818 / evidence-applied ~6,283 / defensible ~3,033). Recommend review + merge. (FT triage.)

PR #2812, opened 2026-06-27
docs(memory): lesson — grounded-LLM verdict ≠ reproducible audit trail (FT badges)

JDerekLomas:
Records the root-cause finding behind "~94% of First-Translation badges have no documented search trail."

## The finding (not a logging regression — an architecture property)
- The bulk FT verifier (`discover-prior-translations.mjs`) used **Gemini + Google Search grounding**, which returns the answer + cited source URLs (`groundingChunks`) but **never the literal queries** it issued.
- The prompt asked only for `{translations[], found_complete_prior_translation, reasoning}`; the apply path's `$set` dropped even `reasoning`/`groundingUrls`, keeping `disposition` + found priors + `tools_called`.
- The #2802 backfill faithfully migrated those into `first_translation_attempts` (`queries: [], // legacy did not store queries`) → 18,010 conclusion-only `gemini_verifier` rows. Only **14** books have the gold-standard `claude_subagent_verify` trail.

## Data (the over-claims question)
Of **2,293** live FT badges that also carry a `result:'found'` verdict: **2,112 weak**, 57 moderate, **124 strong** — and **0 of the strong have a reproducible trail**. So even the strong ones need ft-verify before any demotion; the weak tier is the one that already caused false demotions (see `lesson_derive_from_unverified_source`).

Documentation-only (one bullet in `memory/lessons-learned.md`). No code.


JDerekLomas:
Recommend merge — the lesson (a grounded-LLM verdict is not a reproducible audit trail) is exactly the principle the evidence ledger (#2865/#2866/#2868) and the derivation (#2871) now operationalize. Good knowledge to land. (FT triage.)

PR #2831, opened 2026-06-27
feat(ft): rater-reliability + validity instrument (#2780)

JDerekLomas:
A standing **convergence & accuracy dashboard** for the first-translation census — the instrument Derek asked for to measure how much the verification approaches agree, and how accurate they are against checkable truth. `scripts/analysis/ft-rater-reliability.mjs`, read-only, free (no LLM).

Two measurements, kept separate (high agreement under shared bias is the trap):

### 1. Reliability — Cohen's κ between rater families
Independence is **by family**, not raw method (the same model resampled shares its blind spot). First run:

| pair | n | κ |
|---|---|---|
| catalog_legacy × model_legacy | 10,915 | 0.44 |
| model_legacy × structured_legacy | 1,759 | 0.39 |
| gemini_grounded × catalog_legacy | 1,062 | 0.36 |
| gemini_grounded × model_legacy | 1,151 | **0.16** |

All only *fair* (Landis–Koch) — the verdicts are noisy and approach-dependent, which is exactly why a single weak verdict can't be trusted and cross-family checking matters. **The two strong independent families (Claude agent × Gemini grounded) have too little overlap to score** — the instrument flags it and recommends a shared-sample experiment.

### 2. Validity — accuracy vs a one-sided gold set
A prior found **with a resolvable URL** ⇒ checkable truth ("not first"); absence never is (principle #1). Gold set = 2,151 books. Per-family recall reported (with the selection-bias caveat noted in-report).

### 3. The gap Derek flagged, quantified
**358 books are badged "First Translation" while we already hold a checkable contradicting prior** — the catalog-match the badge-setting process never performed (the cron hit Open Library / Google Books / Internet Archive, never our own 28k `translation_catalogs`). All 358 are **legacy-cited (0 cross-checked by a rigorous family)**, so they're a **prioritized verify queue**, not clean demotes — each needs the evidence-quality guard (the Arithmologia rule). Full list with cited prior + URL written to `scripts/output/` (gitignored).

Going forward this runs every batch so the convergence/validity numbers — and the "priors we hold but haven't matched" gap — stay visible and can't silently reopen.

Refs #2780 · #2567. Pairs with the saved-evidence inventory #2800.

### DIFF
--- a/scripts/analysis/ft-rater-reliability.mjs
+++ b/scripts/analysis/ft-rater-reliability.mjs
@@ -1,0 +1,167 @@
+/**
+ * Inter-rater reliability + validity instrument for the first-translation census
+ * (#2780). The standing "convergence & accuracy dashboard" the steady-state
+ * design needs: every batch, report how much the verification approaches AGREE
+ * (chance-corrected, by family) and how ACCURATE they are against checkable
+ * ground truth — plus the over-claims the catalog already disproves.
+ *
+ * Two distinct measurements, deliberately kept apart (they answer different
+ * questions; high agreement under shared bias is the trap):
+ *   1. RELIABILITY — Cohen's kappa between rater FAMILIES on books both rated.
+ *      Independence is by family, not raw method (same model resampled shares
+ *      its blind spot). Low kappa => the verdicts are noisy & approach-dependent
+ *      => a single weak verdict can't be trusted; cross-family checking matters.
+ *   2. VALIDITY — accuracy against a one-sided GOLD set: a prior translation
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Still valuable, but overlaps the new eval harness #2876 (precision/recall/agreement from existing data). Recommend **folding the rater-reliability / κ instrument into #2876** rather than running two measurement tools — cross-family κ is one of the metrics #2876 needs. Whoever takes #2876 should pull this in. (FT triage.)

PR #2834, opened 2026-06-27
provenance: authenticate the Trithemian watermark + make it a readable colophon

JDerekLomas:
## Why

The invisible ZWC imprimatur (`src/lib/steganographia.ts`) is the text-side provenance mark woven into translations as they leave the system (quote API, downloads/EPUB). Two problems, addressed together because both rework the payload:

1. **It was forgeable.** Payload was raw `bookId[:8]` bytes with no integrity tag; `readProvenance()` decoded with no key — it confirmed a mark was *present*, never that it was *ours*. Since the scheme and the edition ids (in URLs) are public, anyone could forge a "made by Source Library" mark.
2. **It was a covert token, not in the project's spirit.** The image layer already embeds a plain-language message for LLMs. The text mark should be the same — a **colophon**, a message in a bottle for whoever (human or bot) decodes it.

## What changed

**Authentication (security)**
- Framed, MAC-tagged payload: `[VERSION][idLen][id][msgLen][message][MAC]`, MAC = `HMAC-SHA256(key, version‖id‖message)[:6]` (48-bit; blind-forgery ~1-in-2.8e14, uncomputable without `PROVENANCE_SECRET_KEY`). Binds the message too, so a swapped colophon fails to verify.
- `verifyProvenance(text, key) → { editionId, message, authentic }` with a timing-safe compare; `verifyExport(text)` wraps it with the env key for server-side attribution checks.
- `readProvenance()` kept as explicitly **unauthenticated** presence-only, reading both the framed format and the **legacy raw-bytes format** so old exports stay attributable.

**Two-tier colophon (the message)** — mirrors the image layer (every page watermarked; ~1/10 also gets the visible logo):
- **Every export carries a *light* authenticated mark** — just the keyed edition id, ~330 invisible chars, always attributable/verifiable.
- **~1-in-6 also carries the *readable* colophon** — `Source Library · sourcelibrary.org/book/<id> · from humanists to all the newest minds · CC BY-SA 4.0`, with a fuller paragraph rarer still (~2% of pages). The link uses `book.id`, the canonical `/book/[id]` route param (verified), so it resolves back to the original.
- So provenance is **always present and verifiable**, but the readable prose — and its zero-width bulk — rides along only occasionally.

Measured over a 3k-page sample: **0 clean, ~83% light (avg 326 hidden chars), ~17% readable (avg 2,836; ~2% full paragraph), 0 authentication failures.**

## Honesty about the threat model (documented in-file)

This is **attribution, not DRM.** Zero-width characters strip with a one-line regex (and many platforms normalize them away). The MAC makes the mark **unforgeable**, not **unremovable** — deliberately: the goal is that whoever cracks it finds a friendly colophon, and that we can *prove* a present mark is genuinely ours. Strip-proof attribution still exists separately via per-page `content_hash` corpus matching and the image-side luminance watermark.

## Scope / risk

- Marks are applied **at export time and never stored** → no corpus data changes, **no migration**. New exports carry the mark; old in-the-wild exports still read via the legacy path.
- `markForExport` signature unchanged; quote + download call sites unaffected.
- `COLOPHON_RATE` (=6) is the single tuning lever for how often the readable note appears. Redundancy is 2 copies.
- **Out of scope (deliberately):** no public/admin decoder page this round (`verifyExport` is ready to power one); no change to image-side provenance.

## Commit history note

This branch went through a design reconciliation: an intermediate commit made the mark fully *occasional* (~83% of pages completely unmarked) for lightness, which conflicted with the colophon's purpose (a message must be present to be found). The final two-tier scheme keeps provenance on **every** page (light) and gates only the **readable prose** to ~1-in-6.

## Tests

`tests/unit/steganographia.test.ts`: round-trip + authenticate, wrong-key rejection, **forgery rejection**, **message MAC-binding**, colophon round-trip, excerpt survival, legacy compatibility. `npx tsc --noEmit` clean; full unit suite green (491).

### DIFF
--- a/src/lib/provenance.ts
+++ b/src/lib/provenance.ts
@@ -22,6 +22,13 @@ const PROVENANCE_KEY = process.env.PROVENANCE_SECRET_KEY;
 const TAGLINE = 'from humanists to all the newest minds';
 const LICENSE = 'CC BY-SA 4.0';
 
+// Occasional by design. Only ~1 in MARK_RATE exports carries a mark at all.
+// Most pages go out completely clean — lighter payloads and far less zero-width
+// noise for machine readers (LLMs that have to strip it) — while anyone pulling
+// many pages still hits marked ones, where the keyed-MAC colophon authenticates
+// the source. Mirrors the image layer's occasional (~1/10) visible logo.
+const MARK_RATE = 6;
+
 /**
  * The readable colophon woven into the mark. Compact by default; an
  * occasional fuller note addressed directly to whoever found it.
@@ -50,6 +57,17 @@ function wantsFullColophon(text: string): boolean {
 }
 
 /**
+ * Deterministically decide whether this export is marked at all (~1 in
+ * MARK_RATE). Keyed on the content with a domain-separated input so the choice
+ * is stable per passage, independent of the full-colophon roll, and
+ * unpredictable without the key.
+ */
+function wantsMark(text: string): boolean {
+  if (!PROVENANCE_KEY) return false;
+  return crypto.createHmac('sha256', PROVENANCE_KEY).update('imprimatur-gate:' + text).digest()[0] % MARK_RATE === 0;
+}
+
+/**
  * Mark text for export with an invisible provenance imprimatur.
  *
  * @param text - Raw translation or OCR text from the database
@@ -58,6 +76,12 @@ function wantsFullColophon(text: string): boolean {
  */
 export function markForExport(text: string, bookId: string): string {
   if (!PROVENANCE_KEY || !text) return text;
+
+  // Occasional: most exports go out unmarked (see wantsMark). Keeps the common
+  // case clean for readers and machine consumers; provenance still rides along
+  // on a deterministic ~1-in-MARK_RATE subset of pages, fully authenticated
+  // when present.
+  if (!wantsMark(text)) return text;
 
   try {
     // Use first 8 chars of bookId as the structured edition identifier;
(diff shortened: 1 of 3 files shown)


JDerekLomas:
Added a commit (e3b1a257) reconciling this PR with Derek's direction to make the watermark **occasional and lighter** for machine readers.

**What changed:** an outer `wantsMark(text)` gate in `markForExport` — a deterministic ~1-in-6 roll, keyed on content with a domain-separated HMAC input (independent of the existing full-colophon roll). Most exports now go out unmarked.

**Effect** (measured over 5k sample texts):
- ~83% of pages now go out **completely clean** (no zero-width chars at all)
- ~17% carry the keyed-MAC authenticated colophon
- the heavy full paragraph drops from ~12.5% → **~2.1%** of pages

**Preserved from this PR:** the keyed-MAC authentication and 2-copy redundancy are untouched — this only changes *how often* a mark is applied, not its contents or its security. A scraper pulling many pages still hits marked, verifiable pages, so book/export-level attribution coverage holds; the tradeoff is that an individual short quoted passage usually won't carry a mark (acceptable per the 'occasional + light' goal).

tsc clean; the 13 existing steganographia tests still pass (they exercise `imprimatur`/`verifyProvenance` directly, which are unchanged). `MARK_RATE` is a single constant if you want to tune the frequency.

cc @JDerekLomas — this is the 'authenticated but occasional' hybrid you picked.

PR #2838, opened 2026-06-28
feat(ft): implement Tier-0 deterministic catalog/work_id match (#2780)

JDerekLomas:
Implements the **Tier-0** step that `resolve.ts` has *specified* since #2564 but never had: a free, deterministic match of every book against our own `translation_catalogs` registry + `work_id` cluster, run before any paid search.

**Why:** the badge-setting cron only ever queried Open Library / Google Books / Internet Archive — never our own 28k-row registry — so books get badged "first" while we already hold a checkable contradicting prior (the reliability instrument, PR #2831, found 358 such cases). Tier-0 is the step that closes that gap, for free, on every book. Derek flagged this directly.

### What's here
- **`src/lib/first-translation/tier0-catalog.ts`** — `makeTier0Catalog(lookup)`, a pure `Tier` (I/O injected, per the resolve.ts contract). The asymmetry of principle #1 is wired in:
  - a FOUND prior **terminates** the cascade (`not_first` / strong) **only** when it is **complete**, **same source-language**, and passes the evidence-quality guard (#2579) — the Iamblichus (different-source-language) and Arithmologia (anthology/self-match) rules, in code.
  - a catalog **MISS** is weak absence (our registry is incomplete) and **never terminates** — it escalates to the paid tiers.
- **`tests/unit/first-translation-tier0-catalog.test.ts`** — 10 cases (decisive prior, partial-only, different-source-language, self-match, no-match, work_id match). All green.
- **`scripts/maintenance/ft-tier0-resolve.ts`** — dry-run runner; `--apply` writes the **verdict only** (never the flag), so the single-writer reconcile materializes the demote afterward (#2564 invariant preserved).

### Live dry-run
5,772 badged books × 399 complete catalog rows → **1 decisive prior held today** (Plato *Axiochus* → Spenser 1592). Low yield is honest: the catalog is Latin-heavy and mostly `completeness: unknown`, so **catalog enrichment (`source_language` / `completeness`) is the lever that makes Tier-0 bite** — tracked separately. The architecture is correct and runs first/free; yield scales with the registry.

### Out of scope (deliberate)
- Wiring Tier-0 into a production `defaultTiers()` cascade alongside Tier-1/Tier-2 — this lands the implementation + a standalone runner; the full cascade wiring is a follow-up once Tier-1/Tier-2 are also wired.
- Catalog enrichment (the yield lever).

Free, no LLM. `tsc` + unit tests green. Refs #2780 · #2564 · #2567.

### DIFF
--- a/scripts/maintenance/ft-tier0-resolve.ts
+++ b/scripts/maintenance/ft-tier0-resolve.ts
@@ -1,0 +1,119 @@
+/**
+ * Tier-0 catalog resolver — runs the deterministic catalog/work_id match
+ * (`src/lib/first-translation/tier0-catalog.ts`) over the badged set and reports
+ * the books for which we ALREADY hold a complete, same-language, guard-passing
+ * prior English translation. These are over-claims the badge-setting cron never
+ * caught because it never queried our own `translation_catalogs`.
+ *
+ * Dry-run by default. With --apply it writes ONLY the verdict (book.first_translation
+ * = not_first) + appends the attempt — it does NOT flip is_first_translation. The
+ * single-writer reconcile materializes the demote afterwards:
+ *   npx tsx scripts/maintenance/reconcile-first-translation-flag.ts --apply
+ * (gate on Derek's sign-off). This keeps the single-writer invariant (#2564).
+ *
+ * Only `completeness: complete` catalog rows can be decisive, so we load those
+ * once (~hundreds) and match in memory — no per-book DB round-trip.
+ *
+ * Usage:
+ *   set -a; source .env.production.local; set +a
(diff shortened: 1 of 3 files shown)


JDerekLomas:
Still relevant — `tier0-catalog.ts` is not on main. This is the deterministic Tier-0 lib the cascade needs; complements the report-only matcher (`ft-catalog-match.mjs`, now on main + extended in #2866) and the read-before-spend reuse (#2868). Recommend review + merge as part of wiring `resolve.ts` — and validate its precision against #2876 before it drives any flag. (FT triage.)

PR #2956, opened 2026-07-03
feat(admin): browser-proxied image ingest for Cloudflare-walled IIIF sources

JDerekLomas:
## What
Adds `POST /api/admin/ingest-image` — a small endpoint that lets a logged-in operator mirror page images into R2 from IIIF sources that block server-side fetches.

## Why
Gallica/BnF now 403s **all** datacenter fetches at the Cloudflare edge (curl, WebFetch, `/api/import/gallica` on Vercel, and the Hetzner archiver). Only a real browser holds the Cloudflare clearance. This endpoint closes that gap: an in-browser loop fetches each page from the source origin and POSTs the bytes here; the route resizes (display 1600 / thumb 400 / archival 2400) and uploads to R2 server-side, then wires `archived_photo`/`display_photo`/`thumbnail` onto the page doc.

## Proven
Used it to mirror all **145 folios of BnF Arabe 3544** (al-Badrī's *Rāḥat al-arwāḥ*, the PD manuscript) into R2 through the browser — the source is otherwise unreachable to our infra.

## Security
- CORS-open (ACAO `*`) **because** auth is a bearer token, not a cookie — a cross-origin browser POST needs no credentials, and the token gates every write.
- Prefers a dedicated `INGEST_TOKEN` (so operators never expose `CRON_SECRET` in a browser call); falls back to `CRON_SECRET` for server-side use.
- **Inert (503) unless a token is configured** — off by default.

## Scope / out of scope
- Reusable for future Cloudflare-walled IIIF sources (al-Zarkashī, other BnF manuscripts).
- Does NOT auto-run anything; it's operator-driven, one page per request.

### DIFF
--- a/src/app/api/admin/ingest-image/route.ts
+++ b/src/app/api/admin/ingest-image/route.ts
@@ -1,0 +1,106 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3';
+import sharp from 'sharp';
+import { getDb } from '@/lib/mongodb';
+
+export const runtime = 'nodejs';
+export const maxDuration = 60;
+
+/**
+ * Browser-proxied image ingest — mirror one page image into R2 and wire it onto
+ * the page doc. Built for sources whose IIIF images are reachable only from a
+ * real browser (Gallica/BnF blocks datacenter fetches at the Cloudflare edge, so
+ * /api/import/gallica and the Hetzner archiver both 403). A logged-in operator
+ * runs an in-browser loop that fetches each page from the source origin (where it
+ * holds Cloudflare clearance) and POSTs the bytes here; this route resizes and
+ * uploads to R2 server-side, sidestepping both the source block and R2 CORS.
+ *
+ * POST /api/admin/ingest-image?bookId=<id>&page=<n>
+ *   Authorization: Bearer <CRON_SECRET>
+ *   body: raw JPEG/PNG bytes (Content-Type: image/jpeg)
+ *
+ * CORS is open (ACAO *) because auth is a bearer token, not a cookie — so a
+ * cross-origin browser POST needs no credentials. The token gates all writes.
+ */
+
+const CORS: Record<string, string> = {
+  'Access-Control-Allow-Origin': '*',
+  'Access-Control-Allow-Methods': 'POST, OPTIONS',
+  'Access-Control-Allow-Headers': 'Authorization, Content-Type',
+  'Access-Control-Max-Age': '86400',
+};
+
+export async function OPTIONS() {
+  return new NextResponse(null, { status: 204, headers: CORS });
+}
+
+export async function POST(req: NextRequest) {
+  const secret = process.env.CRON_SECRET;
+  if (!secret || req.headers.get('authorization') !== `Bearer ${secret}`) {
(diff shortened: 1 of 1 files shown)


JDerekLomas:
**Review 2026-07-28 — holding this out of a batch-merge. The route needs input validation on `bookId` before it ships.**

I pulled this into a cleanup sweep of long-open PRs and stopped on it. The design (bearer token, open CORS because auth isn't cookie-based, browser holds the Cloudflare clearance) is sound and well-documented. The problem is downstream of auth.

`bookId` comes straight off the query string and is checked only for truthiness, then interpolated directly into three R2 keys:

```
archived/${bookId}/${page}.jpg
pages/${bookId}/${padded}.jpg
pages/${bookId}/${padded}-thumb.jpg
```

Three consequences:

**1. A wrong `bookId` writes R2 objects and reports success.** The Mongo update is `{ book_id: bookId, page_number: page }`, so a bogus id yields `matchedCount: 0` — but the three `PutObjectCommand`s have already run. The response is `ok: true, matched: 0`. That is the #3362 shape (`archived/undefined/N.jpg`) with a different cause: an unvalidated identifier producing a junk key prefix that nothing later reconciles. It also feeds #3005 — orphaned R2 objects are already a ~2-3 TB problem, and this creates them silently.

**2. The key is constructible.** Nothing rejects `/` in `bookId`, so a value like `x/../../pages/<other-book-id>/0001` composes into a key that lands under a *different* book's prefix. R2 keys are flat strings, so `..` isn't traversal — it's worse, it's a literal key that can collide with real page images. An operator typo or a mistyped loop bound could overwrite another book's scans.

**3. It overwrites `archived_photo` unconditionally.** For a page that has no image, that's the whole point. But there's no guard preventing it from running against a page that already has one, and per the paired-artifacts invariant in `CLAUDE.md`, overwriting an archived image without an alignment check is the #3368 / #3186 family.

Suggested minimum before merge:

- Validate `bookId` against a strict pattern (24-hex Mongo id, or whatever `books.id` guarantees) and 400 on anything else.
- Look the book up **before** the first R2 write; 404 if absent.
- Treat `matchedCount === 0` as a failure, not an `ok: true` — and ideally don't leave the R2 objects behind when it happens.
- Consider refusing a page that already has `archived_photo` unless an explicit `?overwrite=1` is passed.

Everything else here I'd merge happily. Not blocking on style — just don't want a new authenticated write path landing with the exact footgun that cost us #3362.

JDerekLomas:
Backlog sweep 2026-09-12: left **blocked** (security + blocked labels already on it). Genuinely unlanded — no `ingest-image` route, no `INGEST_TOKEN` anywhere on main — and the auth regime moved under it (#4470 fail-closed revalidation, #4366 key limits). What Derek must decide: whether a wildcard-CORS, shared-bearer-token, browser-callable write endpoint (R2 + Mongo page mutation) is acceptable at all, or whether it must go through session auth like other admin routes. Ten weeks open.

JDerekLomas:
Closing (backlog decision 2026-09-22): a wildcard-CORS, shared-bearer-token, browser-callable endpoint that writes R2 and mutates page rows is not an acceptable shape for an admin write path; if the browser-proxied ingest is still needed for Cloudflare-walled IIIF sources, redo it behind session auth like the other admin routes. Branch kept.

Issue #3005, opened 2026-07-05
R2 storage reclamation: ~2–3 TB recoverable (display bloat, dup full-res, JP2 orphans, stale crops)


JDerekLomas:
Findings from a sampled audit (2026-07-05) of the 16.4 TB bucket, following up the size breakdown in #2972. Good news first: the feared wholesale `archived/` ↔ `pages/` duplication does **not** exist — `pages/` is mostly legitimate derivatives (display ~430 KB avg + 5 KB thumbs), and both prefixes belong ~100% to live books.

### 1. Display images that were never downsized — ~1–1.5 TB, also a UX win
**34% of display `.jpg`s are ≥90% the size of their full-res archival master** (probe: 101/299, avg master 1,007 KB). The resize path evidently copied instead of resizing for a third of the corpus. Regenerating true 1200px displays (~150–250 KB) via the existing `scripts/migration/backfill-display-images.mjs` machinery reclaims the bytes **and makes page loads ~4× lighter** for affected books. Detection: HEAD display vs `archived/` size, flag ratio >0.9.

### 2. Byte-identical full-res duplicates — ~0.5–1 TB
~8% of pages have `pages/{id}/{NNNN}-full.jpg` (new pipeline); where an `archived/{id}/{N}.jpg` counterpart exists it is **byte-identical (10/10 ETag matches in probe)**. One of the two copies is pure duplication. Needs a decision on which key wins (canonical `-full` vs legacy `archived/` that 77% of readers resolve through) + a repoint-then-delete pass.

### 3. `archive/` JP2 orphans — ~150–190 GB
**11 of 30 sampled JP2 dirs belong to books that exist nowhere** (not in `books`, `deleted_books`, or `failed_imports`; ids from mid/late-2025 eras — purged without R2 cleanup). Sweep: list `archive/` dirs, delete those with no book record. Separate policy question: whether JP2 source masters for *live* books (~200 GB) still earn their keep once converted to JPG in `archived/`.

### 4. `cropped/` stale generations — ~100–170 GB
All crop dirs belong to live books, but re-split books leave superseded crop files behind: filenames are page ObjectIds, and in sampled dirs ~20% of files reference pages that no longer exist (worst case: 2.6 GB dir with 0% live). GC: delete files whose page-oid is absent from `pages`.

### 5. Small fry
[...]

Issue #1814, opened 2026-05-16
Speed up R2 archiving: parallel variant uploads + raise concurrency + investigate stalled archive-ocr


JDerekLomas:
## Why

We have ~5.6M pages of archive backlog (~1.0M live + ~4.6M warehouse, ~12K books). At the current 7-day average of **25K pages/day** that's ~7.5 months to drain. Yesterday's peak of 77K/day would be ~2.5 months. We can do better.

This also surfaced an anomaly: `archive-ocr` has archived **0 pages in 7 days** (47 runs/24h, 1.6h compute, all from `archive-bulk`). Non-IA sources are silently stalled.

## Current state (as of 2026-05-17)

**Workers:**
- `scripts/workers/archive-bulk.mjs` — IA JP2-zip bulk (every 10 min)
- `scripts/workers/archive-ocr.mjs` — per-page IIIF, all non-IA sources (every 30 min)
- `scripts/workers/archive-erara.mjs` — e-rara PDFs on Mac launchd (Hetzner IPs blocked)

**Parallelism:**
- `archive-bulk`: 2 books × 8 pages = 16 in flight, 20 books/run cap
- `archive-ocr`: 5 workers per domain, per-domain rate limits (Gallica 2/s, Wellcome/Bodleian/Stanford 2–4/s, Vatican 0.1/s)
- MongoDB `maxPoolSize: 5` on both

**Per-page flow:** download → (JP2→JPEG via opj_decompress) → upload archived (full-res) → resize display (1200px) → upload display → resize thumb (150px) → upload thumb → update Mongo. Three R2 PUTs happen **sequentially** per page.

## Bottlenecks

1. **Sequential R2 writes.** `scripts/workers/lib/display-image.mjs:181,189,194` does three `await uploadFn(...)` calls in series. Each page waits for the previous variant's PUT to finish. Cheapest win — wrap in \`Promise.all\`.
2. **Page concurrency capped at 8** in `archive-bulk.mjs:40`. R2 has no practical write throttle on small objects; CPU and Mongo pool are the real limits.
3. **Mongo pool of 5** on both workers (`archive-bulk.mjs:409`, `archive-ocr.mjs:224`). Becomes the next bottleneck if we raise page concurrency.
[...]

PR #3013, opened 2026-07-05
R2 storage-reclamation measurement tooling (#3005)

JDerekLomas:
Implements the measurement + execution tooling for #3005. **All measurement is read-only** (full-bucket LIST crawls — Size+ETag come back free, no per-object HEADs). Deletion is deliberately **not** implemented — every delete pass is sign-off-gated per the `sourcelibrary` no-batch-delete rule.

## The headline correction

R2 folders are keyed by the book's **string `id`** (`books.id` == `pages.book_id`), not the Mongo `_id`; crop filenames are the page's string `id`. #3005's sampling probes used `ObjectId(_id)` joins, so they flagged live books/pages as orphans. Full crawls with the correct join (`loadLiveBookIds()`) revise the estimates sharply:

| Pass | #3005 estimate | Measured |
|---|---|---|
| 3 — `archive/` JP2 orphans | ~150–190 GB | **0 GB** (all 2,142 dirs live) |
| 4 — `cropped/` stale | ~20% | **~3.4%** (a first buggy run here showed a false 75%) |
| 2 — `-full`/`archived` dup | ~0.5–1 TB | far smaller (~58K pairs — see summary for bytes) |
| 1 — display never downsized | ~1–1.5 TB | the real prize — 2.8M+ candidates |

## What's here

- `measure-pages-archived.mjs` — Pass 1 (display bloat) + Pass 2 (byte-identical dup) via a sharded streaming merge-join of `pages/`+`archived/` (4096 bookId-prefix shards, constant memory).
- `measure-jp2-orphans.mjs` / `measure-stale-crops.mjs` — Pass 3 / Pass 4.
- `regen-display-bloat.mjs` — Pass 1 execution: force-overwrites bloated displays with true 1200px variants via the shared `generateDisplayVariants()`; source-correct (`getPageSource`, skips old-era split — the #1814 footgun); master untouched → reversible.
- `enrich-full-dup-pointers.mjs` — attaches the per-page reader-pointer decision to each Pass 2 dup pair before any delete.
- `README.md` — method + findings.

## Out of scope (deliberate)

- No deletes and no display-regen prod run in this PR — those are gated on Derek's sign-off of the manifests (in `scripts/output/r2-reclaim/`, gitignored scratch).
- The ~26M bucket objects outside `pages/`/`archived/`/`archive/`/`cropped/` (gallery/, thumbnails/, covers/, uploads/) are not measured here.
- Whether to keep JP2 masters for *live* books (368 GB) after JPG conversion is a separate policy decision, not a delete.

### DIFF
--- a/.claude/handoffs/2026-07-05-r2-reclaim-3005.md
+++ b/.claude/handoffs/2026-07-05-r2-reclaim-3005.md
@@ -1,0 +1,69 @@
+# R2 storage reclamation (#3005) — measurement complete, deletes/regen pending sign-off
+
+**Date:** 2026-07-05
+**PR:** #3013 (branch `worktree-r2-reclaim-3005`) — tooling only, no deletes, no prod writes
+**Tooling:** `scripts/maintenance/r2-reclaim/` (see its README.md)
+**Outputs (gitignored scratch, main checkout):** `scripts/output/r2-reclaim/manifest-*.ndjson`, `summary-*.json`
+
+## What was asked
+
+Implement #3005: measure (dry-run) all four reclamation passes on the 16.5 TB / 46.7M-object
+`sourcelibrary` R2 bucket, run the display regeneration, and bring the delete manifests for sign-off.
+
+## The critical finding — #3005's estimates were built on an id-space bug
+
+R2 folders are keyed by the book's **string `id`** (`books.id` == `pages.book_id`), NOT the Mongo
+`_id`; crop filenames are the page's string `id` (and page `_id`/`id` are both stored as strings and
+differ from each other). The sampling probes behind #3005 used `ObjectId(_id)` joins, which match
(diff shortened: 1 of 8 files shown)


JDerekLomas:
Merged in the 2026-08-04 sweep. Purely additive (1066+/0-) and the handoff is explicit that no delete pass has been run, so nothing here can regress existing behaviour.

One thing to fix before anyone runs a pass in anger: **`const DRY_RUN = has('dry-run')` makes dry-run opt-IN**, so `node regen-display-bloat.mjs` with no flags does the real thing. For R2 reclamation tooling that default is inverted — `scripts/maintenance/clear-fabricated-eap-dates.mjs` (#3312, merged in the same sweep) has the right shape: dry-run by default, `--apply` required to write, backup first, `--revert` available. Worth aligning before the delete passes land, since Data Protection is a CRITICAL section and 16.5 TB is an unforgiving place to discover a missing flag.

PR #3035, opened 2026-07-05
fix(artwork): stop hidden artworks leaking through semantic search

JDerekLomas:
## Problem
`match_artworks_semantic` (Supabase RPC) has **no visibility filter**, and **7,484 of 22,200 `artwork_embeddings` rows (33.7%) are for artworks now hidden in Mongo**. The backfill only embeds `visible: true` artworks, but `--incremental` never removes rows for artworks hidden *after* embedding.

Three callers consume RPC rows **directly** (title/thumbnail/id) without re-resolving against Mongo, so hidden artworks could surface:
- `src/lib/embassy/librarian.ts` — `search_artworks` tool
- `src/app/api/search/unified/route.ts` — artwork lane
- `src/app/api/identify/route.ts`

`/api/artwork/search` and `/api/gallery` already re-resolve `visible:true` and were safe — that's the fix pattern reused here.

## Fix — defense in depth
**Layer 1 (RPC/data):** add a `visible` column to `artwork_embeddings`, filter `visible IS NOT FALSE` (fail-safe for partial backfill / new NULL inserts). `add-artwork-visible-filter.mjs` adds+indexes the column, backfills from Mongo (UPDATE-only, **no re-embed, no Gemini cost**), and swaps the RPC. `add-artwork-visible-filter.sql` is a paste-ready equivalent for the dashboard. Insert path in `backfill-artwork-embeddings.mjs` now stamps `visible=true`.

**Layer 2 (app):** `filterVisibleArtworks()` drops hidden rows against **live Mongo** in the three direct-consuming callers — guards against a stale `visible` column and **closes the leak independent of Layer 1**.

## Rollout
- **Layer 2 ships with this PR** and closes the leak on deploy (filters against live Mongo, never stale).
- **Layer 1 DDL** needs `SUPABASE_DB_URL` (not in local env) or the Supabase SQL editor. Hardening + protects future callers; not required to close the leak.

## Scope
- In: the semantic-leak fix (code + migration).
- Out: `artwork_embeddings` metadata staleness (separate), retina-card / illustration-panel issues, the 1 mislabeled book (separate data edit, already applied).

`npx tsc --noEmit` clean.

### DIFF
--- a/scripts/migration/add-artwork-visible-filter.mjs
+++ b/scripts/migration/add-artwork-visible-filter.mjs
@@ -1,0 +1,148 @@
+/**
+ * Add a `visible` column to artwork_embeddings and teach match_artworks_semantic
+ * to hide non-visible artworks — the RPC-layer half of the hidden-artwork
+ * semantic-leak fix (issue: hidden artworks surfacing in semantic search).
+ *
+ * Background: backfill-artwork-embeddings.mjs only embeds `visible: true`
+ * artworks, but in --incremental mode it never removes rows for artworks that
+ * are hidden *after* embedding. At the time this migration was written 7,484 of
+ * 22,200 embedding rows (33.7%) were for artworks now hidden in Mongo, and the
+ * RPC had no visibility filter — so librarian search_artworks, /api/search/unified
+ * and /api/identify (which consume RPC rows directly) could surface hidden
+ * artworks. This is the parallel of the books visible/hidden invariant.
+ *
+ * What it does (idempotent):
+ *   1. ADD COLUMN visible boolean DEFAULT true (fail-safe default)
+ *   2. Backfill `visible` from Mongo (UPDATE-only — no re-embed, no Gemini cost)
+ *   3. Index (visible)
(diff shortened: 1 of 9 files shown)


JDerekLomas:
**Layer 1 applied + verified in production (2026-07-06).**

Ran `add-artwork-visible-filter.mjs` against the live Supabase project (`SUPABASE_DB_URL` from secret-lover):
- Added `visible` column, backfilled from Mongo: **7,484 of 22,200 rows marked hidden (33.7%)**, indexed.
- Swapped `match_artworks_semantic` to filter `visible IS NOT FALSE`.

Verification: the RPC with threshold −1 (match-all) returns **14,716** rows = exactly the visible count (7,484 hidden excluded). `artwork_embeddings_visible_idx` present.

Because the RPC is a live DB object, **the leak is now closed in production** — every caller (librarian `search_artworks`, `/api/search/unified`, `/api/identify`) hits the filtered RPC immediately. The Layer 2 app-belt in this PR (`filterVisibleArtworks`) is now defense-in-depth against a stale `visible` column and deploys with the merge. Nothing else blocks this PR.

PR #3067, opened 2026-07-07
feat(presence): live "N people are reading right now" (#3059)

### LINKED ISSUE #3059: Show "how many people are reading now" (live reader count)
**Reported via footer feedback (Derek, 2026-07-05, homepage):**
"Would be nice to see how many people are currently reading"

## Why
A live reader count makes the library feel alive and social — a small signal that others are in these texts with you. Fits the "reading room" framing.

## Status
Not built. No presence/concurrent-reader infrastructure exists today (verified — no heartbeat or active-session store; only per-user `reading-history`).

## Proposed approach (lightweight — recommended)
A rolling "N reading now" counter, not full per-user presence:
1. **Heartbeat:** the reader page (`/book/[id]`) POSTs a tiny anonymous ping (visitor_id + book_id) every ~60s while open, to a new `POST /api/presence`. Store `{ visitor_id, book_id, last_seen }` in a small Mongo collection. Heartbeats are the *deliberate exception* to the no-TTL policy (see `feedback_no_automated_data_retention`), so a TTL index on `last_seen` (~3 min) is allowed and keeps the collection self-pruning.
2. **Count:** `GET /api/presence/count` returns distinct `visitor_id`s with `last_seen > now - 3min`. Optionally per-book counts.
3. **Display:** a subtle line on the homepage hero ("N people reading right now") and/or on book pages ("N reading now"). Aggregate only — never expose who.

## Guardrails
- Aggregate counts only; no PII, no per-user exposure. Round/fuzz small numbers ("a few readers") to avoid deanonymizing at low traffic.
- Cache the homepage count (~30s) so it doesn't hammer Mongo.
- Bots excluded (reuse the bot-gate UA check) so the number reflects humans.

## Open question for Derek
Where should it show — homepage hero, book pages, or both? And is a rounded/fuzzy count ("a handful reading now") preferable to an exact number at low traffic?

---
Feedback-ID: 6a4a414efc6bb4d4743fefe8

JDerekLomas:
Closes #3059.

## What
A lightweight, aggregate **live reader count** — "N people are reading Source Library right now" — with an ambient sample of which books are open. Inspiration from [TownSquare](https://townsquare.cauenapier.com/) is the *felt presence*, not avatars/chat; the visual is a quiet live line with a soft pulse dot, tuned for a reading room.

Design was agreed with Derek on the issue: **site-wide total** (not per-book), **quiet live line** visual, **homepage hero + persistent global chip**, titles but never a specific person↔book link.

## How it works
- **`POST /api/presence`** — anonymous heartbeat (~50s) keyed on a random localStorage UUID. Bot-filtered via `classifyTraffic` (same Cloudflare signals as `/api/views`). One self-expiring doc per browser. **No IP, no login, no PII** stored or returned.
- **`GET /api/presence/count`** — distinct visitors active in the last ~2 min + a sample of **visible-book** titles resolved **server-side** (the client only ever sends a slug/id — a crafted heartbeat can't inject a fake or hidden-book title onto the homepage). Cached ~15s.
- **`presence` collection** — TTL index on `expireAt`, the sanctioned exception to the no-TTL policy (heartbeats self-prune).
- **`ReaderPresence`** — hero variant under the homepage tagline (with titles); compact chip in the global footer (persistent, every global page) which carries the single heartbeat.

## Guardrails
- **Tenant lockdown:** no UI and no heartbeat on tenant subdomains / embeds (BPH/EFM stay closed systems). Gated via `useEmbedContext` **and** a synchronous `isEmbeddedNow()` check so not even one beat leaks in the first-render window before the hook settles.
- **Aggregate only, never who.** Random visitor id, never displayed.
- **Low-traffic honesty:** hides below 2 concurrent readers (no lonely "1 reading"); titles hidden below 3.
- Bots excluded so the number reflects humans.

## Out of scope (deliberate)
- Not on reader pages beyond the footer chip (Derek chose homepage + persistent site-wide, not a reader-page line).
- Per-title reader-count floor (currently a single reader's book *can* appear in the sample once total ≥3) is left as a tunable knob, not implemented — flagging for review.

## Verification
- `npx tsc --noEmit` clean.
- Endpoints + tenant-exclusion verified against the Vercel preview (see comment).

### DIFF
--- a/src/app/api/presence/count/route.ts
+++ b/src/app/api/presence/count/route.ts
@@ -1,0 +1,34 @@
+/**
+ * Live reader count — GET /api/presence/count
+ *
+ * Returns the aggregate presence snapshot for the "N people are reading right
+ * now" line (issue #3059):
+ *   { count: number, titles: string[] }
+ *
+ * `count` is distinct anonymous visitors active in the last ~2 min. `titles`
+ * is a small sample of visible-book titles open right now (empty below a
+ * small floor so a lone reader isn't singled out). Never exposes who.
+ *
+ * Cached briefly so it doesn't hammer Mongo — the number only needs to feel
+ * live, not tick every request.
+ */
+
+import { NextResponse } from 'next/server';
+import { getDb } from '@/lib/mongodb';
+import { getPresenceSnapshot } from '@/lib/presence';
+
+export async function GET() {
+  try {
+    const db = await getDb();
+    const snapshot = await getPresenceSnapshot(db);
+    return NextResponse.json(snapshot, {
+      headers: { 'Cache-Control': 'public, max-age=15, stale-while-revalidate=45' },
+    });
+  } catch {
+    // A count failure should degrade to "hidden", never error the page.
+    return NextResponse.json(
+      { count: 0, titles: [] },
+      { headers: { 'Cache-Control': 'public, max-age=15' } }
+    );
+  }
+}
--- a/src/app/api/presence/route.ts
+++ b/src/app/api/presence/route.ts
@@ -1,0 +1,75 @@
+/**
+ * Presence heartbeat — POST /api/presence
+ *
+ * A tiny anonymous ping the reader page (and the global footer chip) sends
+ * every ~50s while open. Powers the "N people are reading right now" line
+ * (issue #3059). One self-expiring doc per browser, keyed on an anonymous
+ * localStorage UUID. No IP, no login, no PII is stored — the visitor_id is
+ * random and is never returned to anyone.
+ *
+ * See src/lib/presence.ts for the aggregate read path and design invariants.
+ */
+
+import { NextRequest, NextResponse } from 'next/server';
+import { getDb } from '@/lib/mongodb';
+import { classifyTraffic } from '@/lib/traffic-classification';
+import {
+  PRESENCE_COLLECTION,
+  TTL_MS,
+  VISITOR_ID_RE,
+  BOOK_KEY_RE,
+  ensurePresenceIndex,
+} from '@/lib/presence';
+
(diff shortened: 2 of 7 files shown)


JDerekLomas:
## Verified on the Vercel preview

Exercised the full write→read path against the preview deployment:

- **Heartbeat records:** 3 anonymous POSTs (real UUIDs, 3 real book slugs) → `GET /api/presence/count` returned `{"count":3, titles:[3 real titles]}`. Titles are resolved **server-side** from the visible `books` collection — the client only sent slugs.
- **Bot excluded:** a 4th heartbeat with a `GPTBot` UA did **not** increment the count (stayed exactly 3, not 4). `classifyTraffic` gate works.
- **Titles floor:** titles appear at count ≥ 3, hidden below.
- **Caching:** first read served a stale `count:0` from the edge (`max-age=15`); a cache-busted read showed `x-vercel-cache: MISS` and the true count — caching behaves as designed (the number goes live within ~15s).
- **Self-pruning:** the synthetic test visitors age out on their own via the TTL index (~3 min) — no cleanup needed, which is the point.
- `npx tsc --noEmit` clean.

**Tenant exclusion** is code-gated (browser screenshot blocked — extension not connected this session): the API is intentionally global; the UI + heartbeat are gated by `useEmbedContext` **and** a synchronous `isEmbeddedNow()` check at beat time, both using the same `isTenantSubdomain` regex, so no chip renders and not even one beat fires from a tenant subdomain — including the first-render window before the hook settles. Worth a quick manual look on `bph.` after merge to confirm visually.

**For review — one knob:** a single reader's book title *can* appear in the sample once total ≥ 3. If you'd rather never surface a book with only one active reader, I can add a per-title floor (≥2 readers) — left out for now because it makes titles rarely show at low traffic.

Issue #3088, opened 2026-07-08
BPH gutter re-crop: gutter-aware audit + targeted cohort sweep (#2898 follow-up)


Mayank-PPL:
Follow-up to #2898 (fixed the reported book + shipped tooling in PR #3082). This tracks the **cohort remediation** that was deliberately deferred.

## Background
Old-era BPH crop-coord splits cut spreads at a single shared gutter line with **zero overlap** (`left.xEnd == right.xStart`), clipping content that reaches the gutter. #2898 fixed the one reported book (**Artis auriferae**, `69751588a88d83c830d99e16`) in place via `scripts/maintenance/recrop-bph-gutter.mjs` (re-cut each half from the immutable full spread with ~3.5% overlap, new R2 paths, keep page IDs). Verified end-to-end.

**Cohort:** ~**1,208 BPH books / ~410k crop pages**, all zero-overlap (`held_by:'bph'` ∩ pages with `crop.xStart` + materialized `cropped_photo` + not `split_from_spread`).

## Why the sweep was deferred (the key finding)
The first-pass audit (`scripts/maintenance/audit-bph-gutter-clip.mjs`) measures per-column ink (`min(R,G,B)<120`) just outside each cut. On the real cohort it **over-flags ~97%** because the raw ink signal can't distinguish clipped text from:
1. **Binding shadow** on tight-bound pamphlets (saturates ~1.0 ink vs ~0.6 for text) — the highest-"severity" flags are actually shadow;
2. **Facing-page bleed** when the two text blocks sit close to the gutter.

Verified on the top-flagged pamphlet: its "clipping" is the dark binding shadow, and a naive +35‰ re-crop there would pull in shadow + the facing page rather than recover content. So the verified +35‰ fix is safe for **wide-gutter** books (Artis auriferae) but **not** for the tight-gutter majority.

### 1. Make the audit gutter-aware (the real work)
Refine `audit-bph-gutter-clip.mjs` so it counts **only left-page content between the cut and the gutter**, excluding saturation and anything past the gutter:
- Locate the gutter per page (deepest/widest ink-free run for bright gutters, or the saturated dark run for tight bindings) — reuse `scripts/lib/gutter-detect.mjs` (`detectGutterPixel`) / the `detectGutterColumn` logic in `batch-split-bph.mjs`.
- Handle the ToC/right-aligned-numeral case (content sits *between* an inter-column valley and the true gutter — see Artis p386).
[...]

Issue #2898, opened 2026-06-30
BPH spread split: zero-overlap crop cuts off gutter-edge text (Artis auriferae + ~1,230-book cohort)


JDerekLomas:
## Summary

A BPH book was reported as "not splitting properly." Investigation showed the spread **is** split into single pages, but the split crop cuts at a **hard gutter line with zero overlap**, shaving off content that sits near the gutter (right-margin numeral columns, catchwords, long line-ends). The same zero-overlap split was applied to a large cohort of BPH books, though visible damage is likely limited to a minority of pages.

## Reproduction

- Book: **Artis auriferae, quam chemiam vocant** (Morienus, Latin, 1593) — `book_id 69751588a88d83c830d99e16`
- Page: printed **386** (Rosarium "Capitula libri primi" table of contents) — `page_id 6976a76a2b4266f8a88c9b02`
- Reader: https://bph.sourcelibrary.org/book/69751588a88d83c830d99e16/page/6976a76a2b4266f8a88c9b02
- The right-margin chapter-number column (`I, II, III, IIII, V, VI`) and the `Ex` catchword are cut off the gutter edge. Comparing against the full spread (`archived_photo`) confirms ~65px of real content is lost on the right.

## Root cause

The page carries `crop = {xStart: 0, xEnd: 505}` (per-mille, 0–1000) and its sibling (`6976a76a2b4266f8a88c9b03`, printed 387) carries `{xStart: 505, xEnd: 1000}`. The split boundary is a **single shared gutter line** — `left.xEnd == right.xStart` — so there is **no overlap**. Page 386's text block extends to ~52% (≈520‰) while the cut is at 505‰, so the numeral column falls past the cut and is lost.

This is the older crop-coord BPH split mechanism (sets `crop` + `split_detection`, **not** `split_from_spread`). The current `scripts/workers/batch-split-bph.mjs` does add overlap (`OVERLAP_PX = 10`), so newly-split books are not affected.

Fix verified locally: re-cropping the left half at `xEnd ≈ 540` (~3.5% overlap) recovers the full numeral column with margin and **without** bleeding in any of page 387.

## Scope (BPH cohort)

Sampling all 3,549 BPH books with pages (`held_by: 'bph'`):

- **~1,230** use the old crop-coord split.
[...]

PR #3090, opened 2026-07-08
fix(bph): gutter-aware clip audit + per-page gutter-relative re-crop (#3088)

Mayank-PPL:
Closes the audit + tooling half of #3088 (follow-up to #2898 / PR #3082). The
cohort **sweep + re-OCR run itself is deferred to Hetzner** (documented), since
it's an image-heavy multi-hour job on the pipeline box.

## Problem
Old-era BPH splits cut spreads at a shared gutter line with **zero overlap**,
clipping left-page content that reaches the binding. The v1 audit measured raw
ink just past the cut and **over-flagged ~97%** of the ~1,208-book / ~410k-page
cohort — it couldn't distinguish clipped text from **binding shadow** (saturates
~1.0; text tops ~0.6 because line leading leaves white rows) or **facing-page
bleed** past the gutter.

## What's here
- **`scripts/lib/gutter-clip.mjs`** — shared gutter-aware primitive (used by both
  audit and recrop so they never drift):
  - locate the **true gutter** in a window around the cut — the *widest* bright
    ink-free run (picking widest beats the ToC inter-column-valley trap, Artis
    p386) or the saturated dark shadow on tight bindings;
  - count recoverable content **only between cut and gutter, on the correct side**
    (sign of `gutter − cut` rejects a cut already at/past the gutter = over-cut/bleed);
  - count only text-like columns and **drop any within `SAT_ADJ‰` of a saturation
    peak** — this kills the binding-shadow *ramp* (the p197 false positive).
- **`audit-bph-gutter-clip.mjs`** rewritten on that primitive. On a 21-book
  sample: **~10% flagged** (vs 97%), every flag text-driven (`shadowFrac 0`).
  Adds `--profile <id>` (per-page geometry dump) and an **HTML eye-check gallery**
  that overlays cut (red) / gutter (green) / recover band on the real spread.
- **`recrop-bph-gutter.mjs`** gains **`--to-gutter`**: per-page, move the cut to
  the *detected* gutter centre — recovers exactly the clipped content with no blind
  overlap guess, and is **clip-gated by the same `measureClip`** so recrop ⇔ audit
  agree (non-clipping pages left untouched). Blind `--overlap` mode kept for
  wide-gutter books. Still non-destructive / idempotent / reversible.
- **`.claude/docs/bph-gutter-clip-audit-and-sweep.md`** — Hetzner run plan +
  re-OCR decision guidance.

## Validation (spot-checked against composited overlays)
- `40 questions… p50` — real left-page clip (text bump ink 0.24 before a wide
  blank gutter) → **flagged** ✅
- `40 questions… p197` — binding shadow (ink peaks 0.88, right-page text starts at
  the cut) → **not flagged** after the shadow-ramp fix ✅
- `Eine kurtze…` / `L'Elixir…` — trailing line-end punctuation / marginalia past
  the cut → **flagged** (modest, low severity) ✅
- `--to-gutter --dry-run` on the top flagged book: **36/244 pages** touched, cut
  moved to each detected gutter; rest left as-is.

## Out of scope (deliberately deferred, per issue)
- Running the full audit + sweep on Hetzner (multi-hour, image-heavy).
- Re-OCR of swept pages — separate, paid (Gemini), pipeline paused; low-yield on
  the sampled cohort (mostly trailing glyphs). Decision guidance in the doc.

### DIFF
--- a/.claude/docs/bph-gutter-clip-audit-and-sweep.md
+++ b/.claude/docs/bph-gutter-clip-audit-and-sweep.md
@@ -1,0 +1,113 @@
+# BPH gutter-clip: gutter-aware audit + targeted re-crop sweep (#3088)
+
+Follow-up to #2898 (which fixed one book — *Artis auriferae*, `69751588a88d83c830d99e16` — and
+shipped the first tooling in PR #3082). This is the **cohort remediation** that was
+deferred: find the BPH books that *actually* clip text at the gutter, and re-crop only
+those, per-page, to the detected gutter.
+
+## The defect
+Old-era BPH splits stored each half as a per-mille crop against the full spread with
+**zero overlap** — `left = {0, c}`, `right = {c, 1000}`, the two facing pages sharing
+the cut line `c`. Where the old splitter put `c` short of the true binding gutter
+(classically it snapped to an **inter-column valley** — the gap between the body text
+and a right-aligned numeral/catchword column — instead of the binding), left-page
+content between the cut and the gutter is shaved off.
+
+These pages carry a **materialized `cropped_photo`**, so the reader serves the pre-cut
+file (`getPageSource()` returns `cropped_photo`, `cropRegion()` returns undefined —
+`src/lib/page-image-url.ts`). Editing `crop` coords alone changes nothing; the half
+image must be regenerated from the immutable full spread and written to a new R2 key.
+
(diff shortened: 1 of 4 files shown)


JDerekLomas:
DCO check is blocking the merge — the commit is missing your Signed-off-by line. One-liner to fix: `git commit --amend --no-edit -s && git push --force-with-lease`. Everything else is green; happy to merge as soon as that's in.

PR #3108, opened 2026-07-09
fix(reader): strip untagged conversational AI preambles from OCR text

JDerekLomas:
## Problem

Reported by Derek from https://sourcelibrary.org/book/69528aa5ab34727b1f04ed4e/page/69528aa5ab34727b1f04ed60 — the page renders:

> Note: The text in the image is in French, not Latin. I have transcribed it exactly as it appears, including the use of the long 's' (ſ) and original punctuation.

as if it were source text. Dec-2025-era OCR batches asserted the book's `language` field in the prompt; on mistagged books (this one imported from Gallica as Latin, actually French) Gemini prepended an untagged disclaimer before the transcription, and the whole reply was stored in `ocr.data`. Because the chatter is bare prose — not inside any editorial wrapper tag — `stripEditorialWrappers` can't catch it, and it surfaces **indistinguishable from the page text** in the reader and every quote/snippet surface. Same class: "Here is the transcription:" prefaces and outright refusals ("I'm sorry, I cannot…").

Affected: 80/462 pages in this book alone; a library-wide scan is running (results will be posted on the tracking work).

## Fix

Add `stripLeadingAiPreamble` as the **first step of `cleanOcrArtifacts`** — the same single choke point used for the #2764 lacuna guard, so the reader (`NotesRenderer`) and all `stripEditorialWrappers` consumers (quote API, search, IIIF, likes, embeddings cleaning) inherit it with no per-surface wiring.

Conservative on purpose, to protect real page text:
- anchors to the very start of the text; leading paragraph(s) only (max 3), each ≤500 chars
- fires only when a conversational opener **AND** self-referential transcription vocabulary ("I have transcribed", "the image", "as requested", …) both match
- printed editorial notes that merely begin with "Note:" and body text beginning "The image …" survive (tested)
- tagged content is left to the wrapper strip

## Out of scope (follow-ups)

- **Prompt hardening**: OCR prompt v15 with an explicit output contract (any commentary must live inside `<warning>`; never address the requester) — DB change, being applied separately.
- **Data sweep / re-OCR**: stored `ocr.data` still contains the preambles (harmless once this read-time guard ships); refusal-only pages may warrant re-OCR pending the library-wide scan.

## Tests

- 9 new unit tests incl. the verbatim production case; all 35 in the file pass.
- `npx tsc --noEmit` clean.


JDerekLomas:
Library-wide scan complete (48,068 books with OCR, per-book indexed scan):

**35 books / 1,859 pages** have untagged conversational preambles.
- **1,839 disclaimer pages** — real transcription follows (language-mismatch notes like this PR's production case, 'Here is the transcription…' prefaces). Fully neutralized at read time by this guard.
- **20 refusal pages** — no transcription at all; the source image was a solid-black scan or an IA 'image not available' placeholder. The guard strips these to empty (honest render); actual fix needs image repair + re-OCR, tracked separately.

Top affected books (all Dec-2025-era, prompt asserted a wrong language): Mysterium Magnum 310pp, Vie de sainte Thérèse 261pp, Proclus Theology of Plato 241pp, Recueil de théologie mystique 229pp, La chymie charitable 196pp… concentrated in 10 books ≈97% of pages.

Openers observed in production all match the guard: 'Note: Although the prompt asks…', 'Note: The text in the image is…', 'Here is the transcription…', 'I cannot fulfill this request…', "I'm sorry, but…", plus the 'If you provide…'/'Please provide…' refusal tails added in 51195230.

PR #3111, opened 2026-07-09
fix(reader): float the magnifier toggle, hide the lens over it

JDerekLomas:
Follow-up to #3095 (magnifier lens on/off toggle).

Two problems with the toggle as shipped:

1. **It scrolled away.** The button was `absolute top-2 left-2` on the image, so once you scrolled down a tall facsimile it was gone — exactly when you want it, since the lens captures the wheel for zoom. It now floats, tracking the top of the visible slice of the image.
2. **The lens covered it.** The lens is centred on the cursor, so as you reached for the button the lens sat on top of it. Hovering the toggle (within 8px) now hides the lens, and the button paints above the lens (`z-index: 110` vs the lens's `100`) so it stays visible when the lens is merely nearby.

## Why not `position: sticky`

Every caller wraps the image in a `rounded-lg overflow-hidden` box. That box becomes sticky's scroll container, and it never scrolls — so a sticky button would simply sit still. Instead the effect walks up to the nearest ancestor whose `overflow-y` is `auto`/`scroll` (the reader panel, `[data-reader-panel]`) and offsets the button from the container's own top edge, clamped so it never runs past the image's bottom. Updates are rAF-throttled and bound in the capture phase, since `scroll` doesn't bubble.

Page-scrolled callers (`ArtworkHero` with `fitWidth`) have no scrollable ancestor; there the ceiling is the bottom of the sticky site header (`[data-site-header]`), or 0 when the header is static.

## Verified

Driven in Chrome against the reader (`/book/four-books-on-human-proportion-durer/page/…`, a 315-page Dürer facsimile):
- scrolled the image panel to its bottom → button pinned 8px below the panel's top edge (`style.top` 8px → 340px), still on screen
- scrolled back to top → button returns to the image corner
- hovered mid-image → lens appears, and renders *under* the button
- hovered the button → lens disappears

`npx tsc --noEmit` clean.

## Out of scope

The deep-zoom button (top-right, `PageDeepZoomButton`) still scrolls with the image. Same treatment would apply, but it's a separate component and this PR stays on the magnifier.

### DIFF
--- a/src/components/reader/PageDeepZoomButton.tsx
+++ b/src/components/reader/PageDeepZoomButton.tsx
@@ -16,9 +16,10 @@
  * see memory: inline state there silently failed (reactCompiler + IIFE layout).
  */
 
-import { useState, useEffect, lazy, Suspense } from 'react';
+import { useState, useEffect, useRef, lazy, Suspense } from 'react';
 import { Search, X, Maximize2 } from 'lucide-react';
 import type { DeepZoomManifest } from '@/lib/types/book';
+import { useFloatingOverlayTop, LENS_AVOID_ATTR } from '@/hooks/useFloatingOverlayTop';
 
 const DeepZoomInline = lazy(() => import('./DeepZoomInline'));
 const DeepZoomOverlay = lazy(() => import('@/components/artwork/DeepZoomOverlay'));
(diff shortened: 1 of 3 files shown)


JDerekLomas:
Folded in the deep-zoom button, as discussed.

- Float logic extracted to `useFloatingOverlayTop` and shared by both controls.
- The deep-zoom button now floats too (it had the same problem — pinned top-right, gone once you scrolled the scan).
- The lens also clears out of the deep-zoom button's way. Rather than teach `ImageWithMagnifier` about a sibling component, controls opt in with a `data-lens-avoid` attribute and the lens hit-tests whatever it finds under the shared positioning wrapper.

Verified in Chrome on a page with a deep-zoom manifest (`/book/on-famous-women-de-claris-mulieribus-boccaccio-4/page/…`): after scrolling the panel, both controls sit 8px below the panel's top edge (`style.top` 245px / 246px — they differ by the wrapper's 1px border, since their `offsetParent`s differ); the lens renders under both; hovering either one hides the lens.

`npx tsc --noEmit` clean.

PR #3127, opened 2026-07-10
feat(kdp): rich reflowable EPUB, portrait cover, and bundle test CLI

fractastical:
## Summary
Enhances the existing assisted KDP publishing workflow (Amazon KDP has no publish API, so this stays an assisted, human-uploads workflow — this PR improves the *artifacts* it produces).

- **Rich reflowable EPUB** (`src/lib/kdp-epub.ts`, rewritten with `archiver`): embedded cover, title page (with optional original title-page scan), copyright/license page carrying the AI disclosure + methodology, chapter-grouped body with folio markers (`[p. N]`) and inline facsimile/illustration images, glossary (from `vocabulary`), index (`keywords`/`people`/`places`/`concepts`), colophon, and nested nav. All page text routes through `stripEditorialWrappers` + `markForExport`, so no editorial annotation or AI preamble leaks into published prose (verified by the new smoke test).
- **Portrait cover** (`src/lib/kdp-cover.ts`): fixed to KDP's eBook spec **1600×2560** (was landscape 2560×1600); layout coordinates are derived from the canvas so text stays centered, with a font step-down for long titles.
- **Callers updated**: bundle route + both `mint-doi` routes now pass `db` and the new options (`license`/`introduction`/`methodology`); bundle route `maxDuration` bumped to 300s for inline image fetching.
- **New test CLI** `scripts/kdp/test-bundle.ts`: generates a full bundle (EPUB + cover + metadata) to disk for one book without the dashboard/login — `npx tsx scripts/kdp/test-bundle.ts <bookId|slug>` — so you can run the EPUB through Kindle Previewer before uploading.

## Test run
Verified against the live book *Easy and Charitable Chemistry for Ladies* (Marie Meurdrac, 1666, 371 pp): portrait cover 1600×2560, reflowable EPUB with 54 chapters, nested nav, 100-term index, folio markers, inline glosses preserved, no wrapper leakage. Glossary/illustrations correctly absent (this title has neither). `tsc --noEmit` clean; `kdp-epub-smoke.test.ts` passes.

## Out of scope
- Automated upload to KDP (no public API; deliberately not built — see workflow discussion).
- Tightening the stale `BookIndexData` interface in `src/lib/book-index.ts` (misleading typing only; runtime shape is correct). Can follow up separately.

## Test plan
- [ ] `npx tsx scripts/kdp/test-bundle.ts easy-and-charitable-chemistry-for-ladies-meurdrac` (needs `.env.production.local`)
- [ ] Open the generated `.epub` in Kindle Previewer; confirm chapters, nav, cover, front/back matter
- [ ] Generate a bundle from `/admin/kdp` and confirm parity

Made with [Cursor](https://cursor.com)


JDerekLomas:
**Left unmerged in the 2026-08-04 sweep — the one PR here that needs a real review, not a rubber stamp.**

Everything else in the stale tail was either docs or purely additive (no deletions, so it cannot regress existing behaviour). This one is **888+/400−** across `src/app/admin/kdp`, `src/app/api/admin/kdp/bundle/[id]`, and `src/app/api/[tenant]/books/[id]/editions/mint-doi` — real modifications to live routes, one of which **mints Zenodo DOIs**.

DOI minting is irreversible in the way that matters: a minted DOI is a public, permanent identifier, and a bug that mints against the wrong edition cannot be taken back, only superseded. It also went 25 days with main moving underneath it.

Not blocking it — no defect found, because I did not review it properly. Flagging that it needs someone to read the mint-doi diff against current main before merge, and that merging it requires `npm run deploy:prod` afterwards.

JDerekLomas:
## Unblocked and reviewed — the flagged risk isn't in this diff

Picking up @JDerekLomas's [note above](https://github.com/Embassy-of-the-Free-Mind/sourcelibrary-v2/pull/3127#issuecomment-0), which asked for the `mint-doi` diff to be read against current main before merge, on the grounds that a minted DOI is irreversible.

**Read it. The concern doesn't attach to this change.** Both `mint-doi` routes get exactly two lines: `db` added to the `generateKdpEpub(...)` call, and `edition.license` passed through. Nothing touches Zenodo, the DOI, or which edition is minted. The `db` argument exists solely so the generator can reach `gallery_images` for inline plates — that's the whole reason for the signature change.

### Was conflicting, and that's why it looked green

The checks read green while sitting **24 days stale** (last run 2026-07-28). The branch had gone `CONFLICTING` against main, so GitHub could not build a merge commit and simply queued nothing — no failure, no annotation. Merged `origin/main` in and resolved it.

The conflict was one hunk, and it was a false alignment: this branch rewrote `kdp-epub.ts`, while #4053 changed four lines of the **old** colophon on main, so git lined up two unrelated regions. Kept this branch's Introduction block and carried #4053's change to the colophon this branch actually builds — place now resolves through `resolveImprintPlace()` instead of reading the bare `place_published` column.

### Three defects found and fixed

1. **Duplicate images.** A figure is the full page scan (deliberate, per 249c47a0 — it frames better than a tight crop), so every detection on a page resolves to the *same* url. Keying the asset by detection fetched that scan once per plate, embedded N identical JPEGs, and rendered the plate N times in a row. Now keyed by url, captioned with all of that page's detections.
2. **No time budget.** Up to 60 serial fetches at a 60s timeout each, inside a route with `maxDuration = 300` — one slow scan host could burn the whole budget and return nothing. Added a 180s ceiling for the loop, dropped the per-fetch timeout to 25s, and it logs what it omitted rather than skipping silently.
3. **Cover picked by truncation, not by bias.** `pickKdpCoverImageUrl` sorted by `gallery_quality`, cut to 20, *then* applied the type bias — but eligible quality spans 0.25 (0.75–1.0) while the bias spans 0.35 (+0.15 frontispiece to −0.2 diagram), so a frontispiece below the cut can outscore a diagram above it. Pool widened to hold every eligible plate.

The smoke test now pins the dedup — it counts upstream fetches and asserts one embedded asset and one `<figure>` for a two-detection page. Confirmed it **fails** when the url-keying is removed, so it's a guard rather than decoration.

`npx tsc --noEmit` clean; full suite green (2,994 unit + 41 integration).

### Still owed before merge

- #3128 — the golden-sample pass through Kindle Previewer. Opened by the PR author, still unchecked, and it is the only step that puts human eyes on the artifact this PR exists to produce.
- Merging this needs `npm run deploy:prod` after, per the note above.

PR #3171, opened 2026-07-16
Reader: edition bibliographic info + Source Library link in the Info panel

JDerekLomas:
Closes #3170.

## What

The reader's existing **Info** button now also shows the book's full bibliographic record, so a reader who lands on any page view can see (and cite) exactly what edition they're reading without leaving the page.

- **`PageMetadataPanel`**: new optional edition props; renders a "This Edition" section above the page metadata containing `BibliographicInfo` (in its native dark container) + a "View on Source Library" link to the canonical book page.
- **`TranslationEditor`**: lazy-fetches the full book doc from `GET /api/books/[id]?pageLimit=1` the first time the panel opens — once per book per session, cached in state. Graceful degradation: fetch failure shows a quiet "couldn't load" line; page metadata always renders.
- **`BibliographicInfo`**: new `defaultExpanded` prop (panel opens it pre-expanded; book detail page behavior unchanged).

## Why lazy-fetch instead of widening the reader projection

Reader pages are the hottest, most heavily CDN-cached routes. Widening `BOOK_NAV_PROJECTION` would serialize the same bibliographic blob into every cached render of every page (500-page book = 500 copies across the edge cache) for a panel most readers never open. The on-demand fetch (~145 KB, verified against prod — the doc carries index/chapters) is paid only by readers who ask for it. Details + the rejected alternative are in #3170.

## Tenant lockdown

Embed/tenant readers share these components. `showExternalLinks={!isEmbedded}` suppresses VIAF/Wikidata/USTC/IIIF/source_url outbound links in embed mode; the book link is relative (`${tenantPrefix}/book/<id>`) so it resolves on-host, and the link label drops the "Source Library" name when embedded.

## Out of scope (per issue)

Light-theme restyle of `BibliographicInfo`; `field_provenance` in the reader.

## Verification

- `npx tsc --noEmit` clean.
- `GET /api/books/[id]?pageLimit=1` verified against prod: full book doc (publisher/place/image_source present) + exactly 1 page row.
- Visual check on the Vercel preview (screenshots in comments).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3173, opened 2026-07-17
docs: deploy:prod can fail after promotion but before purge — recovery steps

JDerekLomas:
One-paragraph addition to the CLAUDE.md deploy invariant, from a live incident tonight (details in the commit message; hit during the #3171 deploy). The failure mode: `write EPIPE` from the Vercel CLI mid-poll reads as a failed deploy, but the build had already promoted to production server-side — and the script died before its purge+warm steps, opening the stale-HTML/dead-CSS window. The fix documents: on any deploy:prod failure, check `npx vercel inspect sourcelibrary.org` first; if promoted, run purge + deploy-warm manually.


JDerekLomas:
**Superseded by #3412 (merged) — closing.**

This went `CONFLICTING` while open and the content never reached `main`, so the deploy:prod EPIPE recovery lived only in one machine's private memory where no other contributor could reach it. #3412 redoes it on current main, re-verified today.

Closing rather than rebasing, per what this PR itself demonstrates: a conflicting docs PR usually conflicts *because the doc moved under it*, which is the same reason its assertions may have gone stale.

PR #3174, opened 2026-07-17
Kircher station legibility: sticky results, visible strike marker, explained response bars

JDerekLomas:
Follow-up to #3172, from Derek's feedback (screenshot showed Strike A pressed with every readout at 0%).

Three fixes to Station VIII:
1. **Results no longer erase themselves.** The 2.6s reset that zeroed the readouts is gone — the response stays until the next strike; only the string-vibration visual decays with the sound.
2. **The struck tone is now visible.** Strings sit at their true positions on a log-frequency axis, and each strike draws a dashed 'struck · N Hz' marker — so 'Strike between A and B' visibly lands between the strings and matches none.
3. **The numbers are explained.** Response bars under each string replace the five percent cards; two sticky readouts summarize (Struck / Strongest answer) and the caption states the mechanism: one tone fed equally into all five untouched resonators, each absorbing energy only at its own natural frequency.

Verified in-browser: Strike A → A bar 100% (neighbours 8/8/3/2%), readouts persist past the old reset window; between-strike → 'matches no string', strongest answer A · 16%, 'no string answers'. tsc clean. Derived values precomputed as scalars per the React Compiler hoist hazard.

### DIFF
--- a/src/components/blog/lab/KircherResonanceDemo.tsx
+++ b/src/components/blog/lab/KircherResonanceDemo.tsx
@@ -2,7 +2,7 @@
 
 import { useEffect, useRef, useState } from 'react';
 import { ToneEngine } from './audio';
-import { LabCard, Chip } from './LabCard';
+import { LabCard, Chip, Readout } from './LabCard';
 
 /** The five tuned strings of the bank. Spaced whole steps apart or more, so resonance is selective. */
 const STRINGS = [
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Superseded the five-string bank in this same PR: the station is now literally Kircher's two-string demonstration — one plucked string fixed at A · 220 Hz, one untouched string the reader tunes (slider + G/A/B presets, cents-from-unison readout). Starts mistuned so the first pluck fails; finding unison is the experiment. Verified in-browser: mistuned pluck → 8% 'stays silent'; retune → 'pluck again' hint; unison pluck → 100% 'sings back', sticky.

PR #3277, opened 2026-07-20
fix(seo): return real 404 for unknown /book and /collections slugs (#3232)

### LINKED ISSUE #3232: Unknown /book and /collections slugs return HTTP 200 soft-404s ('Not Found' body, no 404 status)
Observed during the #2266 investigation (2026-07-19): requesting a non-existent `/book/<slug>` or `/collections/<slug>` returns **HTTP 200** with a 'Not Found' title/body instead of a real 404 status.

Consequences:
- Crawlers index junk URLs / waste crawl budget (Google treats them as duplicate soft-404s at best).
- Link checkers and our own audits can't detect breakage by status code — CLAUDE.md already documents the grep-the-body workaround for reader pages; this extends it to book/collection routes.
- Monitoring (e.g. not_found_reports) may undercount.

Note: `src/app/book/[id]/page.tsx` deliberately does its existence check *before* streaming so it can return a real 404 (see the comment at the top of `BookDetailPage`) — verify why a 200 still escapes (CDN-cached fallback? notFound() inside a flushed stream on some path? collections route lacking the early check?) and fix per-route.

Acceptance: `curl -s -o /dev/null -w '%{http_code}' https://sourcelibrary.org/book/definitely-not-a-real-slug-xyz` returns 404; same for `/collections/`.

Refs #2266

Mayank-PPL:
Fixes #3232.

## Problem
Requesting a non-existent \`/book/<slug>\` or \`/collections/<slug>\` returned **HTTP 200** with a "Not Found" body (a soft-404). Crawlers index junk URLs, status-code link checkers can't detect breakage, and \`not_found_reports\` undercount.

Confirmed against production:
\`\`\`
/book/definitely-not-a-real-slug-xyz        → 200
/collections/definitely-not-a-real-...-xyz  → 200  (x-vercel-cache: MISS — freshly rendered, so NOT a caching issue)
\`\`\`

## Root cause
Each route had a \`loading.tsx\`. A \`loading.tsx\` wraps its \`page.tsx\` **and any children below** in an automatic \`<Suspense>\` boundary; Next.js flushes the 200 loading shell as soon as the page's data fetch suspends — **before** the page component's \`notFound()\` runs. Once the 200 shell is flushed, \`notFound()\` can only swap in the not-found UI; it can't change the already-sent status. This is exactly the mechanism the proxy already documents for the reader route ("that route is ISR, so it returns HTTP 200 with a Page Not Found body").

The book page's existing "existence check *before* streaming" was silently defeated by this — the \`loading.tsx\` boundary sits *above* that check.

## Fix — per route, keeping the skeleton UX
- **\`book/[id]\`**: delete \`loading.tsx\`. The detail page already does a shell-level existence check (\`getBookForMetadata\` → \`notFound()\`) and streams its heavy content through an inner \`<Suspense fallback={BookInfoSkeleton}>\`, so the skeleton is preserved and the shell \`await\` now commits a real 404. The reader (\`page/[pageId]\`) and \`pipeline\` keep their own \`loading.tsx\` (unaffected).
- **\`collections/[id]\`**: move the existence check into the shell via a \`cache()\`'d \`getCollectionDoc\` (reused by \`fetchCollectionData\`, so no extra DB round trip; throws on DB error → \`error.tsx\`/500, returns null only when genuinely absent → 404). The heavy body streams through an inner \`<Suspense>\` with the former loading skeleton. Both \`collections/loading.tsx\` and \`collections/[id]/loading.tsx\` are deleted — the \`/collections\` list page self-streams via its own inner \`<Suspense>\` boundaries, so it keeps its skeleton.

## Out of scope
- The **reader** route (\`/book/<slug>/page/<pageId>\`) soft-404 is only partially handled here (numeric page numbers are already caught in \`proxy.ts\`); a real-but-missing page id still soft-404s. #3232 asks specifically for \`/book\` and \`/collections\` — the reader can be a follow-up.

## Verification
Curl matrix against the preview deploy (to be pasted once the preview builds):
- \`/book/<bad-slug>\` → 404, \`/collections/<bad-slug>\` → 404
- a real book and a real collection still → 200 and render fully


Mayank-PPL:
### ✅ Verified on the preview deploy

Curl matrix against the auto-deployed preview:

| URL | Status | Notes |
|---|---|---|
| `/book/<bad-slug>` | **404** | not-found body present |
| `/collections/<bad-slug>` | **404** | not-found body present |
| `/book/four-books-on-human-proportion-durer` | 200 | full render (86 anchors, no "not found") |
| `/collections/mycology` | 200 | full render (100 anchors, hero title) |
| `/collections/sacred-texts` | 200 | renders |

Response line on a bad book slug is now `HTTP/2 404` (was `200`). Real pages still stream their skeleton via the inner `<Suspense>` and render fully.

Also added `tests/unit/soft-404-loading-guard.test.ts` pinning the absence of the three `loading.tsx` files, so re-adding one can't silently reintroduce the soft-404 (3 tests pass).

Issue #3293, opened 2026-07-20
books.pages_count has two conflicting conventions (hidden pages counted or not)


JDerekLomas:
## The bug

`books.pages_count` / `pages_ocr` / `pages_translated` mean two different things depending on which writer last touched the book:

- **Visible-only** — pages with `page_number > 0`. This is what the read path assumes: `src/app/book/[id]/page.tsx` prints `pages_count` as "N scans" and divides by it for `hasTranslations`, the ≥90%-readable filter, and the `TranslatedSiblingNotice` <5% gate. Negative `page_number` is a deliberate soft-hide — those pages never render.
- **All pages** — what the pipeline's counter writers actually compute. None of these filter on `page_number`:
  - `scripts/batch/collect-batch-results.mjs`
  - `scripts/workers/batch-collector.mjs`
  - `scripts/batch/collect-multipage-ocr.mjs`
  - `scripts/workers/batch-split-bph.mjs`
  - `scripts/batch/realtime-translate.mjs`

Sampling 400 visible books that have hidden pages: **387 store the visible-only count, 13 store the all-pages count.** So the read path's assumption is right and the writers are the minority — but any book those writers touch after a split/hide flips to the wrong convention.

## Observed damage

Reported by Derek 2026-07-20: [/book/histoire-de-la-magie-...-constant](https://sourcelibrary.org/book/histoire-de-la-magie-avec-une-exposition-claire-et-precise-constant) showed "This edition is not yet translated" while the reader served English. Stored `pages_translated: 20`; true visible count 581 of 620. It also displayed "929 scans" for 620 readable pages — 929 is the all-pages number.

A gate-scan of all 1,918 books that trip the <5% notice found **12** with a genuinely false banner. All 12 are repaired (visible-only rule), revalidated, re-synced, purged; a re-scan now returns **0**. The largest were *De re metallica* (592/600 stored as 0), Agrippa (617/624), the Pymander (465/472), Suetonius (683/695), Celsus (609/615).

## What to fix

1. Add `page_number: { $gt: 0 }` to the $match in every writer listed above, so they stop producing the minority convention.
2. Decide whether "N scans" should ever include soft-hidden pages (I don't think so — they're unreachable), and pin it with a test.
[...]

PR #3297, opened 2026-07-21
docs: pin the visible-only page-count invariant + postmortem

JDerekLomas:
Follow-up to #3288 / #3293 / #3295 — the doc half, so the lesson doesn't live only in a merged PR body.

## Two invariants added under **Visibility & Stats Invariants**

1. **`pages_count` / `pages_ocr` / `pages_translated` count visible pages only** (`page_number > 0`), and every writer goes through `scripts/lib/page-counts.mjs` (#3295). Five writers used to `$match` on `book_id` alone, which flipped any book they touched after a split/hide to an all-pages count.

2. **Before batch-correcting a denormalized counter, check whether the gap equals a known subset — and read a few of the "recovered" units.** A gap that exactly matches the hidden-page count is a definitional difference, not staleness.

The second one is the expensive lesson. Fixing this by mirroring the pipeline's own writer (rather than the read path that renders the number) wrote 13,274 phantom scans and 12,852 hidden-page translations to 88 books — making 66 books whose "not yet translated" banner was *correct* look translated — off an aggregate of "209,130 pages missing" that was arithmetic over a wrong predicate. Live ~30 min, reversed. The true numbers were 12 false banners and 3,858 pages.

## Also

`.claude/handoffs/2026-07-20-page-counter-convention-conflict.md` — full postmortem, added with `git add -f` per the handoff policy. Technical only: no PII, secrets, or business material.

## Still open

Residual corpus-wide drift is **unmeasured** — books touched by the old writers before #3295 stay on the wrong convention until something recounts them. `recount-page-stats.mjs --stale` does the sweep; run it on Hetzner where a long aggregation won't get killed.


JDerekLomas:
**Claim re-verified in the 2026-08-04 sweep — still true, still needed. Please rebase rather than close.**

The skill's warning about conflicting docs PRs is that the doc moved underneath and the assertion may now be false. Checked both halves:

- `scripts/lib/page-counts.mjs` **exists on main** — the code half shipped.
- `grep` for the invariant in `CLAUDE.md` returns **0** — the doc half did not.

So this is the exact failure this PR exists to prevent: the lesson living only in a merged PR body. It conflicts because `CLAUDE.md` has moved a lot (I added a request-path-query invariant to it earlier today), not because it is wrong.

JDerekLomas:

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. The code half landed as #3295 (`scripts/lib/page-counts.mjs` states the visible-only rule in its header) and the doc family this belongs to now lives in `.claude/docs/invariants/visibility-and-stats.md` (#3747 'the numerator has to exclude what the denominator excludes'). The CLAUDE.md anchor this appends to no longer exists (738ddecc split, ~60 commits since). The one sentence still missing — 'a gap equal to a known subset is that subset' — is a three-line PR against `visibility-and-stats.md` if anyone wants it; the handoff goes to the private ops repo.

Issue #3307, opened 2026-07-21
1,446 undated Bhutanese manuscripts carry a fabricated year 1700 (reader-facing + citation metadata + ngram axis)


JDerekLomas:
## What

`books.year = 1700` holds **1,557 books with `pages_count > 0`** against ~50 in each neighbouring year — a 30x spike. **1,446 of them are one cohort**: Tibetan manuscripts from the British Library's Endangered Archives Programme, imported April 2026 across three projects (EAP105: 726, EAP310: 436, EAP039: 283). They carry `published: "1700"` as a literal string and `year: 1700`.

**The source has no date at all.** The EAP IIIF manifests (e.g. `https://eap.bl.uk/archive-file/EAP105-2-4-16/manifest`) expose Identifier / Title / Extent / Place / Citation / Digitised by / Language / Scripts — and **no Date field**. These are undated Bhutanese manuscripts. `1700` was not a rounding of a real date, it was invented at import time. `field_provenance` records nothing for `year` (it only carries the later `is_first_translation` bulk assumption), so nothing marks it as a guess.

This is *not* the same as the ordinary round-year inflation elsewhere. 1450/1500/1550/1600/1650/1750/1800/1850/1900 are all mildly elevated too, but each is diffuse across many providers and languages — that's normal cataloguer "circa" rounding and should be left alone. 1700 is the only year that is a single fabricated stamp.

## Why it matters (in priority order)

1. **It is published as a bibliographic claim.** 1,445 of these are `visible: true`. The book page renders "(1700)" in the title, the description, and in **Google Scholar citation metadata**: `<meta name="citation_publication_date" content="1700">`. A scholar citing one of these gets a date we made up. This is the authored-citation-integrity rule in CLAUDE.md applied to metadata rather than to quoted text — same failure, different surface.
2. **It distorts every year-indexed surface.** The cohort is 279,275 pages / 259,588 translated pages, so year 1700 carries **104M tokens in the `en` ngram corpus — 8x the 99th percentile and 32x the median**. That single column was flattening the entire ngram backdrop before PR #3306 clipped it. It also feeds `/explore/timeline`, the year filters on search (#3267/#3268), and the USTC coverage panel.
3. **788 of them carry `is_first_translation: true`** (set by `bulk-flag-tibetan-ft.mjs` on a `bulk_assumption` / medium confidence basis). A first-translation claim on a manuscript we have also misdated is two soft claims stacked.

[...]

PR #3312, opened 2026-07-22
fix(citations): stop emitting free-text junk as a publication date

JDerekLomas:
Addresses the citation-integrity half of #3307; the data half needs a sign-off (below).

## The bug

`books.published` is FREE TEXT (CLAUDE.md says so explicitly). It was being emitted **verbatim** as a machine-readable citation date. Measured against production: of 33,316 visible books with a `published` value, **7,652 (23%) are not a year**:

| count | value |
|---|---|
| 704 | `""` |
| 703 | `"Unknown"` |
| 373 | `"Ur III / Old Babylonian (c. 2100–1600 BCE)"` |
| 186 | `"[19th century]"` |
| **1,467** | **raw Wikidata QuickStatements**, e.g. `"1573date QS:P571,+1573-00-00T00:00:00Z/9"` |

Verified live before the fix on `/book/feast-in-the-house-of-levi`:

```
<meta name="citation_publication_date" content="1573date QS:P571,+1573-00-00T00:00:00Z/9"/>
<meta property="og:title" content="The Feast in the House of Levi (1573date QS:P571,+1573-00-00T00:00:00Z/9)"/>
```

The same string reached the **BibTeX `year` field and key** and the inline citation in both quote-API twins — i.e. straight into bibliographies people paste into papers, and into the headline of every social share.

## The fix

`src/lib/publication-date.ts` splits two questions that were conflated:

- **`displayPublished()` — what may we SHOW?** Keeps hedged-but-informative values (`15th century`, `circa 1700`); strips QuickStatements tails; drops non-answers (`Unknown`, `n.d.`, empty).
- **`citationYear()` — what may we ASSERT to a machine?** A bare year only.

**A century or a range is real knowledge but it is not a publication date**, and these fields cannot express the hedge. Flattening `"15th century"` into `1450` would fabricate a precision the source never had, so we emit nothing instead. `citationYearOrNd()` supplies the scholarly `n.d.` convention where BibTeX needs a value.

Deliberately **does not** fall back to the numeric `books.year`: that field is derived and sometimes an import placeholder (#3307). Keying strictly off `published` means correcting `published` is sufficient to stop the claim.

Wired into: book-page metadata (`citation_publication_date`, `og:title`, `description`) and both `quote` route twins (inline citation, BibTeX year, BibTeX key — previously emitting keys containing spaces and commas).

## Out of scope — deliberately

- **The #3307 cohort is NOT fixed by this.** 1,446 undated Bhutanese manuscripts carry `published: "1700"`, which is *well-formed* — a format-based fix cannot detect a fabricated-but-plausible value. That needs a data change, so this PR ships `scripts/maintenance/clear-fabricated-eap-dates.mjs` (dry-run by default, writes a backup, supports `--revert`) **without running it**. Dry run: 1,445 matches, all visible, all `published: "1700"`, split EAP105=726 / EAP310=436 / EAP039=283.
- **91 files reference `.published`.** This covers the citation-bearing surfaces. EPUB front matter, DOI minting, KDP and `BibliographicInfo` display should be swept separately rather than bundled here.

## Verification

`npx tsc --noEmit` clean. 11 new unit tests pinned to the real production strings above.


JDerekLomas:
Heads-up for whoever picks this up: **`test` is failing and the branch now conflicts with `main`** (`mergeStateStatus: DIRTY`) as of 2026-07-29. Needs a rebase plus a look at the test failure before it can be judged.

Also worth re-measuring the data half before acting — a sweep of the `year=1700` cohort today returns **1,660** books, against the 1,446 in #3307.

JDerekLomas:
Merged `main`, resolved, and fixed the failing check. Two notes:

**The red X was `tsc`, not a test.** `src/lib/publication-date.ts:32` used the `s` (dotAll) regex flag; tsconfig targets ES2017, where it isn't available — `TS1501`. All unit tests were passing the whole time; the failure was the `npx tsc --noEmit` step inside the same job, which is easy to misread as a test failure from the check name alone. Rewritten as `[\\s\\S]`, equivalent and target-independent.

**The conflict was semantic.** `main` rewrote the book page's metadata block around `buildSeoTitle`/`buildSeoDescription` (`src/lib/book-seo.ts`). I took main's structure — but those helpers interpolate `year` **raw**, so merging them as-is would have reintroduced the exact bug this PR exists to fix, in a new location. Both builders now receive `displayPublished(book.published)`.

Worth a reviewer's eye: that's the second place free-text `published` has leaked into a user-facing string. If a third appears, the sanitizer probably belongs inside `book-seo.ts` rather than at each call site.

`tsc` clean · 950/950. Separately, the data half is still worth re-measuring — the `year=1700` cohort reads **1,660** today against the 1,446 in #3307.

PR #3314, opened 2026-07-22
fix(reader): stop browser translation from crashing the page

JDerekLomas:
## The report

An Italian reader emailed: *"Il Pimandro di Ermete Trismegisto — più volte ci ho provato, ma si blocca dopo poche pagine, oppure rimane sempre sulla stessa."* Tried it with and without an account, on a second PC, and on their phone. Their screenshot is our reader error boundary, machine-translated into Italian by Chrome.

The Chrome translate icon in that screenshot is the whole diagnosis.

## Root cause

Chrome/Edge's built-in translator (and the Google Translate widget) replaces every text node in the document with a nested `<font>` pair. React still holds a reference to the **original** text node, so its next commit calls `removeChild` / `insertBefore` with a node that is no longer a child of the parent React recorded. The DOM throws `NotFoundError`, React re-throws out of the commit phase, and the nearest error boundary replaces the page.

In the reader that is: open a book, turn two or three pages, get an error screen. On every book, every device, for as long as auto-translate is on — which for a non-English-speaking audience is most of the time.

**Reproduced** by simulating the translator's DOM rewrite on `/book/il-pimandro-hermes/page/...` (wrap each text node in `<font><font>`), then clicking *Next page* once:

```
NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
Page editor error: NotFoundError: ...
```

— and the rendered page is pixel-for-pixel the reader's screenshot. With the guard installed via an init script, the same sequence turns 5+ pages with correct, changing page text and no error.

Same failure family shows in `application_errors` beyond the reader: `Cannot read properties of null (reading 'parentNode')` (1,046 in 7d), `insertBefore ... is not a child of this node` (84), `removeChild` (23), React #418 hydration-text mismatches (595 on reader URLs in 14d) — all concentrated on text-heavy routes (collections, gallery, reader).

## The fix

Guard `Node.prototype.removeChild` / `insertBefore` so a call whose node has been re-parented out from under React becomes a no-op instead of a throw (the long-standing React issue #11538 remedy). React's virtual tree stays authoritative and the next render re-syncs; the translator re-translates the new text as it always does. **Calls that would have succeeded are completely untouched** — this only ever converts a hard crash into a normal update.

It ships as an inline `<head>` script next to the existing `VIEW_MODE_INIT_SCRIPT`, because the translator can rewrite the DOM before the React bundle has even parsed; patching from a client component would be too late for the hydration commit. `window.__slTranslateGuardHits` counts swallowed calls for support triage.

## Also in this PR

- **Reader error boundary copy.** It said *"Unable to load page editor — Something went wrong while loading this page. Your edits are safe."* to readers who were not editing anything. Now: *"This page didn't load — Something went wrong while opening this page of the book. Try again, nothing is lost."*
- **The boundary now reports.** Route-level `error.tsx` is handled by Next.js before the global `ErrorReporter` boundary, so it only ever `console.error`'d — the reader's single most visible failure was invisible in `application_errors`. That is why this ran unmeasured. It now routes through `reportError` as `reader_error_boundary`.
- Unit test pinning the guard's swallow/pass-through behaviour and that the layout still ships it. No new dependency (there is no DOM environment in the unit suite, so it evaluates the real script string against a minimal `Node.prototype` stand-in).

## Out of scope

- Serving our own Italian edition of the reader UI. The right long-term answer for non-English readers, but a separate piece of work — this PR just stops their browser's translation from breaking the page.
- The `$RS` / React #418 hydration noise on collections and gallery pages. The guard should reduce it, but it is not verified here and deserves its own measurement.

### DIFF
--- a/src/app/book/[id]/page/[pageId]/error.tsx
+++ b/src/app/book/[id]/page/[pageId]/error.tsx
@@ -3,6 +3,7 @@
 import { useEffect } from 'react';
 import Link from 'next/link';
 import { AlertTriangle, RefreshCw, ArrowLeft } from 'lucide-react';
+import { reportError } from '@/components/providers/ErrorReporter';
 
 export default function PageEditorError({
   error,
@@ -11,8 +12,18 @@ export default function PageEditorError({
   error: Error & { digest?: string };
   reset: () => void;
 }) {
+  // Report to application_errors. Route-level error boundaries are NOT caught by
+  // the global ErrorReporter boundary (Next.js handles them first), so without
+  // this the reader's single most visible failure was invisible in the logs —
+  // the browser-translation crash ran for months unmeasured.
   useEffect(() => {
-    console.error('Page editor error:', error);
+    console.error('Reader page error:', error);
+    reportError({
+      message: error.message || 'Reader page error',
+      stack: error.stack,
+      source: 'reader_error_boundary',
+      componentStack: error.digest ? `digest:${error.digest}` : undefined,
+    });
   }, [error]);
 
   return (
@@ -22,10 +33,10 @@ export default function PageEditorError({
           <AlertTriangle className="w-8 h-8 text-accent-rust" />
         </div>
         <h2 className="text-xl font-semibold text-stone-900 mb-2">
-          Unable to load page editor
+          This page didn&apos;t load
         </h2>
         <p className="text-stone-600 mb-6">
-          Something went wrong while loading this page. Your edits are safe.
+          Something went wrong while opening this page of the book. Try again — nothing is lost.
         </p>
         {process.env.NODE_ENV === 'development' && (
           <pre className="text-left text-xs bg-stone-100 p-4 rounded-lg mb-6 overflow-auto max-h-32 text-red-700">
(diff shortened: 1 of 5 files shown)


JDerekLomas:
## Second commit: the guard was only half the fix

Tested the guard alone on the preview against a translator model, and it stopped the crash but **not the staleness** — pages 3, 4 and 5 all rendered page 3's text, page 7 rendered a splice of two. That is the reader's second symptom (*"rimane sempre sulla stessa"*), so the first commit would have shipped a half-fix.

Cause: the guard stops React's mismatched DOM call from *throwing*; it does not make the update *arrive*. Once the translator owns the text nodes, React's writes land on nodes that are no longer in the document.

`6bed2cfe` keys the reader's panel container by page id **while a translator is detected**, so each page builds fresh nodes instead of React patching translator-owned ones. Untranslated readers get `key={undefined}` — the existing in-place path, unchanged. Panel toggles, font size and trace mode live above the key in component state, so they survive the remount; I deliberately did not key the whole reader, which would have reset a translating reader's panel choices every page.

## Verification

The Google Translate *widget* is blocked by our own CSP (`translate-pa.googleapis.com` is not in `script-src`), so it can't be driven from CDP. Chrome's **built-in** translator is browser-level and not subject to page CSP, so real users are unaffected by that block — but it also means the widget was not available as a test harness. Instead: a model that replicates what the translator does to the DOM (nested `<font style="vertical-align: inherit">` per text node, MutationObserver-driven, `translated-ltr` + `<html lang>` set), applying batches **asynchronously** rather than synchronously inside React's commit, which is the behaviour that matters.

| | first page turn | 10 page turns vs untranslated control | guard hits |
|---|---|---|---|
| production (no fix) | **error boundary** | — | — |
| preview (both commits) | fine | **10/10 exact match**, all translated | **0** |

Zero guard hits on the preview is the meaningful number: React never even attempts a mismatched operation any more, so the guard is belt-and-braces rather than load-bearing.

An earlier, cruder model (synchronous surgery inside React's commit — something no real translator does) showed residual staleness on the preview. That was an artifact of the model: the production control passes/fails identically under the async model, so the harness still discriminates.

**Not verified:** a real Chrome-built-in translation session end to end. Worth one manual pass on prod after deploy — open a book, let Chrome offer Italian, turn ten pages.

Issue #3361, opened 2026-07-26
Encyclopedia/index page citations are fabricated: entity mentions smeared across whole batch page ranges (only ~22% real)


JDerekLomas:
## The report

Feedback `6a6553991bea89b3a6e94279` (2026-07-26, Derek, on `/encyclopedia/Matthiolus`): **"What is this page about? Real?"** — arrived at the index/encyclopedia page from an author page and expected to see the author's *books*.

Two separate problems. The second one is a data-integrity bug of the same family as the #2232 misquote.

---

## Problem 1 — the page attributions are fabricated at page granularity

`/encyclopedia/<name>` renders `entities.books[].pages` as a row of clickable `p. N` links under each book ("Appears in 117 Books"). Those page numbers are **not** where the entity appears.

`scripts/workers/enrich-worker.mjs:542-556` (and its batch twin `scripts/batch/batch-generate-indexes.mjs:327-344`) expand each ~50k-char Gemini batch's **whole page range** and attribute *every* person/place/concept found anywhere in the batch to *every page* in the range:

```js
const pageNumbers = Array.from(
  { length: batch.pageRange.end - batch.pageRange.start + 1 },
  (_, i) => batch.pageRange.start + i
);
for (const person of batch.people) peopleMap.get(person).push(...pageNumbers);
```

So one appearance in a 12-page batch becomes 12 page citations.

**Measured** (30 random `type:'person'` entities with a `wikidata_id` and `book_count >= 2`, first book entry each, up to 20 claimed pages, name-stem match against `pages.ocr.data` + `pages.translation.data`):

- 415 claimed page attributions → **91 pages actually contain the name = 21.9%**
- 3/30 entity-book pairs had **zero** real hits among their claimed pages
- The stems used were deliberately loose (`paul`, `juli`, `fran`, `bell`), so 21.9% is an **upper bound** on true precision.

Worked example — `Matthiolus` in *Raphael Explaining the Art of Medicine* (`6909d388cf28baa1b4cb0103`, 157 pages):

| source | pages claimed |
|---|---|
| `entities` (batch smear) | 47–58 (12 consecutive) |
| `books.index.vocabulary` (tag extraction) | 23, 51 |
| **actual text** (`matthiol\|mattiol` over ocr + translation, all 157 pages) | **44, 100** |

Neither derivation is right, and they don't agree with each other. Every one of `p. 47 … p. 58` is a link to a page that does not name Matthiolus.

[...]

PR #3379, opened 2026-07-27
fix(index,embeddings): entity-index correctness — a fifth unfixed writer, a composer that dropped every entity, and the cost constants

JDerekLomas:
Three fixes to the entity-index subsystem, in **two independently reviewable commits**. All were found while sizing a proposed corpus-wide index re-extraction; two of them are why that re-extraction is now on hold.

> **Reviewer note:** `fa1955cd` (cost/model) and `0f143e0f` (the two defects) touch disjoint files and can be reviewed or reverted separately. Happy to split into two PRs if preferred — they are grouped here because they are the same subsystem and were found in one investigation.

---

# Commit 1 — `fa1955cd` · run batch index extraction on flash-lite, and fix the cost constants

`scripts/batch/batch-generate-indexes.mjs` hardcoded `gemini-3-flash-preview`. The enrich-worker's Phase 6 — another writer of the same `book.index` — runs `gemini-3.1-flash-lite`, which is what CLAUDE.md specifies for summary+index.

**The preview model bought nothing.** The extraction prompt returns bare name lists, and page attribution is done locally by `attributeEntityPages()` with no inference call. Identical work, ~7x the price.

**The cost constants are not cosmetic.** `INPUT_COST_PER_1M` / `OUTPUT_COST_PER_1M` are what land in `gemini_usage.cost_usd`, so a stale rate silently misreports every spend estimate derived from that collection. This is concretely how a re-extraction estimate came out at ~$521: pricing historical index token volume at preview rates gives $528.34, while the true logged cost of **all** index extraction to date is **$74.27** across 3,572 rows and 2,792 books. Solving for the implied rates on those rows reproduces the logged cost to the cent at $0.075/$0.30 — confirming the work actually ran on lite.

Measured unit economics from those 3,572 rows (1,211,994 translated pages): **659.8 input** and **39.3 output** tokens per translated page. Over the 4,877,782 translated pages in scope:

| model | est. full-scope cost |
|---|---|
| gemini-3.1-flash-lite | **~$299** |
| gemini-3-flash-preview | ~$2,184 |

Anyone extrapolating later should use the **per-page** rate, not the per-book mean — already-indexed books average 434 pages against a corpus average of 274, so the per-book mean ($0.0266) overshoots to $473.

---

# Commit 2 — `0f143e0f` · two silent defects

Neither is visible without reading the data. Each produces well-formed output that says nothing.

## 2a. A fifth writer of `entities.books[]`

`CLAUDE.md` documents **four** writers and warns that a fix must land in all of them. There are **five**. `src/app/api/[tenant]/books/[id]/index/route.ts` is a near-complete twin of the global index route and never received the #3361 fix that #3363 applied to the fourth. Until this PR it still:

- credited **every page of a ~50k-char Gemini batch to every entity in it** — the original smearing bug, of which ~22% of resulting citations were real;
- `$addToSet`'d whole book subdocuments, so re-indexing a book appended a *second* entry rather than replacing one;
- computed `total_mentions` by summing raw page-array lengths across those duplicates.

**It is reachable.** The route's `GET` auto-generates an index on demand, and **110 visible tenant books currently have none** (93 on the largest tenant, 17 on the second). A reader opening one on a partner subdomain fired this writer against the *global* `entities` collection.

Fixed by porting the global route's implementation: `attributeEntityPages` with the section-precision fallback, `$pull`+`$push`, `entityCounters`.

## 2b. `book_embeddings` discarded every entity and topic term

`composeBookEmbeddingText` read `.name` on people/places/concepts. `buildConceptIndexFromBatches` emits `{ term, pages, page_precision }`. Every element mapped to `undefined`. Separately, topic terms were keyed on `entries` — a legacy shape present on ~1,017 of 18,243 `book_indexes` docs — while the extractor actually writes `vocabulary` and `keywords`.

Measured against live Supabase before this change:

| | rows |
|---|---|
| `book_embeddings` total | 36,078 |
| containing the literal line `People: , , , ,` | **14,237** |
| containing `Concepts: , ,` | **14,318** |
| containing any `Topics:` line | **1,019** |

So the book-level semantic lane in `/api/search` and `/api/search/unified` has been matching on title, author, summary and section headings alone — discarding every entity and term the extraction pipeline produced.

The composer existed in **three copies** (enrich-worker, backfill migration, search eval spike) and all three carried both bugs, so a fix in one would not have reached the others. Collapsed to one implementation in `scripts/lib/book-embedding-text.mjs`, imported by all three, which also exports the Mongo projection so a missing field can't quietly empty a line.

Verified on live index data — books that previously produced empty comma runs now compose full lines (e.g. *Famous Women*: `People: Sulpicia, Irene, Circe, Camilla, Penelope, Lavinia, Dido…`).

---

## Out of scope, deliberately

- **Re-embedding the corpus.** This PR changes what gets *written*; the 14,237 degraded rows stay degraded until re-embedded. Paid run, separate decision.
- **The `entities` repair sweep.** Still unfinished — 62% of entity→book rows carry no `page_precision` (sampled 20k entities) and are demoted at read time.
- **The book index panel**, which reads `book.index.entries` — present on 5.6% of `book_indexes` docs, so the panel is absent from ~17,200 book pages. Real, but it must land *behind* the repair sweep or it renders fabricated citations at scale.
- **Deduping the ~2,800 lines** shared between the global and tenant index routes. That duplication is what caused 2a; collapsing it is its own PR with its own risk profile.
- **Other scripts still on `gemini-3-flash-preview`** (dedup, chapter re-extraction, artwork classification, page splitting, gutter detection, curation) — several are vision calls where the model choice is load-bearing, and none write the entity index.

## Verification

- `npx tsc --noEmit` clean. 744 unit tests pass (61 files).
- `gemini-3.1-flash-lite` resolves against the live API (the `-preview` suffix is retired and 404s).
- **Both new guards negative-controlled against the unfixed code**: reverting the composer fails 8 of the 14 new assertions; restoring the pre-fix tenant route from `HEAD` fails 4 of the writer-set assertions. A guard that passes on the broken code proves nothing.
- `tests/unit/book-embedding-text.test.ts` asserts on *content*, not shape, because the failure mode is well-formed empty output.
- `tests/unit/entity-page-attribution.test.ts` now asserts the writer **set** — CLAUDE.md's remedy was "grep for the writer set before declaring one fixed"; this runs that grep every time. It permits both legitimate write shapes: extractors verify pages and `$pull`/`$push`; the rebuild route has no page text, so it carries the precision marker forward and `$set`s a whole array keyed by `book_id`.

`scripts/**` changes go live via the Hetzner auto-pull; the tenant route change needs a prod deploy.

Related: #3361, #3363, #3377


JDerekLomas:
## The cost constants are duplicated across ten files, and two lanes disagree by 3.3×/5×

Found while auditing for silent-drift defects (PR #3569). This is a concrete mechanism for "`cost_usd` fabricates estimates."

`gemini-3.1-flash-lite` is priced two different ways for the same model:

| rate (per 1M tokens) | files |
|---|---|
| **0.075 / 0.30** | `scripts/workers/enrich-worker.mjs` (OCR), `src/lib/ai.ts`, the whole eval lane |
| **0.25 / 1.50** | `scripts/workers/translate-worker.mjs`, `src/lib/gemini-logger.ts`, `scripts/workers/batch-collector.mjs`, `scripts/workers/lib/supabase-usage-logger.mjs`, `scripts/maintenance/reconcile-batch-usage.mjs`, `scripts/maintenance/backfill-batch-costs.mjs` |

3.3× on input, 5× on output. Consequences:

- **An OCR row and a translation row for the same model are costed differently.** `gemini_usage.cost_usd` is a blend of two rates depending on which worker wrote it, so any total over mixed phases is not a quantity with a single definition.
- **Two files inside the app disagree with each other** — `src/lib/ai.ts` vs `src/lib/gemini-logger.ts`.
- Nothing surfaces it, because a cost figure always looks plausible. Same shape as the stale prompt pin in #3569: a local copy asserting an external truth, drifting silently.

**Neither rate fits the stored data.** Solving 400 recent realtime `gemini-3.1-flash-lite` rows for their implied rate gives a mean relative error of **0.39** against 0.075/0.30 and **1.83** against 0.25/1.50. So recorded costs were computed with something near the first pair, but not cleanly — cached-token pricing or a third historical rate is the likely remainder. Worth noting the second pair would overstate by ~180%.

**What #3569 does and deliberately does not do.** `scripts/lib/model-pricing.mjs` is now the single table, and it does **not** claim the values are verified against Google's price list. The disputed rate is preserved in `DISPUTED_LEGACY_PRICING` so historical `cost_usd` stays reproducible. Eval-lane consumers are migrated (estimates only, safe).

**The production writers of `cost_usd` are untouched on purpose** — changing a constant there silently rewrites what the dashboard reports for historical rows, which is this issue's decision to make, not a refactor. They are named individually in the allowlist of `tests/unit/model-pricing-single-source.test.ts`, which fails on any *new* inline price table, so the set cannot grow quietly while this is open.

Open questions for this issue: which rate is correct against Google's published pricing (including batch's 50% discount and cached-token tiers), and whether historical rows should be recomputed or left as-is with a documented caveat.

Issue #3049, opened 2026-07-06
Gallery viewer 200s on orphaned artwork image ids (dead-end, broken image)


JDerekLomas:
**Reported via footer feedback (Julika, 2026-07-03):**
`/gallery/image/artwork-5d4029957171a78dfa6fd785-0` — "another dead end and broken imGe"

## Finding
The id `5d4029957171a78dfa6fd785` resolves to **no record** anywhere — not in `books` (by `_id` or `id`) and not in `gallery_images` (by `_id` or `book_id`). Yet the page returns HTTP **200** because the `/gallery/image/[id]` viewer layout rescues unknown ids via `NEXT_REDIRECT` (PR #2958, `lesson_gallery_image_id_namespaces`). So a status-code check looks healthy while the reader sees a broken/empty image.

This is the orphaned clip-prefix / stale-id class: `artwork-`, `gallery-`, `cover-` prefixed ids and bare ObjectId hex that no longer (or never) map to a materialized `gallery_images` row.

## Suggested action
- Where an id genuinely resolves to nothing, the viewer should render a clean 404 (or redirect to `/gallery`) rather than a 200 empty shell.
- Audit `gallery_images` for how many inbound `/gallery/image/*` links point at ids with no backing row (crawl `not_found_reports` / sitemap for `artwork-*` ids), and normalize/purge at the producers so these links aren't emitted in the first place.

---
Feedback-ID: 6a473930a5f38bf2d79673bf

JDerekLomas:
Addressed in #3089. The viewer layout now:
- verifies the target book exists before redirecting an `artwork-`/`cover-<bookId>` id to `/book/<bookId>` (no more bouncing into a second dead page), and
- renders a clean 404 for a bare `<pageId>-<n>` id only when it resolves in **neither** page detections **nor** `gallery_images` (mirrors `/api/gallery/image/[id]`, so orphaned-page images the API still serves are preserved).

Note: the specific reported id `5d4029957171a78dfa6fd785` now backs a real Leonardo artwork record (imported after the report), so it redirects correctly to `/book/…`; the fix handles the genuinely-orphaned class. A producer-side audit to stop emitting orphan ids is a separate follow-up.

Will close after merge+deploy.

JDerekLomas:
[...]

PR #3381, opened 2026-07-27
fix(seo): real 404s for unknown gallery-image, artwork, artist and author slugs (#3376)

### LINKED ISSUE #3376: Soft 404: /gallery/image/<nonexistent> returns HTTP 200 with the not-found body
Split out of #3370 so it isn't lost when that closes. Pre-existing, unrelated to tenant work.

## The bug

```
/gallery/image/anything                        http=200  notfound-marker=1
/gallery/image/6978f8c9951b55be75c3f116-1      http=200  notfound-marker=0
```

A nonexistent gallery-image id renders Next.js's not-found body but the response status stays **200**. A real image id renders normally, so it's the missing-record path, not tenant-related.

## Why it matters more than it looks

This is the exact failure mode CLAUDE.md warns about twice:

- Under quote/link validation: *"Validate links by grepping the response body for a not-found marker, NOT by status code. `curl -w '%{http_code}'` returns 200 for these soft-404s."* Every link checker, ours and Google's, reads this as a live page.
- Under the ISR Suspense invariant: the likely cause is a Suspense child calling `notFound()` after streaming began, so the status was already committed. `scripts/audit-bph-leaks.mjs` detects it only via the `NEXT_HTTP_ERROR_FALLBACK;404` digest baked into the HTML.

Consequence: dead gallery URLs are indexable and unreportable. The `not_found_reports` telemetry can't see them either, since nothing 404s.

## Fix direction

Resolve the image record **before** the streaming boundary so `notFound()` runs while the status is still settable, or return an explicit 404 from the route. Check `/artwork/[id]` and `/book/[id]/page/[pageId]` for the same shape — the reader route has a known soft-404 (documented in CLAUDE.md: `/book/<slug>/page/<number>` returns 200 with "Page Not Found").

## Verifying

```
curl -s https://sourcelibrary.org/gallery/image/definitely-not-real -o /tmp/x -w '%{http_code}\n'
grep -c 'NEXT_HTTP_ERROR_FALLBACK;404' /tmp/x
```
Should be `404` and `0`. Today it's `200` and `1`.

`node scripts/audit-bph-leaks.mjs` also reports it under "Silent 404s" and will go quiet when fixed — it's currently the only issue that audit still finds on the BPH subdomain (foreign books are now 0, down from 102 in #3364).

JDerekLomas:
Fixes #3376.

## Problem

`/gallery/image/<nonexistent>` returns **HTTP 200** with the Next.js not-found body. Checking the neighbours the issue asked about turned up three more of the same shape. Measured against production:

| URL | status | 404 digest in body |
|---|---|---|
| `/gallery/image/definitely-not-real-9` | **200** | yes |
| `/gallery/collections/definitely-not-real` | **200** | yes |
| `/artwork/definitely-not-real` | **200** | yes |
| `/author/definitely-not-real` | **200** | yes |
| `/book/definitely-not-real` | 404 | — (fixed in #3277) |
| `/collections/definitely-not-real` | 404 | — (fixed in #3277) |

Fresh renders, not cache artefacts (`x-vercel-cache: MISS`).

## Root cause

The mechanism #3277 documented, in four more places: a `loading.tsx` at **or above** a route wraps it in an automatic `<Suspense>`, so Next.js flushes the 200 loading shell as soon as the data fetch suspends — **before** `notFound()` runs. Once the status is committed, `notFound()` can only swap the body.

`/gallery/image/[id]` is the instructive case: its layout has had a correct existence gate since #3049 (`imageResolves` → `notFound()`), and that gate was **silently defeated** by `src/app/gallery/loading.tsx` — three segments up, written for the gallery index grid. A correct check in the right place still fails if any ancestor opens a boundary above it.

## Fix

Delete the seven `loading.tsx` files that sit at or above an existence check. No skeleton is meaningfully lost:

- `/gallery` self-streams through its own inner `<Suspense fallback={<GalleryShell/>}>` — the segment file was redundant for the index and harmful for everything below it
- `/artwork` is `revalidate = false` (fully static)
- `/gallery/collections`, `/artwork/[slug]`, `/artwork/artist/[slug]`, `/author/[name]` are ISR (21600–86400s) and serve prerendered

`src/app/gallery/image/[id]/loading.tsx` is deliberately **kept** — it sits *below* the layout that 404s, so it cannot flush ahead of the gate.

The `/author/[name]` route resolves through the author thesaurus (variant slugs 301 to canonical) *before* its `notFound()`; that logic has to stay in the page, which is why the fix here is removal rather than a layout-level gate.

## Guard

`tests/unit/soft-404-loading-guard.test.ts` (added by #3277) is extended from 3 to 10 pinned absences, each with a comment saying what it protects. Re-adding any of these files silently reintroduces the soft-404.

## Verification

`npx tsc --noEmit` clean; guard test 10/10. Curl matrix against the preview once it builds — each of the four must return **404** with no `NEXT_HTTP_ERROR_FALLBACK;404` needed to detect it, and a real image / artwork / author / gallery collection must still return 200 and render.

## Out of scope

- The reader soft-404 (`/book/<slug>/page/<pageId>` with a real-but-missing page id) — still open from #3232's follow-up note. Numeric page numbers are already caught in `proxy.ts`.
- Backfilling `not_found_reports` for URLs that never 404'd.


JDerekLomas:
## Verified on the preview

`https://sourcelibrary-v2-git-worktree-fix-so-d790ab-dereklomas-projects.vercel.app`

**Unknown slugs — all four now return a real 404** (were 200 on production):

| URL | before | after |
|---|---|---|
| `/gallery/image/definitely-not-real-9` | 200 | **404** |
| `/gallery/collections/definitely-not-real` | 200 | **404** |
| `/artwork/definitely-not-real` | 200 | **404** |
| `/author/definitely-not-real` | 200 | **404** |

**Real content unaffected** — every affected route still 200s and renders, including the two index pages whose segment `loading.tsx` was removed:

| URL | status | bytes |
|---|---|---|
| `/gallery/image/6959067195a91542b28bcdce-0` | 200 | 55,964 |
| `/gallery/collections/chinese-woodcuts` | 200 | 663,198 |
| `/artwork/strange-shapes-seen-over-basel-1566-…` | 200 | 57,230 |
| `/author/sarah-price` | 200 | 45,429 |
| `/gallery` (index) | 200 | 401,954 |
| `/artwork` (index) | 200 | 129,942 |

No `NEXT_HTTP_ERROR_FALLBACK;404` digest in any of the 200 bodies. Checks green: test, DCO, Vercel.

PR #3382, opened 2026-07-27
fix(e2e): locate Librarian suggestion chips structurally, not by wording (#3358)

### LINKED ISSUE #3358: E2E tests failed (2026-07-26)
## E2E test failure

**Date:** 2026-07-26
**Run:** https://github.com/Embassy-of-the-Free-Mind/sourcelibrary-v2/actions/runs/30194810316
**Base URL:** https://sourcelibrary-v2.vercel.app

Download the `playwright-report` artifact from the run for details.

---
*Auto-created by the E2E test workflow.*

JDerekLomas:
Fixes #3358.

## What actually failed

`Librarian › clicking suggestion populates input` failed 3 retries deep on the 2026-07-26 scheduled run ([30194810316](https://github.com/Embassy-of-the-Free-Mind/sourcelibrary-v2/actions/runs/30194810316)) and is still reported as **"1 flaky"** on the green 2026-07-27 run. The other failures in that run — the `bph-iframe` strict-mode violations — did not recur and are not touched here.

**The app is correct. The test assumed the chips hold still; they don't.**

`LibrarianClient.pickSuggestions` redraws the suggestion set **twice** after first paint:

1. on mount — the initial set is seeded deterministically on purpose (a random one hydration-mismatches, React #418), so the shuffle is deferred to a `useEffect`;
2. again when the public-threads fetch resolves and blends up to two real visitor questions into the set.

That gave the spec two independent ways to fail, and it hit both:

| # | Assumption | How the redraw breaks it |
|---|---|---|
| 1 | chips are locatable by keyword (`/world soul\|Agrippa\|Kabbalah\|…/`) | the redraw picks 4 of ~30 suggestions at random, so every keyword in the alternation can be absent — the click times out |
| 2 | `first().textContent()` still describes the chip you're about to click | the redraw can swap the chip between the read and the click — this is the reported symptom, `unexpected value "Tell me about the Emerald Tablet"` |

## Fix

- `data-testid="suggestion-chip"` on the chip button — a stable hook, since the wording provably isn't one.
- The spec locates on that testid and asserts the **behaviour**: textarea goes from empty to a whole 4+ word question. *Which* question is deliberately not asserted, with a comment saying why so nobody "tightens" it back into a flake.

## Verification — forced the race instead of waiting for it

Re-running the old test 8× locally passed **8/8**, which is exactly why this read as noise for weeks. So I built a probe that delays `/api/embassy/threads` by 2.5s, putting the redraw *after* the read — the slow-network timing CI hits:

| | old assertion | new assertion |
|---|---|---|
| under forced redraw | **failed 3/3** | passes |
| plain repeat run (no delay) | 8/8 pass | 4/4 pass |

The probe also caught failure mode #1 independently: the old keyword locator timed out on `.click()` when no drawn chip matched the alternation. The probe was throwaway and is not committed.

`npx tsc --noEmit` clean.

## Note for the reviewer

The one app-side change is a `data-testid` attribute — no behaviour, no styling. The e2e suite runs against a **deployed** base URL, so this test only goes green once the attribute ships; expect the scheduled run to still report the old flake until then.

### DIFF
--- a/e2e/embassy.spec.ts
+++ b/e2e/embassy.spec.ts
@@ -23,29 +23,37 @@ test.describe('Librarian', () => {
     await expect(page.getByText('The Librarian searches the collection')).toBeVisible();
   });
 
+  // The suggestion chips are NOT stable content. LibrarianClient redraws the
+  // set twice after first paint (`pickSuggestions`): once on mount, and again
+  // when the public-threads fetch resolves and blends in real visitor
+  // questions. So neither the wording of a chip nor its presence in the set
+  // survives long enough to assert on — both produced the #3358 flake:
+  //
+  //   1. locating chips by keyword (/Agrippa|Kabbalah|.../) — the redraw picks
+  //      4 of ~30 suggestions at random, so a keyword can vanish entirely and
+  //      the click times out;
+  //   2. capturing a chip's text and asserting the textarea matches THAT text
+  //      — the redraw can swap the chip between the read and the click.
+  //
+  // Both are fixed by locating chips structurally and asserting the behaviour.
+  const chips = (page: import('@playwright/test').Page) =>
+    page.getByTestId('suggestion-chip');
+
   test('shows suggestion chips', async ({ page }) => {
-    // Suggestions may vary — just verify at least 2 are visible
-    const suggestions = page.locator('button', {
-      hasText: /world soul|philosopher|Agrippa|Kabbalah|Emerald|Ficino|alchemist/i,
-    });
-    await expect(suggestions.first()).toBeVisible();
-    expect(await suggestions.count()).toBeGreaterThanOrEqual(1);
+    await expect(chips(page).first()).toBeVisible();
+    expect(await chips(page).count()).toBeGreaterThanOrEqual(1);
   });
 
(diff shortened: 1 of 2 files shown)


JDerekLomas:
## Verified on the preview — plus an accidental negative control

Ran the two suggestion specs 4× each against `sourcelibrary-v2-git-worktree-fix-li-21e319…`:

```
8 passed (44.8s)
```

**The accidental part is the useful part.** My first attempt set the wrong env var (`PLAYWRIGHT_BASE_URL`; the config reads `BASE_URL`), so it silently ran against **production** — which does not yet have the `data-testid`. Result: **6 failed, 0 passed**, both specs, every repeat.

That is the negative control this change needed. It proves:

1. the rewritten tests genuinely depend on the new hook rather than passing vacuously — a locator that matched nothing would have been an easy way to write a green, worthless test;
2. the failure is deterministic on a build without the attribute, and deterministic-green on a build with it.

It also means the scheduled e2e run will keep reporting this spec as failing until this merges **and** a production deploy ships the attribute — expected, not a regression.

Checks: test ✅ DCO ✅ Vercel ✅ (the first Vercel build failed and the retry passed, per the CLAUDE.md note about not trusting the first result).

PR #3383, opened 2026-07-27
fix(tenant): restore Account / Sign out on partner subdomains (#2150)

### LINKED ISSUE #2150: Restore the scholar view-options / account menu (EmbedUserMenu) on real tenant subdomains
## Context
Follow-up from the BPH AI-summary work (PRs #2145 → #2147).

While fixing AI-summary hiding on `bph.sourcelibrary.org`, we found that **`EmbedUserMenu` (the floating top-right gear menu) is only mounted by `src/app/embed/layout.tsx`**, which runs only on the legacy path-based `/embed/*` routes. After the tenant-as-filter migration, real tenant **subdomains** (`bph.sourcelibrary.org`, etc.) serve the **global routes directly** (e.g. `/book/[id]`, via `x-tenant-*` headers, no `/embed` rewrite — see `src/proxy.ts`). The root layout wraps those, but it does **not** render `EmbedUserMenu`.

**Verified live (2026-05-28):** `bph.sourcelibrary.org/book/...` has no gear menu in the DOM.

## Why it matters
`EmbedUserMenu` carries two responsibilities (see its header comment):
1. **Scholar view options** — the "Hide AI introductions & summaries" / "Hide reading guide" toggles. With #2147, AI summaries are now hidden-by-default on BPH, but there is **no in-page way to opt back in** on the real subdomain.
2. **Auth** — it is described as "the only sign-out affordance" on embed pages (the `SiteHeader` is stripped). If it never renders on the subdomain, **signed-in users on `bph.sourcelibrary.org` may have no Account / Sign-out control**. This should be confirmed and is the higher-priority half.

## Scope / plan
- Render `EmbedUserMenu` for real tenant subdomains. Because the root layout deliberately avoids `headers()` to preserve ISR caching (site mode is detected client-side, see `SiteModeProvider` / `getClientSiteMode`), mount it in the **root layout** and **client-gate** it: show only on tenant/embed contexts (e.g. `*.sourcelibrary.org` subdomain that isn't the apex, mirroring `isTenantSubdomain` in `EmbedUserMenu.tsx`), and keep it **off the main apex site**.
- Avoid double-mounting on path-based `/embed/*` (it is also rendered by `embed/layout.tsx`). Either gate the root-layout instance to subdomains only, or move the single source of truth to the root layout and remove it from `embed/layout.tsx`.
- The hide toggles already work via the `sl_hide_ai` / `sl_hide_guide` cookies + the root-layout `VIEW_MODE_INIT_SCRIPT` (added in #2147); only the menu UI is missing on the subdomain.

## Acceptance criteria
- Signed-in user on `bph.sourcelibrary.org` sees the gear/account menu with working Account + Sign out.
- The "Hide AI introductions & summaries" toggle appears, starts **checked** on BPH (hidden-by-default), and unchecking it shows summaries and persists across reloads (`sl_hide_ai=0`).
- The menu does **not** appear on the main apex site (`sourcelibrary.org`).
- No ISR cache regression (no `headers()` in the root layout path).

## Files
- `src/components/embed/EmbedUserMenu.tsx`
- `src/app/layout.tsx` (root) / `src/app/embed/layout.tsx`
- Reference: `src/proxy.ts`, `src/lib/site-mode.ts`

## Note
Product decision (2026-05-28): for now BPH is intentionally "hidden, no toggle." This issue is the path to restore the toggle if/when we want the opt-in back — but the **sign-out affordance** half should be validated independently and sooner.

JDerekLomas:
Fixes #2150 — the half the issue flagged as *"higher-priority"* and *"should be confirmed"*. Now confirmed, and fixed.

## Confirmed live

The issue asked whether signed-in users on a tenant subdomain have any account control. They do not. Checked 2026-07-27 against the **hydrated** DOM of `bph.sourcelibrary.org/book/aurora-consurgens-…`, in a browser holding a **valid session** (`/api/auth/session` → `dereklomas@gmail.com`):

```
headerPresent:     false
navHrefs:          null
embedUserMenu:     false
authAffordances:   []      ← zero sign-in/out/account controls anywhere on the page
```

A signed-in reader on a partner subdomain is stranded: no Account, no Sign out, on any page.

## Why it hid for two months

Two individually-correct behaviours compose into the gap, and **`curl` cannot see it** — the server HTML still contains a header, which is removed only after hydration.

After the tenant-as-filter migration a partner subdomain serves the **global** routes directly (`src/proxy.ts` rewrites only `/`, `/catalogue*`, `/catalog/*`). So on `/book/<slug>`:

1. `src/app/embed/layout.tsx` never runs — and it is the **only** mount point for `EmbedUserMenu`, which its own header comment calls *"the only sign-out affordance"* for surfaces where the SiteHeader is stripped;
2. `ConditionalSiteHeader` then strips the SiteHeader post-hydration on a tenant host, taking the header's account control with it.

## The fix

Mount the menu from the **root layout**, client-gated, in a new `account-only` variant.

- **Root layout, not a tenant layout.** Reading `headers()` there would force dynamic rendering and blank the book page's ISR cache (`revalidate = 86400`) — the same constraint that already makes site-mode detection client-side. `useEmbedContext` resolves the host in a post-mount effect, so server HTML stays host-agnostic and cacheable.
- **`account-only` renders nothing for anonymous visitors**, and omits the view-options toggles and BPH catalogue search. Anonymous BPH pages are byte-identical to today. Restoring auth is a bug fix; putting the toggles back is a **separate product decision** (2026-05-28: BPH is deliberately "hidden, no toggle"), so this deliberately does not reopen it — the plumbing is there to flip later.
- **`/embed/*` untouched** — it still mounts the full menu from its own layout, and `TenantAccountMenu` returns `null` there so the two never stack.

## Guard

`tests/unit/tenant-account-menu.test.ts` pins each link in the chain: root-layout mount, inside `<Providers>`, no `next/headers` in the root layout, no double-mount on `/embed/*`, anonymous renders nothing, toggles stay out, Account + Sign out still present. Removing any one restores the bug silently.

725 unit tests pass; `npx tsc --noEmit` clean.

## Verification — read this before approving

**A Vercel preview cannot verify this.** Two independent reasons:

1. CLAUDE.md's known trap — curling a preview with `Host: bph.sourcelibrary.org` makes Vercel's router serve the **production** deployment for that domain, producing a convincing false result.
2. The gate is `isTenantSubdomain(window.location.host)`, and a preview host is `sourcelibrary-v2-git-….vercel.app` — not a `.sourcelibrary.org` subdomain, so the menu correctly does not mount there regardless.

What the preview **can** confirm, and what I'll check on it: `/embed/bph/*` still shows exactly one floating menu, and the apex site is unchanged.

The tenant path must be verified **after deploy, on the real subdomain, while signed in** — same method as the confirmation above. I can run that check once it ships.

## Out of scope

- The view-options toggles on tenant pages (product decision above).
- `EmbedUserMenu`'s `navigateToAccount` comment is now stale — it says the proxy rewrites *every* non-`/embed/` path, which stopped being true. Its behaviour (routing `/account` to the apex, where the shared `.sourcelibrary.org` cookie carries the session) is still correct, so I left it alone rather than widen the diff.

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -11,6 +11,7 @@ import AnalyticsScripts from "@/components/providers/AnalyticsScripts";
 import EmbedLinkInterceptor from "@/components/embed/EmbedLinkInterceptor";
 import EmbedHostNavigationListener from "@/components/embed/EmbedHostNavigationListener";
 import EmbedHistoryPatch from "@/components/embed/EmbedHistoryPatch";
+import TenantAccountMenu from "@/components/embed/TenantAccountMenu";
 import ScrollReveal from "@/components/layout/ScrollReveal";
 
 
@@ -162,6 +163,12 @@ export default async function RootLayout({
           <EmbedLinkInterceptor />
           <EmbedHostNavigationListener />
           <EmbedHistoryPatch />
+          {/* Tenant subdomains serve the global routes with the SiteHeader
+              stripped, which left signed-in readers with no way to reach their
+              account or sign out (#2150). Must sit inside <Providers> for the
+              session, and stays client-gated so the root layout never reads
+              headers() and the book page keeps its ISR cache. */}
+          <TenantAccountMenu />
           <div id="main-content" className="flex-1">
             {children}
           </div>
--- a/src/components/embed/EmbedUserMenu.tsx
+++ b/src/components/embed/EmbedUserMenu.tsx
@@ -101,7 +101,28 @@ function writeCookie(name: string, on: boolean) {
   document.cookie = `${name}=${on ? '1' : '0'}; path=/; max-age=${COOKIE_MAX_AGE}; samesite=lax`;
 }
 
-export default function EmbedUserMenu() {
+export interface EmbedUserMenuProps {
+  /**
+   * `full` (default) — the partner landing/catalogue surface under
+   * `/embed/<tenant>/*`: view-options toggles, BPH catalogue search, and auth.
+   * Rendered to everyone, signed in or not.
+   *
+   * `account-only` — the global routes a tenant subdomain serves directly
+   * (`/book/*`, `/collections/*`, …). Renders NOTHING for anonymous visitors
+   * and only the auth affordance for signed-in ones (#2150).
+   *
+   * The split exists because those pages strip the SiteHeader entirely on a
+   * tenant host, which left signed-in visitors with no Account or Sign-out
+   * control anywhere on the subdomain. Restoring auth there is a bug fix;
+   * putting the view-options toggles back is a separate product decision
+   * (2026-05-28: BPH is deliberately "hidden, no toggle"), so this variant
+   * deliberately does not reopen it.
+   */
+  variant?: 'full' | 'account-only';
+}
+
+export default function EmbedUserMenu({ variant = 'full' }: EmbedUserMenuProps = {}) {
+  const accountOnly = variant === 'account-only';
   const { data: session, status } = useStableSession();
   const [isOpen, setIsOpen] = useState(false);
   const [imgError, setImgError] = useState(false);
@@ -127,8 +148,8 @@ export default function EmbedUserMenu() {
   useEffect(() => {
     setHideAi(readHideAi());
     setHideGuide(readCookie(COOKIE_HIDE_GUIDE));
-    setShowSearch(isBphSurface());
-  }, []);
+    setShowSearch(!accountOnly && isBphSurface());
+  }, [accountOnly]);
 
   // Focus the field as it expands.
   useEffect(() => {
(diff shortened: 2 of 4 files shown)


JDerekLomas:
## Verified on the preview — the parts a preview *can* prove

Preview: `sourcelibrary-v2-git-worktree-fix-te-ac9745…`. Checked against the **hydrated** DOM (3.5s settle), since server HTML is meaningless for a client-gated component.

**1. No double-mount on `/embed/*`** — the risk of adding a second mount point:

```
/embed/bph    floatingMenus: 1    gearButtons: ["View options"]
```

Exactly one control, and it is the **full** menu (`embed/layout.tsx`'s), not the account-only one — `TenantAccountMenu` correctly returns null there.

**2. The global site is untouched:**

```
/book/platonis-opera-plato   floatingMenus: 0   siteHeaderPresent: true
```

No floating control on a non-tenant host, and the normal SiteHeader still renders. The gate does not leak to the apex.

**3. Full unit suite:** 725 passing, including the 7 new guard assertions. `tsc --noEmit` clean. Checks: test ✅ DCO ✅ Vercel ✅ (first Vercel build failed, retry passed — the known first-result behaviour).

## What is NOT verified here, and cannot be

The tenant path itself. Both reasons are structural, not laziness:

1. the preview host is `sourcelibrary-v2-git-….vercel.app`, so `isTenantSubdomain()` is correctly false and the menu will never mount there regardless of the code;
2. curling with `Host: bph.sourcelibrary.org` resolves to the **production** deployment (the CLAUDE.md trap that produced a convincing false negative during #3367).

So the acceptance check is: **after this deploys, on `bph.sourcelibrary.org`, while signed in** — re-run the exact DOM probe from the issue and expect `authAffordances` to be non-empty with Account + Sign out reachable. I can run that as soon as it ships; it is the same probe that produced the `[]` in the issue comment.

PR #3384, opened 2026-07-27
fix(seo): real 404 for unknown reader page ids (#3376 follow-up)

JDerekLomas:
Closes the soft-404 family opened by #3232 and continued in #3376. Follow-up to #3381.

## The last one

Measured on production today:

```
/book/platonis-opera-plato/page/definitely-not-a-real-page-id   200
/book/platonis-opera-plato/page/abc123def456abc123def456        200
/book/curious-physics-hellwig/page/zzzzzzzz                     200
```

All three serve the "Page Not Found" body with the `NEXT_HTTP_ERROR_FALLBACK;404` digest baked in. Purely **numeric** page numbers are already caught by `proxy.ts` and 302 — this is the non-numeric id case that slips past it, and it's the one CLAUDE.md warns about under quote-and-link validation (*"`curl -w '%{http_code}'` returns 200 for these soft-404s"*).

## The fix is deliberately NOT the sibling fix

Same mechanism as #3277 / #3376 — `loading.tsx` wraps `page.tsx` in an automatic `<Suspense>`, the 200 shell flushes when the data fetch suspends, and `notFound()` arrives after the status is committed.

But #3277 **deliberately left this segment's `loading.tsx` alone**, and that was the right call: the reader is the one route where the skeleton earns its keep, because it shows on *every page turn*. Deleting it would trade a status code for a visible regression on the highest-traffic URL set in the site.

So instead the existence check moves **up into the segment's own `layout.tsx`**, which renders above the boundary `loading.tsx` creates.

The layout already had `getPageData()` for `generateMetadata`, and already **knew** about this case — it returns `title: 'Page Not Found - Source Library'` for it. It just never acted on it. Now it does, and the lookup is wrapped in `cache()` so the gate costs no extra query.

## Why this can't over-404

The layout's condition is a strict **subset** of `page.tsx`'s four `notFound()` cases:

| case | page.tsx | layout (this PR) |
|---|---|---|
| page id doesn't exist | 404 | 404 |
| book id/slug doesn't resolve | 404 | 404 |
| page.book_id ≠ book in URL | 404 | 404 (inside `getPageData`) |
| **book is hidden** | 404 (unless `allowHidden`) | **not checked — deliberate** |

The hidden-book gate stays in `page.tsx` alone. `findBookByIdOrSlug` does not filter on visibility, so the editor preview route (`page/[pageId]/preview`, which passes `allowHidden`) still renders hidden books through this same layout exactly as before. Gating on visibility here would have broken it.

No ISR concern: this route already reads `headers()` via `getTenantContext()` and is dynamic.

## Guard

The existing `tests/unit/soft-404-loading-guard.test.ts` gains a second `describe` block, because *"the loading.tsx is absent"* — the assertion shape used for every other route — **cannot express this fix**. The new block pins the opposite: the `loading.tsx` is still **present** (the skeleton is load-bearing), the layout calls `notFound()`, the lookup is `cache()`d, and the layout does **not** reference `isHiddenBook`.

722 unit tests pass; `npx tsc --noEmit` clean.

## Verification

Curl matrix against the preview once it builds — the three ids above must return **404**, and a real page (`/book/<slug>/page/<real-id>`) must still 200 and render with its skeleton intact. Status and body checked together, since status alone was 200 for all of these before.

### DIFF
--- a/src/app/book/[id]/page/[pageId]/layout.tsx
+++ b/src/app/book/[id]/page/[pageId]/layout.tsx
@@ -1,4 +1,6 @@
+import { cache } from 'react';
 import { Metadata } from 'next';
+import { notFound } from 'next/navigation';
 import { getReadDb } from '@/lib/mongodb';
 import { findBookByIdOrSlug } from '@/lib/book-lookup';
 import { getTenantContext } from '@/lib/tenant-context';
@@ -21,7 +23,9 @@ const PAGE_META_PROJECTION = {
   'translation.data': 1, 'ocr.data': 1, seo_indexable: 1,
 };
 
-async function getPageData(bookId: string, pageId: string, tenantId?: string): Promise<{ book: Book | null; page: Page | null }> {
+// `cache()`d because both generateMetadata and the layout shell below read it;
+// without it a reader page would issue this book+page lookup twice per request.
+const getPageData = cache(async function getPageData(bookId: string, pageId: string, tenantId?: string): Promise<{ book: Book | null; page: Page | null }> {
   try {
     const db = await getReadDb();
     const [bookResult, page] = await Promise.all([
@@ -44,7 +48,7 @@ async function getPageData(bookId: string, pageId: string, tenantId?: string): P
   } catch {
     return { book: null, page: null };
   }
-}
+});
 
 export async function generateMetadata({ params }: LayoutProps): Promise<Metadata> {
   const { id, pageId } = await params;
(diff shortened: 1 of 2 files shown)


JDerekLomas:
## Verified on the preview — and one case this does NOT fix

Preview: `sourcelibrary-v2-git-worktree-fix-re-d0f090…` (first Vercel build failed, retry passed — the known first-result behaviour).

**Unknown page ids now 404:**

| URL | before | after |
|---|---|---|
| `…/page/definitely-not-a-real-page-id` | 200 | **404** |
| `…/page/abc123def456abc123def456` | 200 | **404** |
| `…/page/zzzzzzzz` (other book) | 200 | **404** |
| `…/page/<real-id>x` (valid id, one char off) | 200 | **404** |

**Real content unaffected, skeleton intact:**

| URL | status | bytes | 404 digest |
|---|---|---|---|
| `/book/platonis-opera-plato/page/69b51e0fefd8df28f2dafa83` | 200 | 152,556 | 0 |
| `/book/platonis-opera-plato` | 200 | 364,987 | 0 |

**Editor preview route survives** — the specific regression risk of gating in a shared layout. A hidden book (`on-the-mysteries-ficino`) still resolves through `page/[pageId]/preview` rather than being 404'd by the new gate, confirming `findBookByIdOrSlug` really doesn't filter on visibility.

## The residual: hidden books still soft-404

Being explicit, because the PR description reads more complete than it is. A **hidden** book's reader URL still returns 200 with the not-found body:

```
preview build   /book/on-the-mysteries-ficino/page/<id>   200  digest=1
production      /book/on-the-mysteries-ficino/page/<id>   200  digest=1   ← identical
```

Production behaves the same, so this is **pre-existing and not a regression** — but it is also not fixed here, and it is the same class of defect.

It's excluded by construction: the hidden-book `notFound()` has to stay in `page.tsx`, because this layout also wraps `page/[pageId]/preview`, which exists precisely to render hidden books to editors. A layout gets no signal about which child segment is rendering, so it cannot apply a rule that must hold for one child and not the other. Fixing it properly means either giving `/preview` a sibling route outside this layout, or moving the visibility gate somewhere that can distinguish them — a restructure, not a one-liner, and out of scope here.

Practical impact is smaller than the unknown-id case: the body is a not-found page either way, so nothing leaks; it just isn't reportable by status code and stays indexable. Worth its own issue if hidden-book URLs turn up in `not_found_reports` gaps or in Search Console.

722 unit tests pass; `tsc --noEmit` clean; test ✅ DCO ✅.

PR #3388, opened 2026-07-28
fix(deploy): refuse to ship a stale main checkout + three verification lessons

JDerekLomas:
## The gap

`scripts/deploy-prod.sh` — which its own header calls *"the ONLY safe way to ship frontend"* — guards the branch **name** but never checks whether that branch is **current**:

```bash
BRANCH="$(git branch --show-current)"
if [ "$BRANCH" != "main" ]; then ... exit 1; fi
```

Being on `main` says nothing about whether `main` is up to date.

## How it surfaced

While verifying a fix on 2026-07-27 I found the shared main checkout **125 commits behind `origin/main`**. It passes the branch guard cleanly, and `vercel --prod` ships whatever is on disk — so a deploy from it would have silently rolled production back past months of merged work.

**Nothing in the output would have said so.** The build succeeds, the Cloudflare purge succeeds, and the script prints `✓ Production deploy complete (deployed → purged → warmed)`. The only way to notice is to already suspect it and run `git rev-list --count HEAD..origin/main` yourself.

This is a structural problem, not carelessness: that checkout is shared by ~10 concurrent sessions (per the Multi-Session Awareness doctrine), so no single session owns pulling it, and it drifts on its own. A convention can't fix a state that nobody is assigned to maintain — the script has to check.

## The guard

Added as step 1b, right after the branch check:

| condition | behaviour |
|---|---|
| **behind** `origin/main` | **refuse**, print the fix, `exit 1` |
| ahead of `origin/main` | warn, continue — `vercel` deploys the tree regardless, so the operator should know they're shipping unpushed work |
| `git fetch` fails | warn, continue — an offline emergency deploy is still legitimate; no network must not mean no deploy |
| `ALLOW_STALE=1` | skip entirely — deliberate rollback to an older commit |

The failure message tells you to run `npm install` after the pull, not just the pull: a merged `package.json` change leaves `node_modules` stale, which is its own known deploy-gate failure.

## Verification

Tested against **real git repositories**, not by reading the code — a scratch bare remote plus two clones, one deliberately 2 commits behind. The guard block was extracted verbatim from the script for the harness, so what was tested is what ships:

```
A  stale (2 behind)        → refuses, exit 1     ✓
B  current                 → proceeds, exit 0    ✓
C  stale + ALLOW_STALE=1   → proceeds, exit 0    ✓
D  ahead by 1 (unpushed)   → warns, proceeds     ✓
E  unreachable remote      → warns, proceeds     ✓
```

Case B is the one that matters most for safety-of-the-guard: a check that blocks legitimate deploys gets bypassed, and a habitually-bypassed guard is worse than none — the same failure the worktree reaper's noisy session count already taught us (`--force` became the habit).

`bash -n` clean.

## Out of scope

- The **EPIPE** path, where `vercel` errors after promoting and the wrapper continues to purge anyway. That already has explicit handling and a documented recovery; it hit again during this session and the recovery worked.
- Any change to what the deploy actually does — this only decides whether it may start.


JDerekLomas:
Added a second commit (7a2daf8e) with three CLAUDE.md invariants from the #3383 revert earlier in this session. Same session, same theme — how verification gave false confidence — so they ride along rather than fragmenting into another PR. Each is a mistake made today, not a hypothetical:

1. **The Vercel-preview trap has a second form.** CLAUDE.md already warns that curling a preview with `Host: bph.sourcelibrary.org` silently resolves to production. The form that actually bit was simpler — just *visiting the preview host*. A preview URL is `sourcelibrary-v2-git-*.vercel.app`, not a `.sourcelibrary.org` subdomain, so every host-gated path (`isTenantSubdomain`, `useEmbedContext`) correctly evaluates false and the tenant branch never executes. **The check cannot fail.** #3383 was "verified" that way at the path-based `/embed/bph`, where the buggy guard happened to work, and shipped a double-mounted menu onto EFM's public landing page.

2. **Never branch on `usePathname()` in a component that can render on a tenant host.** The proxy rewrites `/` → `/embed/<tenant>` internally, but `usePathname()` returns the browser path. So `startsWith('/embed/')` fires on the apex, where it isn't needed, and is inert on the subdomain, where it is. That inversion *is* the #3383 bug.

3. **A test that greps source is not a guard.** The unit test shipped with #3383 asserted seven "this string appears in this file" facts — including one pinning the exact check that was wrong — and passed green the whole time the feature was broken in production.

   The note is careful not to overcorrect: shape assertions are legitimate for **absence** invariants, where deletion *is* the failure mode — `tests/unit/soft-404-loading-guard.test.ts` is the good example, and re-adding one of those `loading.tsx` files genuinely reintroduces the bug. The rule is stated as a question to ask before writing a guard: *what code change would make this fail?* If the only answer is "deleting this line", it's documentation with a green checkmark.

No code in this commit — CLAUDE.md only.

Issue #3409, opened 2026-07-28
Browser-translation DOM crash is still live on collection pages + /search — 11% of Spanish-locale visitors


JDerekLomas:
PostHog captured **15,611 `$exception` events from 1,502 distinct users** in the 30 days to 2026-07-28. Both `$exception_type` and `$exception_message` are **null on every single one**.

So 1,502 people hit an error and we have no idea what it was.

## Why this is more than a nuisance

CLAUDE.md already records the cost of blind error reporting: the browser-translation reader crash (#3314) "ran for months unmeasured" because route-level `error.tsx` bypasses the global `ErrorReporter` boundary. That class of bug — real, user-facing, reported by users in another language before instrumentation caught it — is exactly what these 15,611 empty events would be hiding.

1,502 users is roughly 8.5% of the 30-day fingerprint count. That is not a rounding error.

## Proposed work

- Find why the exception properties arrive empty. Likely candidates: exceptions captured via a handler that doesn't pass the error object through, or `Error` instances being serialised to `{}` before capture.
- Cross-check against the `application_errors` collection — do these correlate with anything already recorded server-side, or are they purely client-side and otherwise invisible?
- Once populated, check specifically for the route-level `error.tsx` gap described in CLAUDE.md's browser-translation section; any route error page that doesn't call `reportError` itself is invisible by construction.

## Note

Related to but distinct from #3400 (PostHog counting bots as users). Fixing the bot contamination may also reduce this count — a headless fleet will throw exceptions of its own — but the null properties are a separate defect and would remain.

JDerekLomas:
**Correction: the properties were never empty.** This issue's premise is wrong, and the fix is to the query, not the code.

`posthog-js` (we're on ^1.398.0) sends exceptions as `$exception_list`, from which ingestion derives the **plural** `$exception_types` / `$exception_values`. The singular `$exception_type` / `$exception_message` I queried are the legacy pre-`$exception_list` names — they are null on every modern event by design, for every PostHog customer, and their being null says nothing about our capture.

Verified against the API: the same 30-day window returns fully populated messages. Nothing in `ErrorReporter.tsx` needs changing — the single-funnel capture is working as designed.

## What the data actually says

Top client errors, 30 days to 2026-07-28 (`count / distinct person_id`):

| events | users | source | message |
|---:|---:|---|---|
| 6,916 | 83 | window.onerror | `Script error.` (cross-origin, no detail) |
[...]

PR #3418, opened 2026-07-28
fix: stop the browser-translation crash on collection pages and /search (#3409)

JDerekLomas:
Closes the collection-page and `/search` half of #3409.

## What was actually wrong

#3314 fixed this crash class for the reader with **two** parts:

1. `TRANSLATION_DOM_GUARD_SCRIPT` — an inline `<head>` script that stops the *throw*.
2. A **remount** while a translator is active — which is what makes React's updates actually *arrive*.

Only part 1 was global. Part 2 was wired into `TranslationEditor` alone. Every other list that re-renders has been running with the un-fixed half ever since.

## The measurement

Crashes (`Cannot read properties of null (reading 'parentNode')` + `insertBefore`) vs pageviews, by path shape, PostHog 30d:

| surface | crashes | pageviews | rate | users |
|---|---|---|---|---|
| **collection** | 2,575 | 2,693 | **~0.96 / pageview** | 46 |
| **search** | 147 | 447 | **0.33** | 41 |
| reader-page | 333 | 502,053 | 0.0007 | 305 |
| book-landing | 1 | 44,084 | ~0 | 1 |

The reader — the one surface with the remount — is two to three orders of magnitude cleaner. Note the collection users: **46 people generating 2,575 events**, ~56 each. That is a page crashing repeatedly for the same visitor, not an occasional blip.

**Grouping matters here and nearly fooled me.** Reader pathnames are unique per page (`/book/<slug>/page/<id>`), so grouping by exact `$pathname` scatters reader traffic across thousands of rows and buries it, while `/collections/psychology` aggregates into one. My first pass did exactly that. The table above groups by path *shape*.

## Locale confirms the cause

Affected users as a share of that locale's audience:

| locale | affected | total (7d) | rate |
|---|---|---|---|
| **es-ES** | 28 | 245 | **11.4%** |
| es-419 | 11 | 176 | 6.3% |
| en-US | 48 | 6,675 | 0.7% |

Spanish-locale visitors are **~16× more likely** to hit it. That is the auto-translate signature, and it is how #3314 was originally found — via an Italian reader whose screenshot showed the Chrome translate icon lit.

(zh-CN is excluded from that read: its denominator is the Tencent headless fleet, not people.)

## The change

`TranslationSafe` — remounts its subtree instead of reconciling it, while `useBrowserTranslation` reports an active translator. The key is `undefined` otherwise, so **untranslated readers are byte-for-byte unaffected**.

Applied to `CollectionAllBooks`'s results region and `/search`'s results `<main>`.

**Both wrappers sit below the controls, never around them.** Sort, filter, query and pagination state live above the key and must survive a remount — that is precisely the mistake the reader's own code comment warns about ("Never key the whole reader… panel toggles/font size live above the key").

## What I am NOT claiming

**I could not pinpoint the throwing frame.** PostHog stacks are minified with no source maps — every frame is `?`. I checked and ruled out three candidate call sites in our own code:

- `useSearchHighlight` — reads `el.parentNode`, but null-guarded and only runs with `?highlight=`;
- `ScrollReveal` — walks `parentElement`, null-guarded, and its MutationObserver watches `childList` only and is rAF-debounced, so it cannot self-trigger;
- the PostHog loader snippet — reads `.parentNode` but on a script element that always has one.

So this ships as **the remedy already proven for this crash class on this codebase, applied to the surfaces that measurably still have it** — not as a fix to an identified line. That distinction is real and I would rather state it than imply more certainty than I have.

**Recommended follow-up:** enable `productionBrowserSourceMaps` so the next occurrence names its function. That is what would close this conclusively, and it is a one-line config change I deliberately kept out of this PR (different concern, and it changes build output).

## Guard, and its honest limits

`tests/unit/translation-safe.test.ts` pins that the wrapper is **transparent** — no extra element, markup byte-identical to rendering the children directly, identity never leaking into the DOM. That matters because both call sites sit inside CSS grid/flex containers: an added `<div>` would break both layouts silently while passing typecheck. **Negative control run — `Fragment` → `div` fails 2 of 3.**

The **remount half is not unit-tested**. It needs a DOM, an effect pass and a simulated translator; the repo's vitest environment is `node` with no jsdom or testing-library, and adding a DOM test runner is a tooling change that does not belong in a bug fix (per the PR conventions on tooling additions).

**So verify by measurement after deploy:** the es-ES crash rate should fall toward the en-US rate. Watch the *locale split*, not the raw count — raw counts are contaminated by the bot fleet. If es-ES stays near 11%, this was the wrong remedy and the source maps become the priority.

### DIFF
--- a/src/app/search/page.tsx
+++ b/src/app/search/page.tsx
@@ -34,6 +34,7 @@ import { LIBRARY_PARTNERS } from '@/lib/library-partners';
 import CollectionBookCard, { type CollectionBook } from '@/components/CollectionBookCard';
 import { getBookThumbnailUrl } from '@/lib/utils';
 import { getEffectiveByline } from '@/lib/byline';
+import TranslationSafe from '@/components/layout/TranslationSafe';
 
 // How many results to show in unified view per section
 const PREVIEW_BOOKS = 5;
@@ -1181,6 +1182,12 @@ export default function SearchPage({ defaultLibrary, forceEmbedded = false }: { 
 
       {/* Results */}
       <main className="max-w-[var(--container-wide)] mx-auto px-6 md:px-12 py-8">
+        {/* With a browser translator active React cannot reconcile these
+            result lists — the translator has replaced every text node — so we
+            remount them instead. Wraps only the results, never the search box
+            or filter state, which live above and must survive. See
+            TranslationSafe and #3314. */}
+        <TranslationSafe identity={`${query}|${viewMode}|${language}|${category}|${offset}|${resultsPerPage}`}>
         {/* Anonymous free-search wall — shown after 5 searches/hour */}
         {signInRequired && (
           <div className="text-center py-16 max-w-lg mx-auto">
@@ -1785,6 +1792,7 @@ export default function SearchPage({ defaultLibrary, forceEmbedded = false }: { 
             </div>
           </section>
         )}
+      </TranslationSafe>
       </main>
 
     </div>
--- a/src/components/collections/CollectionAllBooks.tsx
+++ b/src/components/collections/CollectionAllBooks.tsx
@@ -8,6 +8,7 @@ import CollectionListView from '@/components/collections/CollectionListView';
 import CatalogPagination from '@/components/collections/CatalogPagination';
 import { bookTitle } from '@/lib/collections-utils';
 import { getBookThumbnailUrl } from '@/lib/utils';
+import TranslationSafe from '@/components/layout/TranslationSafe';
 
 const PER_PAGE = 60;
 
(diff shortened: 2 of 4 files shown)


JDerekLomas:
Closing in favour of the real fix — see the PR that supersedes this.

I opened this before finding the cause, and it does not address it. All 120 sampled stacks for this crash carry **`$RS`** — React's inline streaming-SSR segment-replacement helper — with `filename` pointing at the HTML document rather than a JS bundle. The throw is `document.getElementById(...)` returning null (the translator removed the Suspense placeholder mid-stream) and then `.parentNode` being read off it.

`TranslationSafe` remounts React subtrees, which addresses client-side *reconciliation* — the second-order staleness problem from #3314. Real, but a different failure, and it happens after the point where this crash occurs. Merging it would be churn on two high-traffic files with no evidence it helps.

The component is easy to resurrect if staleness is ever measured on collection pages or /search; nothing here is lost.

Issue #3373, opened 2026-07-27
Deploy fragility: /explore prerender counts 1M entities at build time and hard-fails the whole build under write contention


JDerekLomas:
Found while deploying #3371/#3372. A production build failed outright:

```
Sitemap book-count failed: MongoServerError: operation exceeded time limit
[withTimeout] query failed: Executor error during getMore :: operation exceeded time limit
Error occurred prerendering page "/explore"
Export encountered an error on /explore/page: /explore, exiting the build.
⨯ Next.js build worker exited with code: 1
Error: Command "npm run build" exited with 1
```

## Mechanism

`src/app/explore/page.tsx` is ISR (`revalidate = 86400`), so it prerenders **at build time**, and its data fetch runs several `countDocuments` plus a `distinct('type')` against the **`entities`** collection — currently **1,017,740 documents** — each capped at `maxTimeMS: 25000`.

At that size the counts sit close enough to the cap that a concurrent writer on `entities` tips them over. In this case the writer was `scripts/maintenance/repair-entity-page-attribution.mjs` (#3361), which bulk-writes to `entities` continuously.

Evidence it's contention on that specific collection rather than ambient load:
- an earlier deploy the same afternoon passed **with the sweep running** — per-book write volume varies from zero to ~900 entities, so the collision is intermittent
- the retry passed immediately once the sweep was paused
- the two failing queries were both count-style reads (`/explore`, sitemap book-count); nothing else in the build failed

## Why it takes the whole deploy down

This is the prescribed behaviour from the CLAUDE.md invariant — a static page with a fallible fetch should **throw** rather than catch-and-cache an empty render (that's how `/explore/timeline` froze in #2973). Throwing protects the ISR cache. The side effect is that a marginal query fails the entire production build, and the failure mode is indistinguishable from a code error until you read the log.

So this isn't "the invariant is wrong" — it's that the page pays the invariant's cost with a query that is too close to its own timeout.

## Options

1. **Precompute the counts.** `system_config.homepage_stats` already exists for exactly this reason (refreshed nightly by `prewarm-browse.mjs`). `/explore` could read a stored `explore_stats` doc instead of counting 1M documents mid-build. Removes the fragility rather than widening the window.
2. **Raise `maxTimeMS`** on those counts. Cheapest, but it just moves the threshold and lengthens builds.
[...]

PR #3420, opened 2026-07-28
docs: correct the pre-deploy sweep guidance — /explore is no longer the risk

JDerekLomas:
Corrects a `CLAUDE.md` bullet that is now false. Docs only.

## The problem

The pre-deploy guidance said: pause `entities` sweeps, because `/explore` counts ~1M entities at build time and a concurrent writer tips it past `maxTimeMS`. Someone following that today would go optimize a query that **is not run**.

Measured against production 2026-07-29:

`/explore` moved its entity rollups to `system_config.entity_stats` in **#2996 (2026-07-05)**, behind `CANONICAL_ENTITIES_READPATH`, which is live in Production. Verified behaviourally rather than from config — curled the page and matched the discriminating counts:

| count | legacy (raw rows) | canonical | rendered on prod |
|---|---|---|---|
| `with_dates` | 7,948 | **6,291** | **6,291** |
| `with_coordinates` | 2,257 | **1,924** | **1,924** |
| `with_wikidata` | 27,854 | **20,859** | **20,859** |

Zero queries against `entities`. The one remaining build-time cursor runs at **909ms against a 15s cap**.

## What replaces it

The mechanism, not the symptom: bulk writes create **cluster-wide** resource pressure, and *any* marginal build-time cursor can lose to it. That's a property of the deploy, not of one page. Suspect whichever query sits closest to its own `maxTimeMS`, and read it out of the build log rather than from memory.

Two log-reading traps are now recorded, because **both misdirected the original #3373 diagnosis**:

- **`getMore` means a cursor** — an aggregation — never a `countDocuments` or `distinct`. The entity counts were the wrong suspect from the first line of the error.
- **`Sitemap book-count failed` is caught and non-fatal.** `generateSitemaps()` wraps it in `safeQuery` with a 10,000 fallback. It was quoted as evidence in #3373 and never failed anything.

Also widens the `pgrep` from `repair-entity` to `scripts/maintenance`, since the hazard is any bulk writer, not that one sweep.

## Out of scope

Precomputing the century heatmap was considered and **deliberately dropped** — see the closing comment on #3373. Its natural home (`homepage_stats`) uses a different filter regime than the heatmap needs, and `CLAUDE.md` instructs keeping those writers' filters in sync, so parking it there invites a later "harmonization" that silently changes the heatmap. Not worth new infrastructure for a query at 6% of its timeout.

Closes the doc half of #3373 (issue now closed).


JDerekLomas:
**Closing unmerged — the central claim is withdrawn.**

This PR rewrote the pre-deploy guidance on the premise that `/explore` no longer counts `entities` at build time. I cannot support that premise: `vercel env pull --environment=production` returns `CANONICAL_ENTITIES_READPATH=""`, which makes the flag helper return **false** and selects the legacy `entities`-counting path — directly contradicting the behavioural evidence I based this on. Full detail in the retraction on #3373, which is reopened.

Shipping a `CLAUDE.md` change that tells the next person "don't look at `/explore`" would be worse than the stale bullet it replaces, because it would actively steer them away from the right suspect.

**Two parts of this diff are still correct and worth redoing separately**, since both were read from the code rather than inferred from the flag:

- **`getMore` in a build error means a cursor** (an aggregation), never a `countDocuments` or `distinct`.
- **`Sitemap book-count failed` is caught and non-fatal** — `generateSitemaps()` wraps it in `safeQuery` with a 10,000 fallback, so it can never be the line that failed a build.

Those belong in `CLAUDE.md` as log-reading guidance regardless of how the flag question resolves. Deliberately not re-landing them here, so the correct half isn't carrying the withdrawn half.

PR #3445, opened 2026-07-29
Credit the holding library, not just the digital host

JDerekLomas:
## The problem

`https://sourcelibrary.org/book/opera-chirurgica-aquapendente` credited **"From the collection of Internet Archive."** But the 1628 folio belongs to **Fisher — University of Toronto**; IA only scanned it and hosts the scans. Scholarly convention — and IA's own item page — credits the custodian first, so this is an attribution error, not just a wording preference.

Pulling that thread found the same mistake at four layers: the book page, the `/libraries` directory, the Supabase catalog column feeding it, and the stored data itself.

---

## 1. Book page — credit the custodian, not the host

`src/lib/holding-library.ts` resolves two credits and **fails closed**: an unrecognised custodian yields the old single-line rendering rather than an invented attribution.

- **Only aggregators** (IA, e-rara, e-codices, Europeana, HathiTrust, Google Books) can name a second institution. The other ~39 partners digitize their own holdings, so their rendering is untouched.
- **Placeholders rejected** — 2,315 books literally store `"Internet Archive"`, 797 `"IIIF Source"`, 137 `"unknown library"`. Promoting any would state a falsehood *more* confidently than the old wording.
- **Non-libraries rejected**, found by reading all 209 names the resolver would otherwise credit: imprint statements (`Sumptibus Autoris`), BnF agent records (`Savage, John (actif 1690-1707). Graveur`), museum acquisition lines (`Rogers Fund, 1925` — 119 books), and text-dump sites (Library Genesis, Project Gutenberg).

**5,664 of 9,142 visible aggregator books (62%) now credit a holding library**, across 193 institutions.

Surfaces: collection section (custodian in the headline, `Digitized and hosted by X` beneath, still linking to `/libraries/X` which owns the stats and covers), pages subtitle, provenance timeline, both bibliographic panels (`Provided by` → `Held by`), and schema.org `creditText`. Two call sites were also reading `contributing_library` as the *digitizer*, which the field's own doc comment says it is not.

## 2. `/libraries` — group by institution, not by raw string

The directory's largest "contributing institution" was **Internet Archive**, the host. 302 entries, screened only by a four-item exclude list; Bayerische Staatsbibliothek was split across four spellings. Now **203 entries, all real**, with BSB's 3,042 books unified.

Merging is a **hand-verified alias map, deliberately not a heuristic** — both obvious similarity signals are actively wrong on this data:

- Substring containment pairs `"British Library"` with `"University of British Columbia Library"`.
- ASCII-folding collapses every CJK name to the empty string, merging 北京大學圖書館 (Peking University) with 浙江大学图书馆 (Zhejiang University).

Branch libraries stay distinct from their parent — `Fisher - University of Toronto` vs `Robarts - University of Toronto` is exactly the precision this credit exists to give. Left unmerged on purpose: `"Oxford University"` (83 books) is a generic contributor string that could name any Oxford library.

## 3. The catalog column was shipping the wrong institution

`/libraries` reads Supabase, and the column itself held the wrong value. **Two** fields carry a contributing library: the legacy top-level `books.contributing_library` (coarse, usually the provider) and `image_source.contributing_library` (the actual custodian). The sync preferred the top-level one, so **6,931 visible books shipped "Internet Archive" while the record knew "Cornell University Library"**.

Measured over all 19,420 visible books: **8,353 improved, 3,259 correctly cleared to null** (every cleared value verified a non-library), 7,808 unchanged.

This also explained the mystery `"[object Object]"` in the page's old exclude list: some records store an *object* here, which the old mapping passed through for Postgres to stringify. The resolver's `typeof` guard nulls it at the boundary.

## 4. Re-harvest — including values that are confidently wrong

`scripts/maintenance/harvest-holding-libraries.mjs`, read-only by default, two modes:

- **fill** — books with no usable custodian, from `archive.org/metadata`.
- **`--verify`** — books that *have* one, checked against IA and corrected.

`--verify` exists because **a stored custodian can be confidently wrong**. Checking 150 books that already had one against IA gave 91.3% agreement; the disagreements were structured. A batch import had blanket-assigned **`"Biblioteca Nazionale Centrale di Firenze"` to 955 books, and a 120-book random sample found 27.7% are held elsewhere** (National Central Library of Rome ×17, John Carter Brown ×7, Boston Public, Princeton Theological). Extrapolated: **~265 books would have rendered a confident misattribution** — precisely the failure this credit exists to prevent. Overwrites stamp `holding_library_harvest.previous`, so they are reversible.

That accuracy check also caught a bug in the harvester itself: IA sometimes returns the contributor as an **HTML anchor** rather than a name (`<a href="…">Bibliothèque interuniversitaire de santé (Paris)</a>`), which would have written markup into a field we render as text. With the fix, a 120-book verify run agrees 120/120.

## Verification

Corpus-scale, not fixture-scale — which was load-bearing. The fixture-only placeholder regex passed its tests while still crediting `"Google Books (partner libraries)"`; the corpus scan caught it, and the fix then also caught `"Internet Archive (user-uploaded Arabic library bundle)"`, which no fixture would have anticipated.

**Random sampling caught three things nothing else could**, which is why it is part of the process rather than a one-off:

1. Three cross-language merges invisible to any string method — `"Bavarian State Library"` shares no tokens with `"Bayerische Staatsbibliothek"` (same for Florence, Heidelberg).
2. The HTML-anchor contributor bug.
3. The Florence mislabelling.

**44 unit tests**, exercising the resolver against real corpus values rather than asserting on source shape. Includes a positive control for the `--verify` comparison — it must both fire (Florence vs Rome) and stay quiet (spelling/exonym variants), or the sweep is either a no-op or destructive. Full-corpus parity between `src/lib/holding-library.ts` and its `.mjs` twin over every distinct custodian name (fixture: `tests/fixtures/holding-library-names.json`; regenerate the twin with `scripts/lib/regen-holding-twin.py`).

Both rendering branches confirmed on a deployed preview: Bodleian and MDZ unchanged single-line credits; e-rara correctly promotes Bibliothèque de Genève above the host; `/libraries` junk-free with BSB at 3,022.

## Deploy notes

- `/libraries` will not reflect any of this until **`sync-books-catalog.mjs --full`** runs — that page reads Supabase.
- The harvest sweeps have not been applied to production yet.

## Out of scope

- **`display_title` hallucinations.** This same book renders an H1 of *"The Remaining Works of Benedictus de Spinoza"* — a different book entirely — leaking into `og:title`, breadcrumb, and schema.org. The record's own April audit flagged it verbatim and nothing acted. Needs a corpus sweep, because divergent `display_title`s are mostly *legitimate* Latin→English renderings and cannot be separated by heuristic alone.
- **Google Books (226 books)** publishes no holding library; **e-codices (28)** carries no per-item custodian in our records. Out of scope rather than silently "attempted and failed".

https://claude.ai/code/session_011bwShqgCRy19HqPc9dNdLw


JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. The concept shipped repeatedly by other routes: `src/lib/holding-library.ts` with the same AGGREGATORS design (#4367, 2987f62f), the `/libraries` host-vs-holder split (#4585, bf3f7d45), the null-holder citation rule (#4375), and the `[object Object]` catalog repair (`repair-contributing-library-4589.mjs`, #4589) which did the corpus-wide rewrite this PR described. 3,316 dirty lines against six weeks of shipped work is a rewrite, not a rebase. Residue worth a small follow-up if wanted: the `--verify` IA re-harvest, and `sync-books-catalog.mjs:115` still prefers the top-level `contributing_library` over `image_source`.

PR #3449, opened 2026-07-29
feat(welcome): ask for a name on the welcome form

JDerekLomas:
Stacked on #3448 — **base is that branch, not `main`.** Rebase onto `main` once #3448 merges. Without that fix the welcome gate never fires and this form is unreachable.

## Why

`/welcome` is the only place Source Library ever asks for a name, and it never did. Google sign-ins arrive with one from the OAuth profile; magic-link sign-ins never do — **1,762 of 3,854 user documents (46%) have `users.name` null.** That is why the `/welcome` greeting falls back to a bare "Welcome." and why the Resend audience has no first names for nearly half the list.

## What

An optional name field above the existing questions. **The two free-text questions are unchanged, verbatim** — only a field was added.

Behaviour worth keeping (each is easy to "simplify" away later, which is what the tests are for):

- **A blank name never clears an existing one.** A Google user's name is prefilled; clearing the box is far more likely a stray keystroke than a request to become anonymous. Writing `''` would strip names from exactly the users who have one.
- Whitespace-normalised, capped at 100 chars.
- **Skip still stamps `welcomedAt` only** — no name, no profile, no volunteers row. If that regresses, the gate re-fires forever.
- The name flows into the `volunteers` outreach mirror, and backfills `firstName`/`lastName` on the **Resend contact**, which was created at sign-up when a magic-link user had no name to give it. Best-effort — the profile is already saved by then, so a mailing-list hiccup cannot fail the request.
- `RESEND_AUDIENCE_ID` moved out of the inline literal in `auth.ts` so both writers address the same audience.

## Testing

`tests/unit/welcome-name-capture.test.ts` — 7 cases against the **real route handler**, asserting on the update document it actually writes (mocking `getDb`/`auth`, following the pattern in `analytics-event-props.test.ts`). Covers the blank-name guard, skip, normalisation, the ObjectId key, the volunteers mirror, and the 401.

Verified as a negative control: dropping the `if (name)` guard fails the clearing test.

Full unit suite green (897 tests / 78 files). `npx tsc --noEmit` clean.

## Deliberately not done

- **No structured interest checkboxes.** Considered a chip grid over `LIBRARY_CATEGORIES` (29 subjects) or the 20 pinned collections, which would give queryable segments and let us recommend collections. Derek chose free text — the four answers captured so far are genuinely richer than a checkbox set would be. Worth revisiting once there is volume.
- **The welcome email is untouched.** It is a static HTML constant (`WELCOME_HTML`, `src/lib/auth.ts`) with no personalisation, and it fires at `createUser` — *before* a magic-link user has given a name, so it could not greet them by name anyway without being resequenced. Personalising it means rewriting authored copy, which is Derek's call, not a drive-by.
- No name editing on `/account`. Today the only way to change a name is to have never set one.

### DIFF
--- a/src/app/api/me/welcome/route.ts
+++ b/src/app/api/me/welcome/route.ts
@@ -1,13 +1,19 @@
 import { NextRequest, NextResponse } from 'next/server';
-import { ObjectId } from 'mongodb';
-import { auth } from '@/lib/auth';
+import { Resend } from 'resend';
+import { auth, RESEND_AUDIENCE_ID } from '@/lib/auth';
 import { getDb } from '@/lib/mongodb';
+import { toUserId } from '@/lib/user-id';
 
-// POST /api/me/welcome — captures the first-fill of the user profile (languages,
-// interests, how they'd like to help) and marks the welcome step done.
+// POST /api/me/welcome — captures the first-fill of the user profile (name, who
+// they are and what draws them here, how they'd like to help) and marks the
+// welcome step done.
 //
-// Body: { languages?: string[], interests?: string, help_description?: string, skip?: boolean }
+// Body: { name?: string, about_you?: string, help_description?: string, skip?: boolean }
 // Skipping still sets welcomedAt so the gate doesn't fire again.
+//
+// This is the ONLY place we ask for a name. Google sign-ins arrive with one from
+// the OAuth profile; magic-link sign-ins never do, which is why 46% of user docs
+// had users.name null when this was written.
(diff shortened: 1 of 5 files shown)


JDerekLomas:
Superseded by #3451 — GitHub auto-closed this when its base branch was deleted on the merge of #3448, and a closed PR whose base is gone cannot be reopened or retargeted. Same commit, rebased onto `main`.

Issue #3473, opened 2026-07-30
page_revisions mixes re-OCR with re-archiving — 40% of pairs read a different leaf


JDerekLomas:
## Finding

`page_revisions` is used as a double-OCR dataset (109,953 same-page pairs). `scripts/eval/calibration-scorecard.mjs` fits agreement→accuracy on 32 anchor pages and applies it corpus-wide for per-script × per-era accuracy bands. **That inference assumes both sides read the same image.**

Measured, randomized sample (`scripts/audit/revision-image-shift.mjs`, n=3,339 pairs carrying a printed `<page-num>` on both sides):

```
identical page number : 1997 (59.8%)
DIFFERENT             : 1342 (40.2%)   ← the two passes did not read the same leaf
```

Per book (366 books with ≥3 comparable pairs):

- **112 books** have >50% of their pairs shifting
- **88 of those are SYSTEMATIC** — one offset explains ≥80% of that book's shifts
- The dominant offset is overwhelmingly **exactly +1**

```
9cddaf05   18/18 shifted   offset +1 explains 18/18   [SYSTEMATIC]
8fd0ac3e   46/46 shifted   offset +1 explains 46/46   [SYSTEMATIC]
712502ea   26/26 shifted   offset +1 explains 26/26   [SYSTEMATIC]
974800b0   80/80 shifted   offset +1 explains 76/80   [SYSTEMATIC]
```

**The printed page number is the right instrument here** — the model transcribes it *from the page*, so it identifies which leaf was photographed independently of how well the text was read.

## Why ±1 matters

That is the leaf-offset signature this repo already knows: #3368 (IA keeps calibration/`Delete` leaves in the zip but strips them from access derivatives, so a leading Color Card puts every scan one leaf behind its text) and #3186/#3357 (an e-rara PDF cover sheet, same shape).

**Direction suggests these are largely the REPAIR, not the damage.** A book whose images sat one leaf behind would OCR a printed number one *lower* than truth; correcting the alignment raises it. `+1` = new number higher than old, which is what a fix looks like. **Not confirmed** — that needs archive history (`batch_jobs.page_sources`, `archived_photo` provenance), not OCR text.

## The consequence, which holds either way

**Disagreement in this corpus is substantially image churn, not reading difficulty.**

[...]

PR #3475, opened 2026-07-30
audit: what page_revisions actually records, and the tools that measured it

JDerekLomas:
Started as a rescue of three artifacts stranded in one worktree. Two of them (`revision-image-shift.mjs`, `disagreement-typology.mjs`) and the 2026-07-30 handoff have since landed on `main` by another route, so **the remaining content is the measurement work built on top of them** — plus one correction to a claim that is currently on main.

## What this adds

**The measurement loop** (`.claude/docs/ocr-quality-measurement-loop.md`) — a metric computable for free, a named defect, an intervention, and a re-measurement that can come back negative. Signal is repeat instability on the 63,572 true-repeat pairs already sitting in `page_revisions`. It measures **stability, not accuracy**, and the distinction is load-bearing: two passes reciting the same memorised text agree perfectly.

**A difficulty taxonomy** (`.claude/docs/ocr-difficulty-taxonomy.md`) plus the triage queue, disagreement classifier, and the blinded repeat-instability draw with its preregistration.

**A worked intervention, deliberately not applied.** The plate/text specification gap: the OCR prompt says both "capture ALL text" and "describe illustrations with `<image-desc>`", and never says which governs text printed *inside* a figure. Verified on the image (Schott *Mechanica* p513, agreement 0.004, neither pass wrong). Baseline is **0.34% of illustrated pages**, not the 30% an earlier message claimed — 30.4% was the share of unstable pages that *carry* an illustration, a co-occurrence read as an effect size. At 0.34% the honest recommendation is to bundle the prompt clause with the next revision rather than fund a run of its own.

## The correction to main

`scripts/audit/ocr-revision-provenance.mjs` (new, free, ~2 min). Main's handoff resolved the ±1 page slide as the #3357 repair and argued it partly from an **inverted timestamp** — live `ocr.updated_at` older than the revision that preceded it, called "impossible for a re-OCR."

That inference doesn't hold, but the interesting part is why it was never needed: **`page_revisions.source` labels the mechanism of every row.**

```
source                        share of 191,221   numbered pairs leaf-shifted
batch_api                              57.5%                 3.8%
shift-repair-erara-2026-07             29.5%                99.0%   (89.9% +1)
pipeline_preview                        6.8%                 0.8%
ai                                      4.5%                 0%
```

The ±1 population that #3473 characterised by sampling, page-number arithmetic, per-book offset signatures and finally two scans opened by hand is just the rows that say `shift-repair-erara-2026-07`. And because the label doesn't depend on the page having printed a number, it reaches the stratum the printed-number instrument cannot see: **21.5% of pairs printing no number are shift-repair rows** — roughly 13,000 pairs, classifiable for free.

On the timestamps specifically: `created_at` is a *snapshot* clock (it records when the row was written, so it is later than the text it snapshots — the live stamp is older on 84.4% of pairs and 90% of pairs whose model demonstrably changed, which means inversion against it is uninformative). But `original_date` is a genuine reading clock — present on 91.8% of rows, ordering correctly on 99.3% of proven re-OCRs. The commit history here contains an intermediate claim that *no* clock could order these rows; that was one field's failure generalised to the category, and it is retracted in `5fc1c787`.

Also corrects **"re-archiving"** in four places. The dominant ±1 slice is text moving between page docs with the image untouched — the opposite of re-archiving, and it sends a literal reader at the wrong subsystem.

- **The corpus is not indicted and no number is re-fit.** The 63,572 same-leaf true repeats are selected on printed page number, model and prompt — never on a timestamp. What changes is that the clock cannot *audit* them, so "63,572 pairs exist" must not harden into "63,572 independent readings."
- **No re-fit of `calibration-scorecard.mjs`.** Excluding page-number-disagreeing pairs is cheap, but it moves a published methodology number and belongs in its own PR with its own before/after.
- **The prompt clause is drafted, not applied.** A prompt edit changes every future OCR call; 0.34% doesn't justify shipping unmeasured.
- **No CLAUDE.md invariant.** The nearest candidate — "don't stamp a mechanism claim from a clock nothing maintains" — is one measurement old.
- **No change to the corpus selection filter.** `source` is the better exclusion (categorical, and complete where the printed number is silent), but swapping it in moves the published 63,572 and belongs with the scorecard re-fit.

## Follow-up worth filing

Swap the true-repeat selection in `revision-agreement-corpus.mjs` from "equal printed page numbers" to "`source` not in the repair set", keeping the page number as the independent check. It covers the unnumbered stratum the current filter drops entirely, so it should *grow* the usable corpus rather than shrink it — but it moves a published number, so it needs its own before/after.


JDerekLomas:
**Do not merge yet — one commit is unverified against data.**

`4a6878e7` changes the true-repeat selector to segment `page_revisions` by `source` (excluding bulk maintenance sweeps like `shift-repair-erara-2026-07`, 56,413 ocr + 55,272 translation rows, 99.0% leaf-shifted vs 3.8% for `batch_api`). The logic is unit-tested against all 12 production source values, but the **before/after on the true-repeat population has not been run** — Atlas was unreachable when it was written.

The published 63,572 figure is unchanged in the docs, so nothing is currently misstated. But this PR should not land until:

```
node scripts/eval/revision-agreement-corpus.mjs
node scripts/eval/revision-agreement-corpus.mjs --field=translation
```

The expectation is that the count **grows** (the `source` label classifies pairs that print no page number, which the current filter drops entirely) — if it shrinks instead, the change is wrong and needs re-examining before anything else.

Second run is the first time any agreement work has looked at the translation side at all; `--field` was hardcoded to `ocr`.

Removing the `blocked` label once both have run and the numbers are recorded.

PR #3500, opened 2026-08-01
fix(archive): clear transient page-level failure markers so stranded pages can retry

JDerekLomas:
## The trap

Every archiver selects work with `archived_photo` missing/null/empty. Writing a `failed:` **string** into that same field hides the page from every future run — the bad write erases its own repair path. This is the class CLAUDE.md documents under Data Protection ("a bad write can erase its own repair path").

**Nothing clears these today.** `archive-auto-unblock.mjs` unsets only book-level `archive_metadata.*`; it never touches the page field. A page marked during a transient blip stays unarchived permanently.

Measured 2026-08-01: **14,123 pages** carry a marker. 3,000-page sample:

| count | reason | verdict |
|---|---|---|
| 2805 | `HTTP 403` | provider block — **retryable** |
| 133 | `HTTP 404` | absent — permanent |
| 55 | `source-not-found` | absent — permanent |
| 3 | `VipsJpeg: premature end of JPEG` | truncated transfer — retryable |
| 2 | `terminated` | retryable |
| 1 | `timeout` | retryable |
| 1 | `This operation was aborted` | retryable |

93.5% are 403 — a provider refusing us, which is **not** a statement that the page is absent.

## Why classification gets its own module and tests

The stakes are asymmetric in both directions:

- Clearing a **permanent** marker re-queues work that fails again and re-marks — a loop against a provider that already said no.
- Leaving a **transient** marker strands a page that is still there, forever.
- An **unrecognised** reason is left alone and reported. Defaulting unknown→transient would silently mass-re-queue the next novel permanent failure the moment it appears.

`classifyArchiveFailure()` also lets permanent win on a tie, so `"timeout then HTTP 404"` is not retried on the timeout.

## What the script does not do

**It does not archive anything.** It only makes pages eligible again; the normal archivers pick them up on their next run. That separation keeps it cheap and idempotent, and means a mistake costs a re-queue rather than a bad write.

Safety: dry-run by default (`--apply` required); `--max-per-provider` throttles so a 93.5%-403 population doesn't stampede a source already refusing us; the update re-asserts the marker in its filter so a concurrent archiver's real URL can never be clobbered; `--book-id` / `--limit` for scoped runs.

## Sequencing

Pairs with #3493. Clearing markers **without** the stall-timeout fix would just re-fail the largest pages and re-mark them — so #3493 should land first.

## Testing

`npx vitest run tests/unit/archive-failure-class.test.ts` — 7 behavioural cases over real production marker strings, including the unknown-is-not-transient default and the permanent-wins-a-tie rule.

### DIFF
--- a/scripts/lib/archive-failure-class.mjs
+++ b/scripts/lib/archive-failure-class.mjs
@@ -1,0 +1,77 @@
+/**
+ * Classify a page-level `archived_photo: "failed:…"` marker.
+ *
+ * Why this matters more than it looks:
+ *
+ * Every archiver selects work with `archived_photo` missing/null/empty. Writing
+ * a `failed:` STRING into that same field therefore makes the page invisible to
+ * every future run — the write erases its own repair path (CLAUDE.md, Data
+ * Protection). Nothing in the codebase clears these: `archive-auto-unblock.mjs`
+ * only unsets book-level `archive_metadata.*`, never the page field. So a page
+ * marked on a transient blip stays unarchived forever.
+ *
+ * Measured 2026-08-01: 14,123 pages carry a marker; a 3,000-page sample was
+ * 93.5% `HTTP 403` (provider blocking — emphatically not a statement that the
+ * page is absent), with a small genuinely-transient tail.
+ *
+ * The split below is the whole point of the module: clearing a PERMANENT marker
+ * just re-queues work that will fail again and re-mark, burning provider
+ * goodwill on a loop. Clearing a TRANSIENT one recovers a page that is still
+ * there. When a reason is unrecognised we return UNKNOWN and callers must leave
+ * it alone — an unclassified marker is not evidence of recoverability.
+ */
+
+/** Retryable: the source blocked, timed out, or the transfer broke mid-flight. */
+const TRANSIENT = [
+  /HTTP 403/i,          // provider block / rate limit — not "page absent"
+  /HTTP 429/i,
+  /HTTP 5\d\d/,         // any 5xx
+  /timeout/i,
+  /terminated/i,
+  /aborted/i,           // incl. "fetch aborted: stalled …" from the stall timeout
+  /fetch failed/i,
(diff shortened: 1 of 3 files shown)


JDerekLomas:
## Update — two defects found by running it, plus an honest limitation

### 1. The dry-run report was blind on the one distinction that matters

The reporting key collapsed **every** digit, so `HTTP 403` and `HTTP 404` both rendered as `HTTP N` — a single row reading `2438  HTTP N`.

The classification was always correct; only the *report* was blind. That's arguably worse, because it looks like information: an operator reads that line to decide whether to run `--apply`, and it merged the retryable class with the permanent one. Status codes are now preserved, incidental counts still collapse.

Corrected output at `--limit=2500`:

```
  transient: 2324  permanent: 176  unknown: 0

  by reason:
      2317  HTTP 403          <- transient
       121  HTTP 404          <- permanent
        55  source-not-found  <- permanent
         3  VipsJpeg: premature end of JPEG image
         2  terminated
         1  timeout
         1  This operation was aborted
```

transient 2324 = 2317+3+2+1+1, permanent 176 = 121+55 — reconciles exactly.

### 2. Silent multi-minute scan looked like a hang

`archived_photo` is unindexed, so a full pass is a collection scan over ~18.9M pages. The script printed nothing until the end, and **two runs were killed on suspicion of hanging before emitting anything**. Now emits a heartbeat every 500 markers.

### Limitation, stated plainly

**The full-corpus dry run has still not completed inside one run window.** Validated end-to-end at `--limit=2500` only. CLAUDE.md requires a corpus sweep be validated at corpus scale, so this PR is *not* ready for an unbounded `--apply` yet.

Also: the provider mix in that bounded output (`internet_archive` 1432, `gallica` 888) is the first 2,500 markers in collection order — a **partial-scan artefact, not a corpus profile**. Don't plan capacity from it.

Suggested before `--apply` at scale: add an index on `archived_photo` (or a partial index on the `failed:` prefix), then run the full dry run and re-read the true mix.

Issue #3517, opened 2026-08-01
Work anchors still encode the bogus author names from #3434 (86 work_ids, 5 of them indexed)


JDerekLomas:
Follow-up to #3483 / #3434. The book bylines are fixed (#3512, #3516); the **work-identity keys minted from them are not.**

## What's wrong

`work_id` is minted as `local:{a:author_id | n:author-name-slug}:{uniform-title-slug}`, so a run that wrote a search query into `books.author` also **anchored the work cluster to that query**. Clearing the byline did not unwind it.

```
local:a:viridarium-chymicum:garden-illustrious-poets     <- a Pico della Mirandola poetry anthology
local:n:sympatheticum-theatrum:europe                    <- Ortelius' Theatrum Orbis Terrarum
local:a:confessio-fraternitatis:confessio-augustana      <- the Augsburg Confession
```

It reaches the public URL twice over — `work_slug` is `<bogus-token>-<english-title>`, and the book slug got the token welded onto a real title:

```
/work/chymicum-the-garden-of-illustrious-poets
/book/mirandulae-viridarium-illustrium-poetarum-cum-ipsorum-chymicum
/book/europae-sympatheticum
```

## Measured scope

Of the 786 records repaired in #3483:

| | count |
|---|---|
| `work_id` containing a token of the bogus author | 108 books / **86 distinct work_ids** |
| `work_slug` likewise (public `/work/` URL) | 108 |
| book `slug` likewise (public `/book/` URL) | 112 |

**Only 6 of the 86 are in the sitemap.** `src/app/sitemap.ts` `getWorks()` emits `/work/{work_id}` only for works with **≥2 visible editions**; the other 80 are singletons and were never submitted to crawlers.

And one of those 6 is a **false positive** — `turba-philosophorum` is the legitimate work id; it matched only because the bogus author string shares a word with the real work's name. **Do not touch it.**

The 5 genuinely wrong *and* indexed:

```
local:n:sympatheticum-theatrum:theater-world          2 editions  (Ortelius)
local:n:philosophorum-turba:chemical-theater          3 editions  (Theatrum Chemicum)
local:n:chemicum-theatrum:chemical-german-theater     2 editions  (Deutsches Theatrum Chemicum)
local:a:confessio-fraternitatis:confessio-augustana   2 editions  (Augsburg Confession)
local:n:magia-naturalis-porta:history-italy-literary  6 editions  (Storia letteraria d'Italia)
```

[...]

Issue #3483, opened 2026-07-31
is_person:false is read by nothing in src/ — 59 quarantined non-person author pages are live


JDerekLomas:
Found while fixing #3434 (PR #3482).

## The defect

`scripts/maintenance/quarantine-non-person-authors.mjs` (#2179) flags non-person entries in the `authors` collection as `is_person: false` — institutions, placeholders, printer/series labels, and work titles miscatalogued as authors. It has been run: **59 docs carry the flag** (50 before PR #3482, 9 added by it).

**No file in `src/` mentions `is_person`.** The flag is read only by other maintenance scripts. So every quarantined author page is still live:

```
/author/theatrum-chemicum       HTTP 200, 4 book anchors,  <title>Theatrum Chemicum — Source Library</title>
/author/splendor-solis          HTTP 200, 7 book anchors,  <title>Splendor Solis — Source Library</title>
/author/rosarium-philosophorum  HTTP 200, 3 book anchors,  <title>Rosarium Philosophorum — Source Library</title>
```

A reader following one of these is told a book title is a person. `Splendor Solis` is an illuminated alchemical manuscript; `Theatrum Chemicum` is Zetzner's compilation; `Corpus juris civilis` is a legal corpus; `British Museum` and `Phurdrup Gonpa` are institutions.

Confirmed by curl against production on 2026-07-31, and by `git grep -n "is_person" -- src` returning nothing.

## Why this is the same shape as the three-layer crawler gate

The quarantine pass, the read path, and the sitemap are independent layers, and CLAUDE.md already records the lesson for the crawler gate: *changing one layer alone is silently defeated by the others.* Here the write layer was built, applied, and left disconnected from the read layer. The script's own doc comment states its purpose as "so author surfaces exclude them" — that exclusion was never implemented.

It is also the "guard nobody reads" pattern: a flag that looks like enforcement and enforces nothing. The tell is that `quarantine_reason` is set on 23 docs and no user-facing surface can distinguish them from a real author.

## Scope

1. **Decide the behaviour per surface.** A quarantined doc is not the same as a missing one — the books are real and their listing may be useful. Options: 404 the author page; render it as a work/collection rather than a person; or keep the listing but drop the person framing (no portrait slot, no "Works by", no author schema.org markup). Whichever is chosen must be the same in `/author/[slug]`, author listings, search author lanes, and the sitemap.
[...]

Issue #3434, opened 2026-07-29
Work titles and mentioned names landing in books.author (10 catalogues misattributed to Khunrath)


JDerekLomas:
> **Correction (rewritten after review).** This issue originally claimed that badging book-fair and auction catalogues as first English translations "asserts nothing," on the grounds that they are commercial documents. **That was wrong and the claim is withdrawn.** A Frankfurt fair catalogue is a primary source for the history of the book trade; the *Catalogus Librorum a Commissione Aulica Prohibitorum* is a Habsburg censorship record. Both are untranslated Latin primary sources, and a complete English translation of one is a genuine first and a real contribution. **52 of the 72 flagged records are ≥90% translated and their badges are correct.** The reasoning error was judging by genre rather than evidence. The completeness defect that the original triage misattributed to genre is filed separately as a corpus-wide issue.

What survives is an attribution bug.

## Work titles and mentioned names are landing in `books.author`

Found while checking the Khunrath corpus for #3425.

**10 visible books are attributed to `author: Heinrich Khunrath` that he did not write.** Khunrath died in 1605. The records include:

| id | year | title |
|---|---|---|
| `69b51d37261c58d6366491e1` | 1610 | Catalogus novus nundinarum vernalium, Francofurti ad Moenam |
| `69b51d4ef23ccdcced30774b` | 1614 | Catalogus universalis pro nundinis Francofurtensibus autumnalibus |
| `69b51d52f23ccdcced3077a3` | 1614 | Catalogus universalis pro nundinis Francofurtensis autumnalis |
| `69b51dc9cf111105c4290430` | 1651 | Catalogus universalis pro Nundinis Francofurtensibus |
| `69b51df5261c58d63664d47a` | 1758 | Catalogus librorum per quinquennium a Commissione Aulica prohibitorum |
| `69b51df5cf111105c4294131` | 1762 | Catalogus Librorum A Commissione Aulica Prohibitorum |
| `69b51df5768235dc6598a621` | 1758 | Catalogus Librorum per Quinquenium à Commissione aulica prohibitorum |
| `69b51df6768235dc6598a6e2` | 1765 | Catalogus Librorum per Quinquenium à Commissione aulica prohibitorum |
| `69b51df7efd8df28f2dae884` | 1765 | Catalogus Librorum, A Commissione Aulica Prohibitorum |
[...]

PR #3518, opened 2026-08-01
docs: #3483 handoff, and fix the self-matching pre-deploy sweep check

JDerekLomas:
Two things, both documentation.

## 1. Handoff for #3483

`.claude/handoffs/2026-08-01-non-person-author-pages.md` — the flag that was written and never read.

Records the finding that redirected the whole fix: the obvious reading ("these are mislabelled listings, reframe them as work pages") is wrong, and the work layer already held the proof. The 4 books under "Theatrum Chemicum" carry **3 distinct `work_id`s**, correctly, because they're 3 unrelated works. There is no work to redirect them to.

Also closes the #3434 open question about the originating importer — by inference rather than by finding the script:

- the search-term class is **108 records, 106 BSB**, with **96 in the single hour of 2026-03-14 08h**
- `import_candidates` BSB harvesting begins **2026-03-15 12h**, a day later, and writes the catalogue's real author (`"Kircher, Athanasius"`)
- the bsb id of a contaminated book has **no candidate row at all**

So the run predates the enumerate→dedupe→import loop and was replaced ~28 hours later. The code was deleted, not hidden. Stated as inference, with the check that would confirm it (deployment logs for that hour).

## 2. CLAUDE.md: the pre-deploy sweep check matches itself

The documented check is `pgrep -af repair-entity` on both machines. **The pattern matches the shell running it.** Any `bash -c` wrapper or ssh command string contains the literal text, so pgrep reports its own invocation:

```
$ ssh root@… 'pgrep -af repair-entity'
3240086 bash -c pgrep -af repair-entity || echo "  none on hetzner"
```

That's the command, not a sweep. It returned a confident-looking hit on **both machines simultaneously** today with no sweep running anywhere. `[r]epair-entity` matches the sweep but not the question.

Worth fixing rather than remembering, because it errs in the expensive direction: a false "sweep is running" reads as a reason to **skip** the deploy. And a check that cries wolf is one people learn to bypass — the same failure mode CLAUDE.md already records for the `ps | grep -i claude` session count that reported 34 sessions on a machine running 3 and trained everyone to reach for `--force`.

## Related

- #3512, #3516 — the fixes this documents (both merged)
- #3517 — the work anchors, filed with the measured breakdown
- #3434 — the originating contamination


JDerekLomas:

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. The pgrep bracket fix is on main at `deploy-and-caching.md:118` (b3c4311a, re-homed by 7efa0f4b), the fixes the handoff documents merged (#3512, #3516) and #3483 is closed. The only unique content is a six-week-old retrospective handoff; handoffs now go to the private ops repo.

PR #3546, opened 2026-08-02
feat(hero): a short mosaic repairs itself in the background

laubaumau:
## The problem

A book's hero mosaic is a cached image. Its grid shape is fixed by how many tiles survived **fetching** at generation time, not by how many pages the book has:

```
rows = floor(surviving_tiles / cols)   // capped at 4
```

So a run where a few page-image fetches flaked leaves a permanent 10×2 on a book with hundreds of usable pages — and nothing retries it, because the stored version still matches `HERO_MOSAIC_VERSION` and the route short-circuits to the stored image.

**Measured on production 2026-08-02** — 400 random books with a cached mosaic and ≥40 pages, each decoded and its grid counted:

| Shape | Sample | Est. corpus |
|---|---|---|
| 10 cols, 2–3 rows | 52 (13%) | ~1,500 |
| Narrow **and** short | 29 (7%) | ~840 |
| 7–8 cols, 4 rows (aspect branch, working as designed) | 34 (8.5%) | ~980 |
| Full 10×4 | 271 (68%) | ~7,800 |

119 of 400 are short; **115 of those have 40+ unused non-junk pages**. Alciato's *Emblemata*: 10×2 from 182 pages. *Corpus Inscriptionum Latinarum*: 10×3 from 880. Re-running the tile pipeline against Alciato today yields a full 40 with nothing discarded — the material was always there, the cached image is a fossil of a worse moment.

## The approach

**Not a version bump.** That rebuilds all ~12,660 cached mosaics including the ~70% already correct.

Instead: a book whose stored grid is short serves its existing mosaic (instant, no flash), and the reader's browser fires **one** background rebuild after the image has painted. This reader waits on nothing; the next visitor gets the fuller grid.

- `?upgrade=1` on the mosaic route — rebuild, judge the result, record it. Refuses on maxed / already-full / cooldown before doing any work.
- The generator now records `hero_mosaic_tiles/cols/rows`, so a grid's shape stops being recoverable only by decoding pixels.
- For the 12,660 mosaics predating that, the first ping **measures** the stored image rather than rebuilding it (one fetch + decode). ~70% get recorded as full and are never rebuilt — this is what stops it becoming a corpus-wide regen.
- The R2 key carries the tile count. Overwriting a single key would leave the CDN serving the old image, defeating the point.
- After a rebuild that changed the shape, the route's cached 302 is purged. Most books reach their mosaic through that route and the redirect is edge-cached for a week.

## Runaway protection

This fires from every reader's browser, so the limits matter:

- **A decided no-op is cached for a week.** The book page usually reads the Supabase catalog mirror, which has no `hero_mosaic_*` columns, so it can't rule an upgrade out itself and offers one whenever it can't. Caching the "nothing to do" answer per book turns those into browser/CDN hits that never reach a function.
- **A 10-minute cooldown, claimed before the expensive work**, so concurrent readers of one book don't all rebuild it.
- **Two rebuilds that fail to improve** mark the book maxed and it is left alone for good.

## Verification

- `tests/unit/hero-mosaic-upgrade.test.ts` — 16 cases over the decision rules, including that a book which never improves converges to maxed rather than retrying forever.
- `tests/unit/hero-mosaic-measure.test.ts` — the grid reader against constructed grids of every shape the compositor produces, through AVIF encoding (the stored format, lossy enough that gap bands aren't exactly the background colour), and a dark-scan case that could otherwise read as row gaps.
- The measurer also agrees with three production mosaics of independently known shape: 10×2, 10×3, 10×4.
- 1454 unit tests pass; `tsc --noEmit` clean. The two eslint errors in `HeroVariants.tsx` are pre-existing (verified against the unmodified file) and untouched here.

## Cost

No AI in this path. Per rebuild: ~64 thumbnail fetches, ~10s compute, one ~120KB R2 write. Spread over real traffic rather than run as a batch, and only for books measured as short — the other ~9,200 keep their cached image untouched.

## Out of scope

- The ~980 books at 7–8 columns with 4 rows. That's the aspect branch choosing fewer columns for short/landscape scans, which is the designed behaviour; whether it should still target 10 columns is a design question, not a bug.
- Backfilling `hero_mosaic_tiles` for books nobody visits. They're measured lazily on first view; a batch pass would work but isn't needed.

### DIFF
--- a/src/app/api/books/[id]/hero-mosaic/route.ts
+++ b/src/app/api/books/[id]/hero-mosaic/route.ts
@@ -152,7 +152,7 @@ export async function GET(request: NextRequest, { params }: { params: Promise<{ 
         projection: {
           _id: 0, id: 1, hero_mosaic_url: 1, hero_mosaic_version: 1,
           hero_mosaic_tiles: 1, hero_mosaic_maxed: 1, hero_mosaic_upgrade_at: 1,
-          hero_mosaic_upgrade_attempts: 1,
+          hero_mosaic_upgrade_attempts: 1, hero_mosaic_source: 1,
         },
       },
     );
@@ -165,7 +165,12 @@ export async function GET(request: NextRequest, { params }: { params: Promise<{ 
       // Refuse fast and cheaply — this endpoint is fired by every visitor to an
       // eligible book, so the no-op path must cost one indexed read.
       if (book.hero_mosaic_maxed) return upgradeSkipped('maxed');
-      if (priorTiles !== null && priorTiles >= MOSAIC_FULL_TILES) return upgradeSkipped('already-full');
+      const storedSource = book.hero_mosaic_source as string | undefined;
+      // A plate-tiled mosaic is never "already full": the book should be
+      // showing its pages, and those are usually fetchable by now.
+      if (storedSource !== 'plates' && priorTiles !== null && priorTiles >= MOSAIC_FULL_TILES) {
+        return upgradeSkipped('already-full');
+      }
       if (!mosaicUpgradeCooledDown(book.hero_mosaic_upgrade_at as Date | null, Date.now())) {
         return upgradeSkipped('cooldown');
       }
@@ -187,7 +192,13 @@ export async function GET(request: NextRequest, { params }: { params: Promise<{ 
             { id: book.id },
             { $set: { hero_mosaic_tiles: measured } },
           ).catch(() => {});
-          if (measured >= MOSAIC_FULL_TILES) return upgradeSkipped('measured-full', measured);
+          // Only a page-tiled mosaic can be settled by its size alone. A stored
+          // mosaic with no recorded source predates the field; measuring can't
+          // tell pages from plates, so a full grid is accepted as done and the
+          // plate case is caught on the next build, which does record it.
+          if (measured >= MOSAIC_FULL_TILES && storedSource !== 'plates') {
+            return upgradeSkipped('measured-full', measured);
+          }
         }
       }
       // Fall through: rebuild below, then judge the result.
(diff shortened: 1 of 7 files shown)


JDerekLomas:
Review 2026-08-27 (blocked, two findings):
1. The field-sprawl PR lint (field-sprawl-watch, added 2026-08-18, after this PR's last push) fails on the seven unregistered books fields this route writes (hero_mosaic_tiles/cols/rows/maxed/upgrade_at/upgrade_attempts/source — route.ts ~181,193,394-401). Green checks here are stale. Register the fields in scripts/lib/books-known-fields.json (+ baseline) before merge, and confirm the prod $jsonSchema validator is warn-mode — every updateOne is .catch(()=>{}), so an enforcing validator would silently drop the cooldown claim and every visitor would rebuild.
2. Retry hole: a rebuild exiting at distinct.length < MIN_TILES records no attempt, so a book with unfetchable images re-runs the full ~64-fetch rebuild every 10 min forever, and cacheNegative() there nulls a previously working hero_mosaic_url.
R2 keys, embed-branch placement, and drift vs main all check out — this is one registration + one guard away from mergeable.

JDerekLomas:
Backlog sweep 2026-09-12: left **blocked** (label was already on it). Unlanded — no `hero-mosaic-upgrade.ts`, `HERO_MOSAIC_VERSION` still 23, and the prewarm script is a different mechanism. Reader-facing (book hero) and it fires ~64 fetches plus an R2 write from every reader's browser, so it stays Derek's call. Book page and HeroVariants have taken 30+ commits since; expect conflicts.

JDerekLomas:
Closing (backlog decision 2026-09-22): fires ~64 fetches plus an R2 write from every reader's browser, and the book page / HeroVariants have taken 30+ commits since. If the fixed-grid hero problem still matters, it wants a server-side prewarm, which is the mechanism main now has. Branch kept.

Issue #3368, opened 2026-07-27
Bulk JP2 archiver equates zip ordinal with IIIF leaf → images one page behind OCR (2 reader reports)


JDerekLomas:
## Summary

Reader-reported image/text misalignment on `/book/*/page/*`: the scan shown is **one leaf behind** the OCR/translation beside it. Confirmed on two books, both from user feedback. Root cause is an indexing assumption in the bulk JP2 archiver, **not** the e-rara defect fixed in #3357/#3359 — this is a second, independent path to the same symptom, and it predates that work by three months.

## Reader reports

- **Pseudodoxia Epidemica** (`69aec19cdd3c98abc0da86b6`) — 2026-07-26: *"I'm seeing chapter 4 on the image page and chapter 5 in the text"*, *"This page has errors — not the same as image"*
- **The Federalist** (`69aec4473b6ebce5e0ee5929`) — 2026-07-16: *"The page ocr and text of image do not appear to be aligned"* (still unresolved)

## Evidence

**Pseudodoxia, page 221.** Image = printed 200 (start of *Chap. IV, Of the Picture of the Serpent tempting Eve*). OCR = printed 201 (*Chap. V, Of the Picture of Adam and Eve with Navels*). Fetching `page.photo` (IA leaf `n220`) directly returns printed 201 — so **the OCR is correct and the image is shifted**. Book-wide, not a one-page glitch: page 100's image is printed 79 where OCR says 80.

Perceptual hash (same dHash as the #3359 guard), pages 100/150/221/300:

```
pg100: src↔arch[N]=22  src↔arch[N+1]=4
pg150: src↔arch[N]=26  src↔arch[N+1]=4
pg221: src↔arch[N]=27  src↔arch[N+1]=2
pg300: src↔arch[N]=21  src↔arch[N+1]=4
```

`photo_original[N] == archived[N+1]` — the guard's `shift+1` verdict exactly.

**Federalist, page 319.** Image = printed 72, ending on catchword *"without"*. OCR = printed 73, opening *"without recuring more or less…"*. Same signature (dHash 2–5 vs `arch[N+1]`, 27–33 vs `arch[N]`).

## Root cause

`scripts/workers/archive-bulk.mjs:302`

```js
const leafNum = parseInt(leafMatch[1]);   // from IA IIIF URL /page/nN
const srcFile = pageFiles[leafNum];       // index into sorted *_jp2.zip contents
```

[...]

Issue #3362, opened 2026-07-26
Batch OCR served wrong images → corrupted OCR/translations/embeddings across a book cohort (Apr 2026 batch_api)


JDerekLomas:
## Summary

Batch OCR (`source: batch_api`) transcribed the **wrong page images** for a cohort of books OCR'd around 2026-04, storing accurate OCR of an *unrelated* document (mostly a ProQuest EEB scan of Perotti's *Cornucopiae* Latin index, plus Dutch/Greek EEB pages). The corruption is **page-index-aligned and identical across unrelated books**, and it **propagated downstream into translations and almost certainly semantic-search embeddings**. Affected books are **visible in curated collections** right now, serving wrong text and wrong translations.

This is a **data-integrity bug with a code root cause** in the batch OCR image-resolution path — not a model/OCR-quality issue.

## Evidence (what's proven)

Anchor case: *Rasaratnasamuccaya* (Vāgbhaṭa), Sanskrit rasa-śāstra text, `book_id 69d004bee2fdebf52dc4c849`, IA `VagbhataRasaratnasamuccaya1927`, **visible**, in collections `alchemy / indic-traditions / indic-alchemy / south-asia`.

- **819 pages; ~343 OCR'd as Latin, plus Dutch/Greek runs.** Only the first ~23 pages are correct Sanskrit.
- **Two OCR paths ran, only one is broken:**
  - Preview OCR — `source: "ai"`, no batch, model `gemini-3-flash-preview`, **Apr 3** (import day) → pages 1–23 correct **Sanskrit** (used correct images).
  - Full batch OCR — `source: "batch_api"`, model `gemini-3.1-flash-lite-preview`, prompt `v5.2026-02`, **Apr 4** → body pages Latin/Dutch/Greek (wrong images).
- **The injected text is a real, accurate transcription of a different document, page-aligned and identical across books.** Page 31 of *Rasaratnasamuccaya* (Sanskrit), *Micrographia* (English, `micrographia00hook`), and *A History of Hindu Chemistry* (English) are **byte-identical** Latin index text (`gymnaſiarches.521.15…`), running alphabetically in sequence (p28 E/F, p30 F, p31 G, p32 H). This rules out **model recitation** (position-independent, book-specific) and **per-book job cross-wiring** (each job correctly targeted its own `book_id`/`page_ids`, `gemini_state: JOB_STATE_SUCCEEDED`).
[...]

PR #3559, opened 2026-08-02
A revision is not a second read: shared filter + the double-OCR page inventory (#3558)

### LINKED ISSUE #3558: Translation revisions: 41% of the corpus is displaced text, not a second translation (#3473 follow-up)
Follow-up to #3473, which recorded the real `page_revisions.source` enum and flagged this without chasing it: **the e-rara shift-repair sweep touched 55,272 TRANSLATION revisions and no translation-side work has accounted for them.**

Measured 2026-08-02. Nothing is currently wrong — every consumer of the revision corpus filters `field: 'ocr'` — so this is a trap laid for the first translation-agreement study, not a live defect.

## What the rows actually are

The #3357 e-rara repair did not re-translate anything. It **moved** existing `translation` subdocuments between pages, and `page_revisions` snapshotted the displaced text. So a "pair" is not two translations of one page. It is **one translation filed against two page ids**.

Hand-checked, three books, windows of ~100 pages:

| book | window result |
|---|---|
| [Kepler, *Epitome astronomiae copernicanae*](https://sourcelibrary.org/book/epitome-astronomiae-copernicanae-usitata-forma-quaestionum-kepler) (1618) | 100/101 pairs at offset −1, Jaccard **1.00** |
| [*Kabbala denudata*](https://sourcelibrary.org/book/kabbala-denudata-seu-doctrina-hebraeorum-transcendentalis-knorr-von-rosenroth-2) (1677) | 92/101 at offset −1 (rest too short to score) |
| [Artemidorus, *Oneirocritica*](https://sourcelibrary.org/book/artemidori-daldiani-amp-achmetis-sereimi-f-oneirocritica-ibn-sirin-3) (1603, Greek) | 96/98 at offset −1 |

Randomized across the affected corpus — 600 revisions, 150 of the 323 books:

- **571 (95.2%) offset −1**, 27 too short to score, 2 unmatched
- **zero same-page pairs**

Baseline for comparison, sampling `source: 'ai'` translation revisions (n=400): **81.5% are genuine same-page re-translations**, ~4% shifted (that residue is the older #3368/#3357-class damage, not this sweep). A clean lane exists.

## Why it would pass review

Nothing in the metadata distinguishes the two sides:

- same model (`gemini-3.1-flash-lite-preview`), same `prompt_version: v1`
- the revision's `original_date` is **identical** to the live translation's `updated_at` — the moved subdocument carried its own timestamp with it

Only reading the text against its neighbour reveals it. Verified the live data is correct, not the revision: on Kepler pp. 41–43 the current OCR prints page-nums 9, 10, 11 consecutively and each current translation matches its own page's OCR.

## Share of the corpus — worse than the OCR side

| field | total revisions | from the sweep | share |
|---|---:|---:|---:|
| translation | 134,544 | 55,272 | **41.1%** |
| ocr | 191,221 | 56,413 | 29.5% |

A translation-agreement corpus built without the filter would find 41% of its pairs in near-total disagreement and read it as translation quality.

## Done in the PR that closes this

- `scripts/lib/revision-pairs.mjs` — shared filter, both halves (source label + printed `<page-num>`), plus a per-book shift verdict for uniform slides through unnumbered leaves. Unit-tested with negative controls.
- `scripts/eval/double-ocr-pages.mjs --field=translation` — the same inventory for translation, so the question "which pages were translated twice" has an answer before someone needs it.
- Wired into `revision-agreement-corpus.mjs`, `hard-page-sample.mjs`, `disagreement-typology.mjs`. `calibration-scorecard.mjs` reads the corpus summary, so it inherits the fix.

## Left open deliberately

- **No repair of the rows themselves.** They are an accurate record of what the page held before the repair; the defect is in reading them as a double pass, and that is now filtered. Deleting them would destroy the only evidence of the pre-repair state.
- **The ~4% residual shift in the `ai` lane** is unexplained here. It is consistent with the #3368 leaf-offset class and is caught per-pair by the page-number check, but nobody has enumerated it.

JDerekLomas:
Closes #3558.

`page_revisions` is read as a double-OCR corpus — a stored revision is what some later pass overwrote, so (revision, current) looks like one page transcribed twice, and agreement between the two sides is used as a proxy for accuracy. Three populations inside it are not double reads at all, and each is indistinguishable from a real one **from inside the row**.

## What was wrong

| population | why it isn't a second read | how it hides |
|---|---|---|
| `shift-repair-erara-2026-07` (56,413 ocr rows, 323 books) | the #3357 repair **moved** existing subdocuments between pages; the revision is the neighbouring leaf's text | same model, same `prompt_version`, and `original_date` equals the live text's `updated_at` — the moved subdocument carried its own timestamp |
| unlabelled image swaps (#3368 leaf offset) | the scan slid under the text | no source label marks it at all |
| human edits | a real second look, not an OCR pass | looks like any other revision |

The filter for this has two halves and **they are invisible to each other**: the printed `<page-num>` catches an unlabelled swap; only the source label catches the repair sweep, whose relocated text frequently prints the same number as the page it landed on. `hard-page-sample.mjs` had the page-number half inline and not the source half. `revision-agreement-corpus.mjs` and `disagreement-typology.mjs` had neither.

## What this adds

**`scripts/lib/revision-pairs.mjs`** — both halves in one place, plus a per-**book** shift verdict. A uniform slide preserves the printed page-number sequence exactly, so unnumbered leaves in an affected book pass every per-pair check; only the book's own signature reaches them. An unrecognised `source` returns `'unknown'` and is refused rather than folded in, because a bulk writer that borrows a pipeline label is indistinguishable from real model output forever after.

**`scripts/eval/double-ocr-pages.mjs`** — the inventory, which is a different question from the metric: *which pages carry two or more independent model passes over the same leaf*. One row per page with the full pass chain (source, model, prompt version, date, printed page number), per-pair verdicts, and `double_ocr` / `independent_reads` / `leaf_evidence` / `book_shift` to filter on. `--field=translation` runs the same inventory over translation revisions. `--write-collection` materializes `double_ocr_pages` / `double_translation_pages` (upsert-only, never deletes) so other jobs can query it.

It reports `verified-same-leaf` and `unverified-leaf` **separately**. On a smoke run 84% of usable pairs were unverified — most pages print no page number — so the headline total is not a verified total and the report refuses to present it as one.

No agreement metric here on purpose. One implementation of that already exists in `revision-agreement-corpus.mjs`; duplicating it is how two numbers start disagreeing.

**Wiring:** `revision-agreement-corpus.mjs` gains a `not_comparable` eligibility class (checked first — a pair that isn't a double read has no meaningful body-word class), with the exclusion reasons reported in its summary rather than silently dropped. `hard-page-sample.mjs` and `disagreement-typology.mjs` route through the shared filter. `calibration-scorecard.mjs` reads the corpus summary, so it inherits the fix with no edit.

## The translation side

Flagged in 2b8bfa2a and not chased. Measured now: **55,272 sweep rows = 41.1% of all translation revisions**, worse than the ocr side's 29.5%. Sampled 600 rows across 150 of the 323 books — **95.2% match the neighbouring leaf at Jaccard 1.00** (byte-identical), and **zero** are same-page pairs. Hand-checked three books ([Kepler](https://sourcelibrary.org/book/epitome-astronomiae-copernicanae-usitata-forma-quaestionum-kepler), [Kabbala denudata](https://sourcelibrary.org/book/kabbala-denudata-seu-doctrina-hebraeorum-transcendentalis-knorr-von-rosenroth-2), [Artemidorus](https://sourcelibrary.org/book/artemidori-daldiani-amp-achmetis-sereimi-f-oneirocritica-ibn-sirin-3)), all uniform offset −1. Live data verified correct: on Kepler pp. 41–43 the current OCR prints 9, 10, 11 consecutively and each current translation matches its own page's OCR. The clean lane exists — `source: 'ai'` is 81.5% genuine same-page re-translations.

Nothing consumed translation pairs yet. The point is that the filter exists **before** the first study does.

## Verification

- `tests/unit/revision-pair-filter.test.ts` — 21 behavioural tests. **Each guard was deleted and the suite watched go red** (text-move branch → 1 failure, different-leaf branch → 1, the `shift-repair-*` prefix rule → 1, unknown-label refusal → 2). Per CLAUDE.md, a source-shape assertion would have passed with the filter gone; these do not.
- `npx tsc --noEmit` clean; full unit suite 1554 passing.
- Smoke-ran `double-ocr-pages.mjs` and `revision-agreement-corpus.mjs` end-to-end against production Mongo; the new exclusion tallies appear in both reports.

## Out of scope, deliberately

- **No repair of the sweep rows.** They are an accurate record of what each page held before the repair. The defect is reading them as a double pass, which is now filtered; deleting them would destroy the only evidence of the pre-repair state.
- **The ~4% residual shift in the `ai` lane** is unexplained. Consistent with the #3368 class, caught per-pair by the page-number check, but nobody has enumerated it.
- **`revision-agreement-corpus.mjs` applies per-pair checks only.** It streams by `page_id`, so a book's pairs are never all in hand; the per-book verdict needs the two-pass structure and lives in `double-ocr-pages.mjs`. Its report says so rather than implying its unverified pairs are as good as its verified ones.

---

## Final numbers (regenerated 2026-08-03, all filters)

**Five exclusions, not three.** Two were found by running the thing: an
unknown-source alarm surfaced blank-page markers and mechanical tag repairs
(`derived-text`), and chasing a Sanskrit anomaly surfaced `different-script`.

| | OCR | translation |
|---|---:|---:|
| pages carrying a stored revision | 164,664 | 133,287 |
| **pages with a genuine second pass** | **97,421** | **70,502** |
| usable pairs | 117,729 | 70,549 |
| verified same-leaf / unverified | 63,776 / 33,645 | 788 / 69,714 |

Excluded pairs, OCR: `text-move-source` 56,822 · `different-leaf` 5,330 ·
`different-script` 3,809 · `book-shifted` 1,822 · `human-edit` 21 · `derived-text` 1.

Agreement corpus: 108,468 eligible, median 98.3%, mean 86.6%. The
near-zero-overlap tail fell 2,810 → 2,020 (28.1%), independently reproducing the
27.6% cross-script share measured by hand before the filter existed.

## The cross-script finding

A Sanskrit stratum reading 0.2% median agreement turned out to be **#3362 visible
in the disagreement data**. On `rasaratnasamuccaya-vagbhata` p27 the stored OCR
was a Latin index (`erigeron`, `eryngium`) while the scan is a Devanagari table of
contents; the prior text's `original_date` is 2026-04-04, inside the window when
the archiver wrote pages to a shared `archived/undefined/<n>.jpg` key. The re-OCR
of the correct image is the repair.

Across the whole near-zero tail, **27.6% of pairs have their two sides in
different writing systems** (539 latin→devanagari). A quarter of the "hardest"
pages by agreement were not hard pages, and they concentrate in Sanskrit, Tibetan
and Arabic — so any stratified quality ranking that kept them would have put those
languages at the bottom of the library for a reason unrelated to reading them.

**The cross-script test is nearly inert on the translation side** (6 pairs): a
translation is in English whatever the source page is. Translation pairs rest
almost entirely on the source label and the per-book verdict, and that limit
should be stated wherever those numbers are used.

## Instruments that failed before one worked

Recorded because the pattern recurs: prior-side date inside the #3362 window
separates barely (84.8% vs a 78.0% baseline — most OCR happened then); a
`<language>` tag disagreement partly measures the PROMPT, whose tag vocabulary
changed between versions (`<lang>sa</lang>` → `<language>Sanskrit</language>`); and
a first cross-script measurement read 2.3% because it sampled the first 400 rows of
a file ordered by page id, which clusters by book and missed Sanskrit entirely. The
real figure across all 2,810 is 27.6%.


JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded by 813e4b3f (#3475, merged 2026-08-04, two days after this opened). It landed the shared revision-source filter (`scripts/eval/lib/revision-source.mjs` + test), the double-OCR inventory (`reocr-pairing-check.mjs` + results), `ocr-quality-measurement-loop.md`, and the thesis itself as `invariants/page-revisions-corpus.md` — and rewrote the same two eval files this PR edits.

PR #3562, opened 2026-08-02
fix(pages): repair leaves split in half that were never spreads

JDerekLomas:
## What

Reported by Corey (via WhatsApp, 2026-08-02): the frontispiece of Weigel's *Studium Universale* rendered and downloaded as only its left half.

BPH photographed that volume two ways: front matter one leaf at a time (portrait, ~2000×3080), the body as spreads (landscape, ~4110×3096). A splitter applied a uniform 50/50 cut to the whole book, so the individually-shot leaves got cut in half.

It is **not** "the book was split wrong" — 98 of its pages are genuine spreads, split correctly. Exactly 4 pages were broken (p1, p4, p15, p16), all front matter. That is why earlier false-split sweeps missed it: the book looks fine everywhere you would naturally check, and the damage lands on frontispieces and title pages — the pages people actually try to download.

## The discriminator

The source image's own **aspect ratio**, read from the bytes on R2 — not any stored flag. Every flag on these pages says "split me" (`crop` set, `cropped_photo` materialized, `needs_splitting: false`).

- landscape source (w > h) → real spread, the split is correct
- portrait source (w ≤ h) → single leaf, any half-width crop is wrong

Source *sharing* is **not** a usable signal, though it looks like one: the archiver stored a separate copy of each spread per output page, so both halves of a genuine spread also have unique `archived_photo` values. I checked p48–p52 of the Weigel book directly to rule this out before settling on aspect ratio.

## Nothing was destroyed

`archived_photo` still holds the complete leaf at full resolution, and the canonical `pages/{book}/{NNNN}.jpg` and `-thumb.jpg` variants are already full-page resizes of it. Only `crop`, `cropped_photo` and `thumbnail` point at the half. Clearing them restores the full page through the normal resolver in `src/lib/page-image-url.ts`. Pure metadata repair — no re-archiving, no Gemini cost.

**Naming trap:** the half is stored as `{NNNN}-full.jpg`. The file called `-full` *is* the cropped half; plain `{NNNN}.jpg` is the full page. Easy way to misdiagnose this at a glance.

## Status

- Weigel book **already repaired** (4 pages), Cloudflare purged, ISR revalidated. Verified against production HTML that the reader now references `0001.jpg` (full, 1200×1875) instead of `0001-full.jpg` (half, 969×3084) — not just verified as a DB write.
- Corpus scan of visible BPH books was **interrupted at 418/2,835** (author was on airplane wifi). At that point: 14 books affected, ~40 broken pages. Extrapolates to roughly 90–100 books. Needs to finish on Hetzner, where the bandwidth is.

## Out of scope, deliberately

**No guard in the splitter itself.** The durable fix is refusing to split a portrait source at write time (the "verify the pairing at write time and fail closed" rule in CLAUDE.md). I have not written it because I have not yet confirmed which splitter produced this data — it is from Feb–Mar 2026, and `src/app/api/pages/[id]/split/route.ts`, `batch-split`, and `auto-split-ml` are all candidates. Guessing wrong would add a guard to a path that never runs. Separate PR once the scan tells us the real scope and I have traced the writer.

## Safety

Dry-run by default. Writes a JSON backup to `scripts/output/` before any write; `--revert <backup.json>` restores exactly. `--book <slug|id>` for one book, `--all-bph` to scan the cohort.


JDerekLomas:
## ⚠️ Do not run `--apply` on the cohort — the detector has a false-positive class

The full BPH scan finished on Hetzner: 1,220 books with crop-bearing pages, reporting **1,046 falsely split pages across 55 books**. That number is **not trustworthy**, and I found out by checking the largest finding instead of believing it.

`the-divine-and-most-necessary-doctrine-of-the-triune-god-doehring` was flagged at **543 of 543 pages** — 100%, which is the shape that always deserves a second look. It is a false positive:

| page | source | verdict |
|---|---|---|
| p1 (uncropped) | 1415×3106 | genuine single leaf, aspect 0.455 |
| p2 (cropped) | 2780×3105 | aspect 0.895 — **two** 1415-wide leaves side by side |
| p276 (cropped) | 3050×3080 | aspect 0.99 — also a real spread |

The book has tall, narrow leaves. Two of them side by side is still **taller than it is wide**, so my `w > h` test calls the spread "portrait" and every correct split reads as broken.

## The actual flaw

`w > h` assumes a leaf is roughly 2:3. For narrow folios a genuine spread lands at aspect 0.9–1.0 and slips under the threshold. The absolute test cannot work, because single-leaf and spread aspect ratios from *different books* overlap.

Worth being precise about why this is hard: the crop always removes half the width, so `rendered ≈ 0.5 × source` in **both** the broken and the correct case. That comparison carries no signal at all. The only real question is whether the source image is one leaf or two, and answering it requires knowing what a leaf of *that book* looks like.

## Fix: calibrate per book

- Use the book's own **uncropped** pages as the single-leaf reference (Doehring p1/p548 give 0.455/0.474; the cropped sources sit at ~2× that → genuine spreads).
- When a book has no uncropped page (the Weigel case — all 114 carry a crop), use **bimodality** of source aspect ratios within the book. Weigel is cleanly bimodal: ~0.65 singles and ~1.33 spreads, a factor of 2.08. The narrow cluster is the falsely split one.
- Abstain and report "undetermined" when neither reference exists, rather than guessing. Following the CLAUDE.md rule that an unmeasurable is not a zero.

## What still stands

The **Weigel repair is correct and unaffected** — I verified it by eye, not by threshold: `archived/1.jpg` (1974×3084) is visibly one complete bordered engraving with the binding at its left edge, while `archived/50.jpg` (4110×3096) is visibly a spread. That book is cleanly bimodal at a 2.08 factor. The 4 repaired pages and the production verification hold.

Everything else in this PR is a scan tool that currently over-reports. **Not mergeable as-is** — I will push the per-book calibration and re-run before this goes anywhere near `--apply`.

JDerekLomas:
## Applied — 567 pages across 48 books, verified in production

Final scan and repair complete. The detector went through four revisions, every one of them forced by opening an image rather than by anything the classifier reported.

**Result:** 567 visibly broken pages repaired across 48 books. Cloudflare purged, all 48 books ISR-revalidated (14 needed a paced retry after hitting the bot limiter at 429).

**Verified in production, not just as a DB write.** Duret's 1619 title page now serves `0004.jpg` at 1200×1766 — the full leaf — where it previously served `0004-full.jpg` at 1000×2943, the right half. "THRESOR DE L'HISTOIRE DES LANGVES DE CEST VNIVERS" reads complete again. Fludd p2 likewise.

**Validation before applying:** 13 pages hand-checked in total — 8 targeted (including all 3 false-positive classes) plus a 5-page random sample drawn *by book*, since sampling by page would have put 4 of 5 in Duret alone. All 5 random picks were genuine single leaves: an interior text page, an engraved title plate, a bookplate pastedown, a marbled endpaper, an index page.

**A verification bug worth recording.** My first post-apply check reported "0 pages still carrying crop" — from a query that matched *nothing*. The backup serialises `_id` to a hex string while Mongo stores an ObjectId, so `{_id: {$in: [<strings>]}}` matched zero documents and the zero read as success. Re-run with correct casting: 567 matched, 567 cleared. A count of zero needs its denominator checked before it means anything.

**Revert path:** `scripts/output/false-split-repair-2026-08-03T21-31-14-551Z.json`, restorable via `--revert`. Pulled off Hetzner to a local copy, since `scripts/output/` is untracked and the box was otherwise the only copy.

## Still open — the important one first

**The splitter has no guard, and the damage is not only visual.** `src/app/api/books/[id]/auto-split-ml/route.ts` selects pages by `crop: { $exists: false }` and splits every one the ML model returns a position for, with no check that the source is a spread. Worse, the same update runs:

```js
$unset: { ocr: '', translation: '', summary: '' }
```

So every false split **destroyed that page's OCR and translation**. The 567 pages have their images back; their text is only recoverable by paid re-OCR. This also reframes the 7,792 metadata-only pages below — they kept their text only because their crops were never materialised.

The guard belongs immediately before `updateOps.push`, comparing the source aspect ratio against the book's own leaf ratio and refusing at ~1× rather than trusting model confidence. It should share the detector logic (TS twin, parity-pinned), not reimplement it. Separate PR, and it wants careful review: a guard that refuses too eagerly stalls legitimate splitting corpus-wide.

**349 books undetermined.** Of 1,220 BPH books carrying crop data, the test reaches a verdict on 871 and abstains on 349 — books where every page is cropped, so there is no uncropped reference and no bimodal split to infer one from. A book falsely split *in its entirety* lands there by construction. **567 is a floor, not a total.** A possible second signal (a rendered half at ~0.32 aspect is implausibly narrow for a real page) is untested, and the last three untested ideas here were all wrong.

**7,792 spurious crops that still render full** — invisible today, deliberately untouched, but live ammunition for any future pass through the unguarded splitter.

**189 ambiguous pages** abstained on (source between 1.25× and 1.75× the leaf ratio). Those want eyes, not a threshold.

JDerekLomas:
## Final state after human review

**470 pages repaired across 45 books**, plus the 4 Weigel pages from the first pass — 474 live repairs. **97 reverted.**

Human review of the full contact sheet is what closed this out, and it changed the number twice.

### Two failures caught after "verified in production"

**1. The repair was only half applied (196 pages).** Clearing `crop` + `cropped_photo` makes `resolveSized()` fall through to `display_photo` — which for 196 of 567 pages was *itself* a resize of the half. Those read as repaired in the database and changed nothing on screen. My earlier production check was true for the two books I sampled and false for a third of the rest. Fixed by `fix-half-display-variants.mjs`; re-verified at 0 remaining.

**2. 97 pages were genuine spreads I un-split (4 books).** They clustered at aspect 1.19–1.20, immediately under the 1.2 ceiling — a hard cutoff means anything at 1.199 sails through. Hoffmann 36/36, Tauler 29/29, Cabalistic Science 28/28, Eckartshausen 4 of 6. Reverted from backup, purged, revalidated; 97/97 confirmed restored.

Sorting the review sheet by aspect ratio, widest first, is what surfaced this in seconds. The classifier never would have — it had already judged every one of them correct.

### Scorecard for the detector

Five aspect-ratio heuristics were tried. **Four were wrong**, and every single failure was found by opening an image, never by a check that inspected the data:

| version | claimed | truth |
|---|---|---|
| `w > h` absolute | 1,046 | 543 false positives on one book (narrow leaves) |
| per-book calibration | 10,277 | over-counted spurious-but-unmaterialised crops |
| + both conditions | 567 | still un-split 97 real spreads |
| + dead-zone + ceiling | 470 | current |
| auto-classifier for the flagged set | — | discarded; contradicted direct observation |

### A verification bug worth copying into your own habits

My first post-apply check reported "0 pages still carrying crop" from a query that matched **nothing**. The backup serialises `_id` as a hex string; Mongo stores an ObjectId. `{_id: {$in: [<strings>]}}` matches zero documents, and zero reads as success. Caught only because the sample loop printed no rows. **A count of zero needs its denominator checked before it means anything.**

### Follow-ups filed

- **#3593** — the splitter guard. Still the important one: `auto-split-ml` runs `$unset: { ocr, translation, summary }`, so every false split destroyed that page's transcription. The 474 pages have their images back, not their text.
- **#3608** — a *different* defect the review surfaced: archived captures framed too far left, clipping text mid-word (Lauterbeck p943, p979). Lives in `archived_photo`, untouched by this repair, needs re-archiving. Scope unknown — only 4 of 39 flagged pages hand-verified, and deliberately no automated classifier.

### Still unjudged

**349 books** where every page is cropped, so there is no uncropped reference and no bimodal split to infer one from. A book falsely split in its entirety lands there by construction. 470 is a floor.

JDerekLomas:
## The 349 unjudged books: closed by human review

Derek reviewed the 54 pages that still carry a crop and whose OCR reported a sliver — the only remaining way a false split could hide in the 349 books with no internal reference. **Verdict: all fine.** They are volvelles, pasted slips, marbled endpapers and blank leaves, where the "narrow strip" is a real feature of the book rather than crop damage.

That closes the question this PR could not answer by measurement. The reasoning it rests on:

- 1,662 pages corpus-wide carry an OCR warning about being fed a narrow strip; only **54 still carry a crop**.
- Those 54 sit 1–3 per book across 41 books. A systemic split failure concentrates; this scatters.
- Every one was then looked at by a human.

**Caveat kept on the record:** "the OCR did not complain" is not proof of health. A model fed a clean half-page will often transcribe it without remark — which is exactly what the Weigel and Duret pages did. This detector finds damage; it does not certify absence. If another false-split report arrives from a reader, the 349 are still the least-examined part of the corpus.

## Final tally

- **474 pages repaired** across 45 books, live in production
- **97 reverted** — real spreads un-split by an over-eager detector, caught by human review
- **54 re-OCR'd** — 36 gained >15% text, 18 unchanged, 0 lost text ($0.12)
- **Guard shipped** (#3593) in report mode, with a negative control

Issue #3560, opened 2026-08-02
Volunteer participation: three tiers, one sampling frame — and why the review queues have zero rows


JDerekLomas:
## Why

117 people have told us, in their own words, why they came. A translator who works professionally between Turkish, Spanish and English. A scholar of Neoplatonism. A Greek astrologer specialising in Hellenistic and Byzantine sources. A medievalist now working in Special Collections. A published translator of esoterica from French, German and Latin who moderates an alchemy forum. A post-doc in early modern medicine. A doctoral student who needs reliable Rosicrucian manifestos. An independent researcher who says he has a gap list of 1776–1800 Illuminati literature ready to hand over. A French bibliophile offering books from his own collection and corrections to our bibliographic records.

Not one of them has ever been given anything to do.

This issue is about closing that gap, and about being able to say something *true and checkable* about the quality of what we publish.

## The finding that reframes the problem

We already built a volunteer review system. It is live in production right now:

- `/review` — hub, three queues
- `/review/hallucination` — page image beside the AI's description of it, ~10s per judgment
- `/review/gallery-quality` — is this illustration gallery-worthy
- `/review/scan-quality` — classify digitization quality
- `src/lib/review-queue.ts`, `src/components/review/useReviewQueue.ts` — shared rating hook, keyboard shortcuts, anonymous per-browser volunteer IDs
- `POST /api/review/submit` → Supabase `volunteer_ratings`

`volunteer_ratings` contains **zero rows**. Not few — none, ever.

`git grep` finds **no inbound links** to `/review` from the header, the footer, the homepage, or `/contribute`. `/contribute` links only to the signup *form* (`/contribute/volunteer`), which thanks the volunteer and sends them nowhere. `/volunteers` is a near-duplicate of `/review` whose own source comment describes it as "an evergreen direct-link for outreach emails."

So: 132 signups, 3 working queues, 0 connections between them.

**The bottleneck is not queue design, cadence, or gamification. Nobody was ever asked.** Any redesign that skips this fact will reproduce the same result more expensively.

## What the volunteers actually said

134 volunteer records; 117 wrote free text via the `/welcome` flow (`volunteers` collection: `help_description`, `about_you`, `preferred_language`).

| Interest | Count |
|---|---|
| "Just here to read" | 25 |
| Study groups | 22 |
| Reviewing translations | 17 |
| Suggesting books | 17 |
| Annotating texts | 11 |
[...]

Issue #3364, opened 2026-07-26
Tenant lockdown leak: /encyclopedia and /explore/* serve the full global page on tenant subdomains


JDerekLomas:
Found while checking whether a new query on `/encyclopedia/[name]` needed tenant scoping (#3361). **Pre-existing on production, not introduced by that work.**

## The leak

`bph.sourcelibrary.org` serves these routes as the **global** page — global site header and nav, global data, no tenant scoping:

| path on bph.sourcelibrary.org | HTTP | served title |
|---|---|---|
| `/encyclopedia` | 200 | Encyclopedia - Source Library |
| `/encyclopedia/Matthiolus` | 200 | Matthiolus (Person) - Source Library Encyclopedia |
| `/explore/timeline` | 200 | Timeline — Explore — Source Library |

`/encyclopedia/Matthiolus` on the BPH host renders the full global entry: the global nav, the global "Connections" block, and **"Appears in 117 Books"** — a list built across the entire library with no `held_by` / `image_source.provider` / `tenantId` constraint. Each book title links to `/book/<id>`, so a BPH visitor is one click from non-BPH holdings.

This violates two invariants in the CLAUDE.md **Tenant Subdomain Lockdown — CRITICAL** section:

- **#2 "Every server query touching a tenant page filters by tenant."** `getEntity` and `getSharedBooks` in `src/app/encyclopedia/[name]/layout.tsx` are global by construction — they read `entities.books[]`, which is corpus-wide.
- **#3 "Pre-computed cross-references are not safe in embed mode."** The entity co-occurrence graph is exactly a corpus-wide pre-computed cross-reference, and it is not gated behind any `embedPolicy.show*` flag.

Consistent with the known behavior recorded in memory as "BPH subdomain serves GLOBAL routes" — this is the concrete public consequence.

## Why it isn't caught today

`node scripts/audit-bph-leaks.mjs` exits non-zero when an **anchor or one-hop redirect resolves off-subdomain**. These pages leak while staying *on* the subdomain: the URLs are relative (`/book/...`), so they resolve to `bph.sourcelibrary.org/book/...` and pass the audit. The leak is the **content**, not the hostname. The 2 absolute `https://sourcelibrary.org/` references per page are in metadata (og:image), not anchors.

## Options

1. **404 these routes on tenant hosts.** Cheapest and clearly correct — the encyclopedia is not part of any partner's curated reading room. Would need a proxy rule or a host check in the route.
[...]

PR #3568, opened 2026-08-02
fix(review): repair the two 504'd volunteer queues, add notes everywhere, route people to the work

JDerekLomas:
Closes part of #3560.

132 people signed up to volunteer. `volunteer_ratings` held **zero rows, ever**. Three independent causes were live simultaneously; this PR fixes all three.

## 1. Two of the three queues were returning 504

`/api/review/hallucination/next` and `/api/review/scan-quality/next` picked an item by `$sample`-ing `pages` — **19.1M documents** — with predicates no index can serve (`{'ocr.data': {$regex: '<image-desc'}}` and `{archived_photo: {$exists: true}}`). A full collection scan per item fetch, behind `maxDuration = 15`.

Measured against production 2026-08-02 with a browser UA:

| queue | before | after |
|---|---|---|
| hallucination | **504** after 17.8s, 23.2s on retry | 61 ms |
| scan-quality | **504** after 17.2s | 45 ms |
| gallery-quality | 200 in 8.2s | 23 ms |

Running the same aggregation locally does not return inside two minutes either — that is the confirmation, not a coincidence. `gallery-quality` survived only because it reads the much smaller `gallery_images` (207K), and 8.2s per item is its own kind of unusable, so it moves too.

**The fix is architectural, not a tuning knob.** `scripts/maintenance/build-review-candidates.mjs` precomputes a bounded `review_candidates` pool offline — the same relationship `gallery_images` has to `pages.detected_images` — and all three routes read it through `src/lib/review-candidates.ts`.

Two properties of the builder are load-bearing rather than incidental:

- **It never scans `pages`.** It iterates books and queries their pages via the `{book_id, page_number}` index.
- **It caps pages per book** (`--per-book`). An unbounded draw would let one 900-page volume dominate the pool and quietly turn "quality of the corpus" into "quality of that book".

Every candidate also carries a `stratum` (language / century / provider). Nothing reads it yet — the live queues still sample at random — but a stratified draw cannot be reconstructed after the fact, and the sampled assay in #3560 will need it.

Seeded on Hetzner: 3,000 items per queue, drawn from ~1,950 books each.

## 2. Nothing linked to the queues

`git grep` found no inbound link from the header, footer, homepage, or `/contribute`.

- The `/contribute` card titled **"Rate illustrations"** — whose copy describes `/review/gallery-quality` exactly — linked to `/gallery`, which has no rating control. It was the closest the site came to routing anyone into volunteer work, and it missed.
- The volunteer form's success screen said *"we'll be in touch soon"* and offered no next action. All 132 rows still read `contacted: false`. It now hands people something to do in the next thirty seconds, with a comment explaining why it must never become a dead end again.
- `/review` added to the footer's Participate column (en + es).

**`/review` and `/api/review` are added to `tenant-global-paths`.** Items are drawn across every visible book, so on a partner subdomain a reading room would hand its visitors *other libraries'* pages to judge — the #3364 content leak, in a surface that also invites people to act on it. The API prefix is listed separately because the queue pages render client-side and fetch from whatever host served them.

## 3. There was no way to say anything the buttons couldn't

A four-button scale only answers the question we thought to ask, and these volunteers are specialists — a Latinist looking at a page knows things our rating options cannot express ("not a hallucination, that's bleed-through from the facing leaf").

- The note box lives in `QueueShell`, so **every queue has one by construction**, including any added later.
- `rating` becomes **nullable**: a volunteer can leave a note *without* being forced into a button they don't believe. Coercing that into a rating to satisfy a `NOT NULL` would corrupt the very distribution the ratings exist to measure.
- Consequence, and it is a real one: **every rate and agreement statistic must filter `rating IS NOT NULL`.** `/api/review/stats` now does, and the migration adds a partial index plus a column comment so the next person writing that query sees the constraint.

## Migration: applied ✅

`supabase/migrations/20260802000000_volunteer_ratings_note.sql` was applied to production on 2026-08-03 (via `SUPABASE_DB_URL` from secret-lover — note that the password contains an `@`, so `psql` mis-parses the URL and the host resolution fails; split on the LAST `@`).

Verified in `information_schema`: `note text` present, `rating` now `is_nullable = YES`, `volunteer_ratings_has_signal` constraint in place.

**The compat shim has been removed** (commit `chore(review): drop the note-column compat shim`) — the submit route is back to a single insert.

## Testing

- Full unit suite: 1540 passed / 113 files.
- New `tests/unit/image-extraction-filter-parity.test.ts` runs both twins over the same fixtures. **Negative control performed**: deleting one rule from the `.mjs` twin turns it red (2 failures); restoring it turns it green. Per the repo rule that a guard which only catches deletion is documentation with a green checkmark.
- `tests/unit/tenant-global-paths.test.ts` caught that `/api/review` has no handler at its own path. Rather than special-case it, the wiring check now searches child segments recursively — which also removes the existing hardcoded exception for `/api/explore`. A typo still fails.
- `npx tsc --noEmit` clean.

## Out of scope

- **`/volunteers` is kept, deliberately.** It reads as a duplicate of `/review`, but it is a noindex invitation page written for outreach emails — precisely the landing page for writing to the 117 volunteers. It is on the tenant block list with `/review`.
- **The queues still sample at random, not stratified.** Fixing the 504 does not make this a publishable instrument: there is no second rater and no agreement figure, so it remains a defect-finder. The stratified double-judged assay is #3560.
- **Attention checks.** `review_candidates` carries `is_gold` / `gold_answer` scaffolding and `nextCandidate()` already strips `gold_answer` before it reaches the client, but nothing seeds gold items yet.
- The two workers still carry their own inline copies of the skip-filter logic; folding them onto the new shared twin is a separate change.
- Emailing the 117 volunteers — the actual point of all this — is a follow-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -123,6 +123,14 @@ Each time, two artifacts produced by **different code at different times** were 
 
 Full postmortem: `.claude/handoffs/2026-07-27-bulk-jp2-leaf-offset.md`.
 
+## A request-path query must not scale with the corpus
+Two of the three volunteer review queues returned **504** for months (#3568): each picked one item by `$sample`-ing `pages` — **19.1M docs** — with a predicate no index can serve (a regex on `ocr.data`; `archived_photo: {$exists}`). That is a full collection scan per request, behind `maxDuration = 15`. It was never going to work at corpus scale and it degraded **quietly** as the corpus grew, because nothing alarms on "slower every month" and the client only ever showed "Network error".
+
+- **The tell is the query shape, not the timing:** an unindexed predicate on a collection whose size tracks the corpus. `pages` is 19.1M and growing; `gallery_images` is 207K. Confirm offline — the same aggregation also fails to return locally inside two minutes.
+- **Which one survived is the lesson.** `gallery-quality` worked only because `gallery_images` is a *materialized view* — and still cost 8.2s/item, its own kind of unusable. **Precompute a bounded pool** (`review_candidates`, built by `scripts/maintenance/build-review-candidates.mjs`, read via `src/lib/review-candidates.ts`) and the same work takes 23–61ms.
+- **A builder that feeds such a pool must cap items per book.** An unbounded draw lets one 900-page volume dominate and silently turns "quality of the corpus" into "quality of that book" — and write the `stratum` at build time, because a stratified draw cannot be reconstructed afterwards.
+- Same family as the `/explore` prerender timeout (#3373), where counts over `entities` sit close to `maxTimeMS`: a query that merely *fits* today is a deadline you have already scheduled.
+
 ## Data Protection — CRITICAL
 - **NEVER** delete books, pages, or source material without explicit confirmation
 - **NEVER** batch delete — list items first, wait for approval
--- a/scripts/lib/image-extraction-filter.mjs
+++ b/scripts/lib/image-extraction-filter.mjs
@@ -1,0 +1,65 @@
+/**
+ * Scripts-side twin of src/lib/image-extraction-filter.ts.
+ *
+ * Parity-pinned by tests/unit/image-extraction-filter-parity.test.ts —
+ * change both sides together.
+ *
+ * Exists so batch scripts can apply the SAME "would the extraction worker
+ * keep this page?" test the app applies, without a fourth hand-rolled copy.
+ * (scripts/workers/image-extract-worker.mjs and pipeline-orchestrator.mjs
+ * still carry their own inline copies; folding those in is a separate change.)
+ */
+
+export const SKIP_MARKUP_RULES = [
+  { type: 'symbol', significance: '*' },
+  { type: 'stamp', significance: '*' },
+  { type: 'ornament', significance: '*' },
+  { type: 'blank', significance: '*' },
+  { type: 'exlibris', significance: '*' },
+  { type: 'bookplate', significance: '*' },
+  { type: 'decorative', significance: 'low' },
+  { type: "printer's mark", significance: 'low' },
+  { type: 'photograph', significance: 'low' },
+  { type: 'photographic', significance: 'low' },
+];
+
+export const IMAGE_CANDIDATE_PAGE_TYPES = [
+  'illustration',
+  'diagram',
+  'map',
+  'frontispiece',
+  'mixed',
+  'title-page',
+];
+
+export function shouldSkipPageByMarkup(ocrData) {
+  if (!ocrData) return false;
+  if (ocrData.includes('<detected-images>')) return false;
+  const tags = [...ocrData.matchAll(/<image-desc([^>]*)>/g)];
+  if (tags.length === 0) return false;
+  for (const m of tags) {
+    const type = m[1].match(/type="([^"]+)"/)?.[1];
+    const sig = m[1].match(/significance="([^"]+)"/)?.[1];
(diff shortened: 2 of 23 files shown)


JDerekLomas:

`https://sourcelibrary-v2-cysafts0f-dereklomas-projects.vercel.app` (first Vercel check was the usual false-fail; the retry went Ready, `test` passes, DCO passes).

**All three queues serve real items:**

| queue | production before | preview now | item served |
|---|---|---|---|
| hallucination | **504** @ 17.8s / 23.2s | 200 @ 4.0s | *mKha' 'gro snying thig gi khrid yig zab lam*, p12 |
| scan-quality | **504** @ 17.2s | 200 @ 2.5s | *A Discourse of the Damned Art of Witchcraft*, p14 |
| gallery-quality | 200 @ 8.2s | 200 @ 2.2s | *Prejudices of the Modern Age…*, p888 |

Those preview timings include cold start on an unwarmed deployment — the underlying Mongo query measured 61ms / 45ms / 23ms against the seeded pool.

`gold_answer` confirmed absent from every response body, so attention-check items can't be gamed once they're seeded.

**Note paths, exercising the compat shim (migration deliberately not yet applied):**

- rating + note → `200 {"ok":true,"noteStoredInDetail":true}`, and the note was confirmed present in Supabase as `detail.note` — nothing lost while the column is missing
- note only, no rating → `503 "notes are not enabled yet — please include a rating"` — honest refusal rather than inventing a rating
- neither → `400 "a rating or a note is required"`

The three verification rows were then deleted by exact `item_id` marker; `volunteer_ratings` is back to 0 rows, so the first real submission is still the first.

**Pool seeded in production** (run from Hetzner — the local run died on a dropped Atlas handshake): 3,000 candidates per queue from ~1,950 books each. Top strata for hallucination: Latin/16C 422, Latin/17C 309, Latin/15C 222, Tibetan/18C 163, German/18C 144. The staging copy of the script has been removed from the box; after merge it runs from `scripts/maintenance/`.

**Follow-up not in this PR:** a cron on Hetzner to rebuild the pool as the corpus grows and volunteers exhaust items. It can only be scheduled after merge, since the box runs from `main`. Keep it off `:00` — nine schedules already converge there (see CLAUDE.md on cron convergence).

JDerekLomas:
## Post-migration verification (2026-08-03)

Migration applied to production, shim removed, re-verified on the new preview `sourcelibrary-v2-enhl6nujc`:

| submission | before migration | now |
|---|---|---|
| note only, no rating | `503` — "notes are not enabled yet" | **`200 {"ok":true}`** |
| rating + note | `200` (note fell back into `detail`) | **`200 {"ok":true}`** |
| neither | `400` | `400` — unchanged |

Read back from Supabase, the two rows landed in the real column, not the fallback:

```
__verify_noteonly | rating: null     | note: "post-migration check — note with no rating"
__verify_both     | rating: "matches"| note: "post-migration check — rating plus note"
```

`rating: null` persisting is the point — a volunteer can now say something without being pushed into a button they don't believe. Reminder for anyone querying this table: **every rate and agreement statistic must filter `rating IS NOT NULL`**, which is why the migration adds the partial index and the column comment.

Verification rows deleted by exact `item_id`; `volunteer_ratings` is back to **0 rows**, so the first real submission is still the first.

`npx tsc --noEmit` clean, full suite 1540 passed / 113 files.

Ready to merge. Note that merging does not deploy — this needs `npm run deploy:prod` from `main` afterwards for readers to see any of it.

Issue #3511, opened 2026-08-01
/contribute promises 'no account needed' editing that no route provides


JDerekLomas:
From an external security review filed via the public MCP `submit_feedback` tool (Feedback-ID: 6a6d49ba5797a2a6ecce4ba4), item 2. The reviewer flagged a discrepancy between the README ("authenticated / internal", session + editor role) and the contribute page ("no account needed"), and asked which is true, on the theory that open writes would make page vandalism trivial.

**Writes are not open.** `PATCH /api/pages/[id]` is wrapped in `withAuth`, and additionally returns 403 when there is no tenant context — so it is unreachable from `sourcelibrary.org` entirely. Nothing in `src/` calls it from the client either.

**The defect is the other direction.** `src/app/contribute/page.tsx:130` says:

> Every book has an edit button on every page. No account needed.

I could not find a route that makes that true. Either the reader-facing edit path exists and I missed it, or we are promising a contribution flow that does not work — which for a project asking people to help with transcription is worse than a security hole, because it silently loses volunteers.

**Two things to settle:**

1. Does a reader-facing edit path exist? If yes, document it and correct the README. If no, fix the copy to describe what contributing actually involves today.
2. Separately: `PATCH /api/pages/[id]` gates on `withAuth`'s default `minRole: 'reader'`. On a tenant subdomain that means **any signed-in member of that tenant can overwrite OCR or translation text**. The session cookie is set on `.sourcelibrary.org`, so anyone signed in to the main site who visits `bph.sourcelibrary.org` is in scope. `createRevision` runs first so it is recoverable, and no client calls the route — but `editor` is the right gate for a write that replaces a page's text.

(2) is the part with teeth. Small change; kept out of #3505/#3508 because neither is about page writes.

PR #3574, opened 2026-08-03
fix(security): editor-gate the three uncalled AI page routes (ops#6)

Mayank-PPL:
## The gap

`POST /api/pages/[id]/{modernize,ask,detect-split}` and their `[tenant]` twins had **no auth wrapper at all** — not a drifted `minRole` as in #3511, nothing. Each took a page id, called Gemini on our key, and two of the three wrote to `pages`. `{"regenerate": true}` skipped the cache on demand, so an anonymous caller chose how much we spent.

`resolveTenantId()` on the twins is a slug→UUID lookup, not an authorization check. It was the only thing in front of them.

| Route | Writes | Gemini | Cache |
|---|---|---|---|
| `modernize` | `pages.modernized.*` | yes | hash of translation, `regenerate` bypass |
| `ask` | none | yes — **`customPrompt` straight from the body** | none |
| `detect-split` | `pages.split_detection` | yes, **vision** | none — always regenerated |

## Blast radius: none

Nothing calls these three — not `src/`, not `scripts/`, not the MCP server. The `GET` cached read on `modernize` stays open. No user-facing behaviour changes.

`transliterate` is **deliberately not gated here.** It's the one route in the group with real anonymous reader traffic: `TranslationEditor.tsx:830` auto-fires it from a `useEffect` when the panel opens, and the toggle at `:1338` sits outside every `<AuthCheck>`. Editor-gating it would break signed-out reading of non-Latin-script books. It gets `anonActionGate` on the cache-miss path — a rate limit, not a wall — in a follow-up PR, along with the client-side 401 handling that has to ship with it.

## Also in scope

Since no internal caller sends a request body at all (`src/lib/api-client/pages.ts:134,142` post nothing):

- **`model` no longer read from the body** on `modernize` and `transliterate`. A caller-chosen model on a paid call with no consumer isn't worth keeping.
- **`customPrompt` no longer read** on `ask`, which made it an arbitrary-prompt proxy to Gemini on our API key.
- **`ask` and `detect-split` now call `logGeminiCall`.** They didn't, so their spend did not exist in `gemini_usage` — we could not see whether anyone was using them. Billed per **response**, not per saved page, since Gemini charges even when the JSON fails to parse.

## Guard

The six routes join `CASES` in `tests/unit/page-write-auth.test.ts`. The completeness walker's criterion changes from "assigns caller-supplied text into a page field" to "writes page state **or** spends on Gemini" — the old wording explicitly excluded `modernize`/`transliterate` on the grounds that they write only model output, which was correct about text integrity and silent about cost.

**Both negative controls were run**, per the "a test that greps source is not a guard" rule:

- Deleting `{ minRole: 'editor' }` from `detect-split` → 2 cases red, restored → green.
- Dropping `modernize` from the walker's covered set → walker fails naming that exact path. So the broadened regex genuinely finds these files; the allowlist didn't merely grow.

`npx tsc --noEmit` clean, eslint clean, 1475/1475 unit tests pass.

## Out of scope

- The issue's "75 of 108 write routes lack an auth wrapper" — that's a search result, not a finding. Many carry Stripe signature verification, `REVALIDATE_SECRET`, or in-handler session checks. Only the routes read and confirmed are touched here.
- `pages/batch` — cheap read, capped at 20 ids, real prefetch traffic, no spend. Unchanged.
- The sibling ops issue's routes.
- Any reader-facing "modernize" feature. Gating it today doesn't preclude one; it would arrive with its own anon-gate decision.

### DIFF
--- a/src/app/api/[tenant]/pages/[id]/ask/route.ts
+++ b/src/app/api/[tenant]/pages/[id]/ask/route.ts
@@ -1,8 +1,11 @@
-import { NextRequest, NextResponse } from 'next/server';
+import { NextResponse } from 'next/server';
 import { getGeminiClient } from '@/lib/gemini-client';
 import { DEFAULT_MODEL } from '@/lib/types';
 import { MODEL_PRICING } from '@/lib/ai';
 import { resolveTenantId } from '@/lib/tenant-context';
+import { withAuth } from '@/lib/auth-helpers';
+import { logGeminiCall } from '@/lib/gemini-logger';
+import { getTriggerSource } from '@/lib/cron-auth';
 
 interface Message {
   role: 'user' | 'assistant';
@@ -155,19 +158,21 @@ Please answer their question helpfully:
 
 Answer:`;
 
-export async function POST(
-  request: NextRequest,
-  { params }: { params: Promise<{ tenant: string; id: string }> }
-) {
+// Gated at `editor` (ops#6) — see the note on the unscoped twin. `customPrompt`
+// is no longer read from the body.
+export const POST = withAuth(async (request, session, context) => {
+  const startTime = Date.now();
+
   try {
-    const { tenant, id: pageId } = await params;
-    
+    const { tenant, id: pageId } = await context.params;
+    const triggeredBy = getTriggerSource(request);
+
     // Resolve tenant slug to UUID (for validation)
     const tenantId = await resolveTenantId(tenant);
     if (!tenantId) {
       return NextResponse.json({ error: 'Unauthorized' }, { status: 403 });
     }
-    
+
     const body = await request.json();
     const {
       question,
@@ -176,9 +181,6 @@ export async function POST(
       bookTitle,
       bookAuthor,
       pageNumber,
-      customPrompt,
-      authorSearchTerms,
-      personaName
     } = body;
 
     if (!question) {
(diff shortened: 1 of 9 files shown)


JDerekLomas:
**Triage finding (read-only backlog sweep, 2026-08-28): core fix already shipped via #3600.**

`main`'s \`src/app/api/pages/[id]/detect-split/route.ts\` already wraps \`POST\` in \`withAuth\` — landed via commit 5cfd4a0b, "fix(api): gate four Gemini-spending routes that were open to anonymous callers (#3600)". That's the same editor-gate this PR (ops#6) sets out to add, so the security fix itself is superseded.

What's NOT yet on main: the usage-logging additions in this PR (\`logGeminiCall\`, \`getTriggerSource\`) for these three routes. That overlaps with #3684 (worker usage logging) — checked and #3684's own coverage is partial too (only \`embed-gemini.mjs\` got usage logging via #4162; \`ocr-correct-grounded.mjs\`, \`transliterate-greek.mjs\`, and these three page routes did not). So there may be a real, narrow follow-up here: logging Gemini spend on \`detect-split\`/\`modernize\`/\`transliterate\`. Recommend closing this PR's auth-gate claim as done, and re-scoping the logging piece (rebased onto main) if still wanted — note **#3575 is stacked on this branch** and will need to be re-pointed at `main` once this is resolved.

JDerekLomas:

PR #3575, opened 2026-08-03
fix(reader): meter anonymous transliteration instead of walling it (ops#6)

Mayank-PPL:
> **Stacked on #3574** — base is `fix/gate-ai-page-routes`, not `main`, because both PRs touch the two `transliterate` routes and `page-write-auth.test.ts`. Merge #3574 first and this retargets cleanly. The diff shown here is this PR's changes only.

## Why this one is different

Second half of ops#6. `transliterate` is the only AI page route with **real anonymous reader traffic**: `TranslationEditor.tsx:830` auto-fires it from a `useEffect` when the transliteration panel opens, and the toolbar toggle at `:1338` sits outside every `<AuthCheck>`. Editor-gating it the way #3574 gated `modernize`/`ask`/`detect-split` would break signed-out reading of every non-Latin-script book.

So it gets a **rate limit, not a wall**.

## The split that matters

`anonActionGate` runs on the **cache-miss path only**, copying the precedent at `books/[id]/pages/[pageId]/alignment/route.ts:51-65`. A cache hit is free and ungated — it costs nothing and it's the majority of real traffic. Signed-in users, verified crawlers and warm pings are exempt inside the gate.

**Limit is 200/hour per IP, deliberately generous.** One uncached page equals one call, so a reader working through a Tibetan or Arabic book at 60 pages an hour is an ordinary session, not abuse — and a university reading room behind NAT is a single `cf-connecting-ip` sharing one bucket. The intent is to tighten off `{ event: 'gate_hit', feature: 'transliterate' }` rows in `analytics_events` after a week in production, never up front.

Checked before writing it: `getClientIp` reads `cf-connecting-ip` before `x-forwarded-for` ([rate-limit.ts:180](src/lib/rate-limit.ts#L180)), so this does **not** collapse every front-door reader into the ~15 shared Cloudflare edge addresses. That's the #3487 shape and it would have walled everyone within minutes of deploy.

## Client side — ships with it or the gate is just a bug

The panel auto-fires, so an unhandled 401 shows a reader a failure toast for something they did nothing wrong to hit.

- **`src/lib/api-client/client.ts`** — the response interceptor collapsed every error to `new Error(message)`, so a caller could only recognise an anon-gate wall by regex-matching the copy. `src/app/search/page.tsx:1192` does exactly that today, and it breaks the moment the wording changes. The thrown error now carries `status` / `code` / `signIn` (additive; `err.message` is unchanged).
- **`TranslationEditor.tsx`** — keys on `code === 'SIGNIN_REQUIRED'` and renders a sign-in CTA inside the panel (matching the existing `/search` wall pattern, `?reason=limit`) instead of a toast. Nothing retries. Both call sites — the auto-fire effect and the explicit button — go through one shared `handleTransliterationError`.

## Guard

`tests/unit/transliterate-anon-gate.test.ts`, parameterised over **both twins** (per CLAUDE.md's "check the twins" rule — #3511 found two pairs that disagreed). 12 tests. It exercises the real route module with `anonActionGate` mocked at the boundary.

**Three negative controls, each run and restored:**

| Mutation | Result |
|---|---|
| Delete the gate from the unscoped route | 4 red |
| Delete it from the tenant route | 4 red |
| Move it **above** the cache return | the 2 cache-hit tests go red |

The third is the one that matters. It's the plausible mistake, and it's the one that would wall readers on pages we've already paid for.

`npx tsc --noEmit` clean, eslint 0 errors (11 pre-existing warnings in `TranslationEditor`), 1487/1487 unit tests pass.

## Verification still to do on the preview

- Cached page, anonymous → 200, `cached: true`, no gate consulted.
- 201 × `{"regenerate":true}` from one IP → the 201st is 401 with `Retry-After`.
- Signed out in a browser: open a non-Latin-script book, toggle the panel — first open works; past the cap the panel shows a sign-in prompt, not an error toast.

### DIFF
--- a/src/app/api/[tenant]/pages/[id]/transliterate/route.ts
+++ b/src/app/api/[tenant]/pages/[id]/transliterate/route.ts
@@ -6,6 +6,10 @@ import { resolveTenantId } from '@/lib/tenant-context';
 const TRANSLITERATION_MODEL = 'gemini-3-flash-preview';
 import { logGeminiCall } from '@/lib/gemini-logger';
 import { getTriggerSource } from '@/lib/cron-auth';
+import { anonActionGate, SIGNIN_URL } from '@/lib/anon-gate';
+
+/** See the unscoped twin for why this is deliberately generous. */
+const ANON_TRANSLITERATION_LIMIT = 200;
 
 // Simple hash function to detect OCR changes
 function hashString(str: string): string {
@@ -96,6 +100,22 @@ export async function POST(
         script: page.transliteration.script,
         usage: { inputTokens: 0, outputTokens: 0, totalTokens: 0, costUsd: 0 }
       });
+    }
+
+    // Past this point the request costs money — see the unscoped twin (ops#6).
+    const gate = await anonActionGate(request, {
+      name: 'transliterate',
+      limit: ANON_TRANSLITERATION_LIMIT,
+    });
+    if (!gate.allowed) {
+      return NextResponse.json(
+        {
+          error: `You've used your ${ANON_TRANSLITERATION_LIMIT} free transliterations this hour. Sign in (free) to keep reading.`,
+          code: 'SIGNIN_REQUIRED',
+          sign_in: SIGNIN_URL,
+        },
+        { status: 401, headers: gate.retryAfter ? { 'Retry-After': String(gate.retryAfter) } : {} },
+      );
     }
 
     // Perform transliteration
--- a/src/app/api/pages/[id]/transliterate/route.ts
+++ b/src/app/api/pages/[id]/transliterate/route.ts
@@ -4,6 +4,19 @@ import { performTransliteration } from '@/lib/ai';
 const TRANSLITERATION_MODEL = 'gemini-3-flash-preview';
 import { logGeminiCall } from '@/lib/gemini-logger';
 import { getTriggerSource } from '@/lib/cron-auth';
+import { anonActionGate, SIGNIN_URL } from '@/lib/anon-gate';
+
+/**
+ * Hourly per-IP cap on GENERATED transliterations for anonymous readers.
+ *
+ * Deliberately generous. TranslationEditor auto-fires this route when the
+ * transliteration panel opens, so one uncached page read == one call: a reader
+ * working through a Tibetan or Arabic book at 60 pages an hour is an ordinary
+ * session, not abuse. A university reading room behind NAT is also a single
+ * `cf-connecting-ip` sharing one bucket. Tighten this off `gate_hit` data in
+ * analytics_events, never up front.
+ */
+const ANON_TRANSLITERATION_LIMIT = 200;
 
 // Simple hash function to detect OCR changes
 function hashString(str: string): string {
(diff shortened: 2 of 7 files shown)


Mayank-PPL:
## CI note — read before trusting the check list

**The `test` check does not run on this PR, and its absence is not a pass.** `unit-tests.yml` triggers on `pull_request: branches: [main]`, and this PR's base is `fix/gate-ai-page-routes`. It'll run automatically once #3574 merges and GitHub retargets this to `main`. Until then the only evidence is local: `npx tsc --noEmit` clean, eslint 0 errors, **1487/1487** unit tests pass (1475 before, +12 from the new file).

The `Vercel – sourcelibrary-v2` failure is the known stale-status flake — same generic `vercel.link/3Fpeeb1` that showed on #3574. The actual deployment for `a36959a9` is **Ready**.

## Verified on the preview

`https://sourcelibrary-v2-c9z1izkfs-dereklomas-projects.vercel.app`

| Check | Result |
|---|---|
| `POST /api/pages/<id>/modernize` anon | 401 — #3574's gate intact on this branch |
| `POST /api/pages/<id>/ask` anon | 401 |
| `POST /api/pages/<id>/detect-split` anon | 401 |
| `POST /api/pages/<id>/transliterate` anon | **404 Page not found** — reaches the handler, so this PR did **not** wall it |

That last row is the one this PR is about: a 401 there would mean I'd broken signed-out reading.

## Not verified end-to-end, deliberately

**Tripping the 200/hour cap.** Doing it for real requires 200 uncached transliterations against real pages — 200 paid Gemini calls to prove a rate limiter works. Not worth it. What stands behind it instead:

- Unit tests with `anonActionGate` mocked at the boundary, covering allowed / refused / `Retry-After` present / `Retry-After` absent, across both twins — including the negative control where moving the gate above the cache return turns the cache-hit tests red.
- `anonActionGate` itself is unchanged and already in production on `embassy/voice`, `search/unified`, and the alignment route.

**The browser CTA past the cap.** Same reason. The 401 → `code: 'SIGNIN_REQUIRED'` → CTA path is unit-covered on the server side and read through on the client, but nobody has watched the panel render the prompt after a real wall. Worth one manual pass on prod after deploy — open a non-Latin-script book signed out and confirm the panel degrades to a sign-in prompt rather than an error toast.

## After deploy

Read `analytics_events` for `{ event: 'gate_hit', feature: 'transliterate' }` after a week. A non-trivial count means 200/hour is catching readers rather than abusers, and the limit should go up. That's the number that decides whether the constant was right — it was chosen from reasoning about reader behaviour, not from data.

JDerekLomas:
Review 2026-08-27: superseded by #4205, which gated both transliterate twins at limit 15 with allowBotBypass:false after this PR branched. This PR's gate (limit 200, 401, default bot bypass ON) would be a regression if rebased over main. Salvageable: the machine-readable error code + in-panel sign-in CTA in TranslationEditor/api-client, and the twin-parameterized test — rewritten to main's 429 shape. Recommend close-or-rework; do not merge as-is.

Mayank-PPL:
Closing per @JDerekLomas's 2026-08-27 review. Verified against `origin/main` before closing:

**Server half — superseded, and merging this would be a regression.** #4205 gated both twins after this PR branched: [`pages/[id]/transliterate/route.ts:100`](../blob/main/src/app/api/pages/%5Bid%5D/transliterate/route.ts#L100) and the `[tenant]` twin at `:109` both carry `anonActionGate({ name: 'transliterate', limit: 15, allowBotBypass: false })` returning **429**, on the cache-miss path only — the same cache-hit-is-free split this PR argued for. Rebasing this would loosen 15 → 200, flip `allowBotBypass` back to the default ON, and change 429 → 401.

Also worth noting for anyone reading the description: the base branch `fix/gate-ai-page-routes` belongs to #3574, which was **closed unmerged**, so the stack this PR was written against no longer exists.

**Client half — never shipped, and it is a live bug on prod.** `api-client`'s interceptor still collapses every error to `new Error(message)`; `/search` still regex-matches `/free searches this hour/i` to recognise its own wall; and both transliteration call sites in `TranslationEditor` still end in `toast.error('Transliteration failed: ...')`. Since the panel auto-fires the route from a `useEffect`, a signed-out reader who trips the cap gets a red failure toast — and 15/hour is easier to hit than the 200 this PR assumed.

Carried forward in **#4477**, rewritten to main's 429 shape: the machine-readable `code`/`sign_in`, the in-panel sign-in CTA, and the twin-parameterised test with the "gate moved above the cache return" negative control. The limit itself is left exactly as #4205 set it.

Issue #3576, opened 2026-08-03
gemini_usage cost_usd: two price tables disagree 3.3×/5× for the same model


JDerekLomas:
Split out of PR #3379 (merged — its comment thread is closed history, so the open work needs its own issue). Surfaced while auditing for silent-drift defects in PR #3569.

## The defect

`gemini-3.1-flash-lite` is priced two different ways for the same model, per 1M tokens:

| rate | files |
|---|---|
| **0.075 / 0.30** | `scripts/workers/enrich-worker.mjs` (OCR), `src/lib/ai.ts`, the eval lane |
| **0.25 / 1.50** | `scripts/workers/translate-worker.mjs`, `src/lib/gemini-logger.ts`, `scripts/workers/batch-collector.mjs`, `scripts/workers/lib/supabase-usage-logger.mjs`, `scripts/maintenance/reconcile-batch-usage.mjs`, `scripts/maintenance/backfill-batch-costs.mjs` |

3.3× on input, 5× on output.

- **An OCR row and a translation row for the same model are costed differently.** `gemini_usage.cost_usd` is a blend of two rates depending on which worker wrote it, so a total across phases has no single definition.
- **Two files inside the app disagree with each other** — `src/lib/ai.ts` vs `src/lib/gemini-logger.ts`.
- It is invisible in output, because a cost figure always looks plausible. Same shape as the stale prompt pin fixed in #3569: a local copy asserting an external truth, drifting silently.

## Neither rate fits the stored data

Solving 400 recent realtime `gemini-3.1-flash-lite` rows for their implied rate:

| candidate | mean relative error |
|---|---|
| 0.075 / 0.30 | **0.39** |
| 0.25 / 1.50 | **1.83** |

So recorded costs were computed with something near the first pair but not cleanly — cached-token pricing or a third historical rate is the likely remainder. The second pair would overstate by ~180%.

`cost_usd` is **computed, never billed**, so this is the whole basis of the cost dashboard.

## Already done (#3569)

- `scripts/lib/model-pricing.mjs` is the single table. It does **not** claim the values are verified against Google's published pricing.
- The disputed rate is preserved in `DISPUTED_LEGACY_PRICING` so historical `cost_usd` stays reproducible.
- Eval-lane consumers migrated (estimates only, safe).
[...]

Issue #3444, opened 2026-07-29
OCR prompt: span-level provenance — regroup by who wrote it, and the road to a @resp facet


JDerekLomas:
> **Rewritten 2026-07-30** after a day of measurement. Several proposals in the original version were tested and **failed**; they are recorded below as rejected rather than deleted, so nobody re-proposes them. Findings published at [/blog/reciting-not-reading](https://sourcelibrary.org/blog/reciting-not-reading).

## The problem, in one line

**Every span of rendered text must be attributable — the page said it, or the model did.**

We have excellent *artifact*-level provenance (every page traces to prompt, model, job, revision). Within a page we had much less. Two symptoms of the same defect:

- **`<insert>` map labels were classified as AI commentary** → the Notes toggle deleted them → blank pages on atlases. Fixed and shipped in #3437.
- **The model reconstructs text it cannot see and presents it as transcription** → a fabricated citation with no marker at all. Unsolved.

## TIER 1 — ship now (evidence: fixed a live reader bug)

**Regroup the annotation list by provenance instead of by display.** No new tags, no removals, no data migration — a reordering plus one sentence:

> Every annotation tag answers one question: is this text printed on the page, or did you write it? Readers can switch your commentary off; they can never be shown a page with the book's own text removed. When unsure, choose a transcription tag.
>
> **From the page:** `<margin>` `<gloss>` `<insert>` `<unclear>` `<term>`
> **Written by you:** `<note>` `<image-desc>`

**Sharpen `<insert>`.** "Boxed text, later additions" is what caused #3437 — a map's cartouche labels *are* boxed, so the model applied the spec correctly and the reader broke:

> `<insert>` — text set apart from the running body: labels printed inside a map, diagram, chart or illustration (cartouches, callouts, keys); boxed or framed text; material added after printing such as a pasted slip or stamped correction.

This is the only change with a demonstrated win. It is what made `<insert>` recognisable as transcription.

## TIER 2 — plausible, additive, NOT yet evidenced

**Bleed-through instruction.** Promising because it asks for a *perceptual* discrimination rather than introspection — which is precisely why the Tier-3 rejects failed:

> Bleed-through from the reverse side appears as faint mirrored text. Do not transcribe it — it is not on this side of the leaf. Where it overlaps real text and makes a word ambiguous, transcribe your best reading inside `<unclear>` rather than choosing silently.

[...]

PR #3578, opened 2026-08-03
docs: a pinned copy of a live artifact drifts, and flatters whoever pinned it

JDerekLomas:
New CLAUDE.md invariant + session handoff, from the work in #3569.

## Why

Two instances in one session, both invisible in every output they fed:

- **`B-current.txt` asserted in its NAME that it was the production OCR prompt.** It was a v10-era reconstruction missing ~2,800 characters including the entire **Output contract** block (#3108). It emitted 47,813 characters of looping `[...]` on a blank page where the live prompt emits none — reported as a fabrication finding, retracted once the baseline was checked. Every study in the #3444 series had used it as "current production."
- **Gemini price constants sat in ten files** and had split into two camps for the same model — `0.075/0.30` vs `0.25/1.50` — so an OCR row and a translation row are costed 3.3×/5× apart and `gemini_usage.cost_usd` is a blend (#3576).

This is one step earlier than the existing "A metric is a claim about an instrument" section: not a metric computed wrongly, but a *baseline* that quietly stopped being what it claims. It sits directly above that section as its sibling.

## Also records a fact that was written down nowhere

**The live OCR prompt is a Mongo row, not a file** — `prompts`, `type:'ocr'`, `is_default:true` (v15). `prompts/ocr/*.md` are snapshots that lag it. `prompts/README.md` previously said production prompts "live alongside the code that calls them," which is untrue for OCR; corrected in #3569.

## Handoff

`.claude/handoffs/2026-08-03-hard-page-corpus-and-stale-baselines.md`, added with a deliberate `git add -f`. Purely technical — no PII, secrets, or business material — so the public repo is the right home per the handoff policy.

Docs only; no code changes.

### DIFF
--- a/.claude/handoffs/2026-08-03-hard-page-corpus-and-stale-baselines.md
+++ b/.claude/handoffs/2026-08-03-hard-page-corpus-and-stale-baselines.md
@@ -1,0 +1,102 @@
+# Hard-page corpus, and four alarming findings that were all instruments — 2026-08-01/03
+
+Sessions covering #3495 (merge), #3469 (hard-page corpus), #3444 (Tier 1/2 studies),
+#3452 (verification), and two silent-drift fixes shipped as #3569.
+
+**The through-line: every alarming finding this week turned out to be the
+instrument, not the corpus.** Four of them, each invisible in its own output, each
+caught only by looking at the underlying artifact. Two user prompts of the literal
+word "spot check" accounted for three.
+
+## Shipped
+
+| PR | what |
+|---|---|
+| #3495 | merged — `page_revisions` corpus docs; CLAUDE.md conflict resolved keeping both sides |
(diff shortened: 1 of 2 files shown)


JDerekLomas:

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. The facts landed by other routes (`prompts/README.md` names the Mongo prompt row; `production-prompt.mjs`; `model-pricing.mjs` + its single-source test), and `tests-that-are-not-guards.md` line 35 now records the opposite of this PR's pricing claim — 0.075/0.30 was the wrong constant (#4581). Merging would re-add a falsified assertion. CLAUDE.md was also split into invariants (738ddecc), so the anchor no longer exists.

Issue #3283, opened 2026-07-20
PDF download option for books (user request)


JDerekLomas:
## Why

Reader feedback (feedback collection `6a5df8c5efd2638dab8859dc`, 2026-07-20, submitter left an email for follow-up):

> "I really love this site but at the moment your download options are limited and ineffective. I would like a PDF option please. EPUB doesn't work (the translations sit on top of text incorrectly making it impossible to read) and the txt files are useless for me personally."

PDF is the one format that renders identically everywhere — no e-reader compatibility lottery (see companion bug on the EPUB side). For a library whose core promise is *reading the original*, a facsimile PDF (page scan + translation) is arguably the most natural download we could offer, and we don't have it. Current menu: 3 TXT variants, 8 EPUB variants, images ZIP.

## Constraints (learned the hard way elsewhere)

- **Vercel serverless has no system binaries** (#2526 — `pdftoppm` was dead on arrival). PDF generation must be pure-JS. No Puppeteer/Chromium print-to-PDF.
- **Cloudflare cuts the origin connection at ~100s of silence** → buffered responses on big books die as 524 and readers save the error page as a corrupt file (this already bit the EPUB/ZIP path; `images-zip` now streams for exactly this reason). PDF must stream too.
- `maxDuration = 300` is already set on the route.

## Proposed plan

**Library: `pdfkit`** — pure JS, works in serverless, is a Node stream (pipe it like `generateImagesZipStream` does), embeds JPEGs natively, supports custom TTF embedding.

**Phase 1 — two formats:**

1. `pdf-facsimile` — page scan on top (or left), English translation beneath/facing. Reuses the existing bounded-concurrency image fetcher (`fetchPageImagesOrdered`) + `pageExportImageUrl()` (the canonical `getPageImageUrl` resolver — required, split-page invariant). Fonts are a non-issue for the original because it's an image. This is likely what the requester actually wants: the book, as a PDF.
2. `pdf-translation` — text-only English translation. Clean typographic layout, embedded serif font (e.g. Noto Serif subset), front matter + colophon mirroring the TXT/EPUB versions, `markForExport()` provenance marking on translation text (same as every other format).

[...]

PR #3580, opened 2026-08-03
fix(export): preserve tables and render Arabic in PDF/EPUB, content API and corpus snapshot

JDerekLomas:
## What

Three defects on the download/export path, found while spot-checking Kitāb al-Bulhān (Bodleian MS Or. 133). The first came from reading the code; **the other two only appeared once I generated a real PDF and rasterised it** — the unit tests for the first fix passed straight over both.

### 1. Tables lost their structure (code trace)

`stripEditorialWrappers()` flattens GFM tables (`| 1 | 23 |` → `1  23`). That is **correct** for the snippet/quote/search surfaces it was written for — its own comment says *"Only the snippet/quote path routes through here."* Three surfaces that serve a page's **whole text for reuse** inherited it by accident:

| Surface | File |
|---|---|
| PDF downloads | `src/lib/pdf-export.ts` (added later, #3285) |
| Bulk content API | `/api/books/[id]/text` — what AI clients and licensees read |
| Corpus snapshot | `scripts/export/build-corpus-snapshot.mjs` — the distributable dataset |

Flattening keeps every cell **value** but discards the **column** it belonged to, and fails silently because the output reads as ordinary prose. 148/366 pages (40%) of this book carry tables, 13,949 cells, widest 57×16.

### 2. Ragged tables crashed the entire download (found by rendering)

OCR'd tables are frequently ragged — a damaged or merged cell drops a pipe, so row lengths vary *within one table*. Page 203 yields rows of 9, 10 and 11 cells. pdfkit sizes columns from the widest declared row, then throws `unsupported number: undefined` on any cell past it and kills the export. Every fixture in the test file had been rectangular, so nothing caught it.

### 3. Arabic rendered as tofu (found by rendering)

`pdf-fonts.ts` bundled Noto Serif only, which has **no Arabic glyphs**. Arabic inside translation glosses (`<note>original: "دري اللون"…`) came out as `▯▯▯` — 97/366 pages, i.e. most of the scholarly apparatus.

A global font swap does **not** work: Noto Naskh Arabic has no Latin glyphs, so it merely inverts the problem. Only per-run switching does.

And the `rtla` OpenType feature is **required**. Without it pdfkit shapes the letters and orders the words correctly but **drops inter-word spaces** — `قارعت الاثنين` → `الاثنينقارعت`. That silently alters text that gets cited, which this repo treats as a quote-integrity failure, so it is not a cosmetic detail.

`fonts.arabic` is `string | null`, never a Latin fallback: with no Arabic face the `.notdef` boxes are kept deliberately. A visible gap is honest; silently dropped source text is not.

## How

- **`stripEditorialWrappers(text, { keepTables })`** — opt-**in**, so all ~40 existing snippet callers are unchanged. Mirrored in the `.mjs` twin (parity-pinned by `ngram-normalize.test.ts`).
- **`splitPdfBlocks()` / `writePdfBody()`** lay tables out with pdfkit's native `doc.table()` (≥0.16; we pin `^0.18`), rectangularizing ragged rows. Falls back to pipe-delimited text if table support is absent, so a version skew degrades rather than 500s.
- **`writeScriptAwareText()` / `splitScriptRuns()`** switch face per script run. Noto Naskh Arabic bundled (OFL-1.1, 178KB; `next.config.ts` already globs `src/assets/fonts/**`).
- **`pipeTableToHtml()`** gives the EPUB path real `<table>` markup — `markdownToHtml()` had *no* table handling at all, so tables rendered as literal pipe rows in a `<p>`.

## Verification

- 17 cases in `tests/unit/export-table-preservation.test.ts`, exercising the real functions rather than grepping source.
- **Negative-controlled**: removing the `keepTables` plumbing turns the suite red, green on restore. Per CLAUDE.md, a guard whose only failure mode is deleting a line is documentation with a checkmark.
- **Rendered and visually inspected**: a real PDF from live data. Page 203's Coptic calendar renders as a bordered 4-column table with a bold header (was `1  23  2  4`); the ragged 11-column astronomical table renders padded; page 17's gloss reads `[original: "دري اللون" (durrī) – literally "pearl-colored,"]` with correct joining and spacing.
- Full unit suite 1512 passing / 118 files. `npx tsc --noEmit` clean.

## Out of scope

- **`pdf-ocr` / `pdf-both`** remain deferred per #3283. The Arabic work here covers short inline glosses inside LTR prose; full RTL page layout is a different problem and unchanged.
- **Not load-tested at scale** — a 300-page facsimile PDF still hasn't been generated end to end (the July handoff flags the same gap).


JDerekLomas:
### Load-tested end to end (the gap this PR flagged as out of scope)

Generated a real 366-page facsimile PDF from live data and rasterised it. Two things this PR fixes came out of *looking at the output*, not from reading code — and a third finding is filed separately.

**Caught and fixed here:**
- Ragged tables crashed the export outright (`unsupported number: undefined`). Real OCR tables vary in row width; page 203 gives rows of 9, 10 and 11 cells. Every fixture I'd written was rectangular.
- Arabic run continuation leaked across newlines, indenting every line after the first to wherever the previous run ended (visible on page 195). Now scoped per line.

The second regression test needed two attempts. The first fixture put Arabic on only the first line, under which the assertion holds even with the fix removed — it passed 18/18 against deliberately broken code. The negative control caught that; the corrected fixture fails 1/18 when the fix is reverted.

**Confirmed correct at scale:** all 366 images fetched, page 203's Coptic calendar renders as a bordered 4-column table, inline Arabic glosses render with correct joining and spacing.

**Filed separately — #3597:** the route has no headroom. 252 s total against `maxDuration = 300`, peak RSS 707 MB against a 1 GB default, 75.9 MB output. And 6,987 visible books are *larger* than this one. Nothing streams until all images are fetched, so the connection sits silent for ~231 s — Cloudflare drops it well before `maxDuration` would. That is a pre-existing architectural issue, untouched by this PR.

Also filed: #3591 (OCR fabricates text over damaged regions rather than declining) and #3592 (publish coverage and transcription as two separate quality figures).

Issue #3584, opened 2026-08-03
Cover selection can pick another work's title page (bound-with volumes, library surrogates, series title pages)


JDerekLomas:
## What happened

Reported by Derek from the live page for [Magia Adamica](https://sourcelibrary.org/book/magia-adamica-or-the-antiquitie-of-magic-vaughan): the header said 1749 while the summary directly beneath said 1650, and the cover was the title page of **a different book**.

Two independent defects on one record, both now fixed. This issue is for the class, not the instance.

### Defect 1 — invented publication year (fixed)

`year: 1749`, `published: "1749"`, `year_source: 'regex'`. Internet Archive's `date`, `catalog_metadata.date_iso`, the BPH catalogue match, OpenLibrary, and our own AI summary all independently say **1650**. The value matched no source.

It was written at import time (acquisition campaign "Natural Magic & Ritual", 2026-01-03) straight into `published`; a later pass regexed `year` out of `published` and faithfully propagated it. The real IA date landed separately in `catalog_metadata.date_iso` afterwards, and nothing ever compared the two. Note `field_provenance.year` claims `source: import, provider: internet_archive` — that stamp is applied wholesale at import and is **not** evidence of where a value came from.

Two sibling books from the same import wave had it too: *Anima Magica Abscondita* (1888) and *Anthroposophia Theomagica* (1735), both actually 1650. All three corrected; backup at `scripts/output/vaughan-year-fix-backup-2026-08-03.json`.

**Scale.** Of 6,647 visible books where our `year` can be compared against `catalog_metadata.date_iso`, 1,193 disagree (17.9%) — but most of that is legitimate (a 9th-century composition scanned from an 1864 printing; IA storing a hijri year, e.g. 1254 AH = our 1838, where we are right and the catalogue is in a different era). The suspicious direction is **our year LATER than the catalogue's**, since a first edition cannot postdate its own catalogue record: **590 books (8.9%)**, of which 349 are off by more than five years. By writer: 202 with no `year_source`, 106 `gemini`, 41 `regex`. There is also a visible artifact — a run of books all stamped `1803` against catalogue years spread across 1562–1669.

[...]

PR #3585, opened 2026-08-03
audit: detect covers showing another work's title page

JDerekLomas:
Adds `scripts/audit/sammelband-cover.mjs`. Read-only; writes nothing but its report.

Closes part of #3584 (the detector; the sweep triage and the cover-selection fix stay open there).

## Why

Reported from the live page: [Magia Adamica](https://sourcelibrary.org/book/magia-adamica-or-the-antiquitie-of-magic-vaughan) wore the title page of *The Man-Mouse Taken in a Trap*. The Getty copy is a Sammelband — the Man-Mouse fragment is bound in front, and this copy's own title leaf is wanting. Cover selection took the first page classified `title-page` and shipped it. Enrichment, reading the same opening leaves, then wrote a description of the Man-Mouse into `dc_description`, which is what the book API serves to MCP clients.

Neither artifact is wrong alone — the scan really does open with that title page. Paired-artifact class, per CLAUDE.md.

## The detector

A conjunction: the cover names **none** of the book's distinctive title tokens or the author's surname, **and** another title page in the same volume names them with **≥2** signals.

Condition 2 is load-bearing, not defensive coding. A first pass using only condition 1 reported 8.6% of title-page covers as suspect and was measuring the alphabet — it flagged Tibetan, Chinese, Japanese, Arabic, Korean and Tamil books whose romanised catalogue titles legitimately share no characters with their own covers. Requiring the right title page to exist *in the same volume* is what no script boundary can manufacture.

The ≥2-signal bar on the replacement side was added after the pilot's one confirmed false positive: *Book Four — Magick* yields exactly one usable token, and a degraded p.1 carrying it was proposed over the sound decorated title page in use.

Eligibility gates are stated up front in the header comment and every skip is counted and reported by reason, per the convention in the sibling `scan-title-mismatch.mjs`.

## Control

`--control` replays the motivating book in **both** directions: the pre-fix cover (p.7) must be flagged and suggest p.55; the post-fix cover (p.55) must come back clear. The sweep refuses to run if either fails, so a clean report cannot be confused with a dead detector.

The judgment is factored into an exported `judge()` with no I/O, which is what makes the control exercise the real code path rather than a restatement of it. Per the "a test that greps source is not a guard" section — deleting the matching logic makes the control fail.

## Pilot

600 books sampled, 215 judged, 15 suspects. Causes are broader than Sammelbände: library digital-surrogate covers (Cornell, Stanford), series title pages standing in for the work's own (Spinoza's *Ethic*), faded leaves (Hegel, Wolff), and bookplates (Kepler wearing a Persian Radcliffe bookplate).

## Scope — deliberately out

- **No writes, no auto-fix.** Precision on "this cover is wrong" is high; precision on "use this page instead" is lower. Kepler *Opera Omnia Vol. I* proposes the title page of *Volumen Octavum*, which is a `scan-title-mismatch` finding, not a cover pick. The output is a review queue.
- **No change to cover selection itself.** Fixing the selector to prefer a title page that names the book is the real remedy and belongs in its own PR, reviewed against this sweep's output.
- **No year reconciliation.** #3584 covers the 590 books dated later than their source catalogue; different concern, different risk profile.

## Verification

- `--control` passes both directions
- pilot run of 600 books, suspects read individually; the one false positive found is the reason for the ≥2-signal bar
- full corpus sweep (13,802 eligible) run separately; results go to #3584


JDerekLomas:
Two commits added since opening.

**`a934fece` — fold diacritics before matching.** The full corpus run surfaced a false positive in the detector itself: *Ikhwan al-Safa*, whose printed title page reads **IKHWÁNU-S SAFÁ** against a record reading **Ikhwan al-Safa**. Byte-wise prefix matching saw no hit on the correct cover while p.7 carried `purity` + `dowson`, so a good cover was proposed for replacement. Folding applied in `normalise()`, `titleTokens()` and `authorSurname()` — all three, since "Ikhwán" tokenises to the sub-floor "ikhw" without it. Corpus effect: **181 → 175 suspects**, nine dropped (heavily skewed to transliterated titles, which is the tell), three added where folding strengthened the replacement side enough to clear the ≥2-signal bar.

**`5a96cfe2` — a CLAUDE.md paragraph.** Flagging this deliberately as a second concern in one PR, per the one-concern rule: it is doctrine rather than code, and it is the lesson from *this PR's own bug*, so splitting it seemed worse than noting it. Happy to pull it into its own PR if a reviewer disagrees.

The regression control I wrote for the diacritic fix was broken on the first attempt, in a way the existing "a test that greps source is not a guard" section doesn't cover. I hand-wrote the fixture page text and included "Translated from the Hindustani", which matched the title's own `translation` token via the six-character prefix rule — so the control **passed with the fold deliberately removed**. Real code path, exercised, asserting nothing. The actual page in Mongo holds only `# IKHWÁNU-S SAFÁ.`

Two details kept it invisible: the stopword list screens `translated` but not `translation`, and I wrote the fixture *after* knowing the verdict I wanted. The fixture text is now copied from the database verbatim, and the negative control was run rather than reasoned about — stripping the fold turns it red, restoring it turns it green.

Full sweep results and the suspect breakdown are in #3584.

Issue #3614, opened 2026-08-04
prompts: two is_default rows for Standard Summary, and a stale defaults table in data-provenance.md


JDerekLomas:
Found while building `scripts/audit/doc-enum-drift.mjs` (#3473 follow-up). Two separate problems, one live.

## 1. LIVE: two `is_default: true` prompts for the same (type, name)

```
{type: 'summary', name: 'Standard Summary'} -> 2 rows with is_default: true
   one has version: 1, the other has NO version field
```

`.claude/docs/data-provenance.md` states the invariant and even prescribes the check:

> Only one prompt per `(type, name)` should have `is_default: true`
> 4. VERIFY: `db.prompts.countDocuments({ type, name, is_default: true })` must be exactly 1

**Consequence.** `getSummaryPrompt()` resolves via `findOne({type, is_default: true})`, so which of the two it returns is arbitrary — it depends on natural/index order, not intent. Summaries generated through the unversioned row record `prompt_version: undefined`, which breaks the provenance chain that document exists to guarantee. This is not cosmetic: provenance is the thing that lets us say which prompt produced a given piece of reader-facing text.

Also: `prompts.version` is **mixed string and number** across the collection (`1` vs `'v1'`), so any sort or comparison on it is unreliable. One default lookup already printed as `vundefined` and another as `vv1`.

**Fix:** decide which summary prompt is canonical, unset `is_default` on the other, normalise `version` typing, and add the invariant to a standing audit rather than a prose instruction — a rule that must be remembered is a rule that will be missed.

## 2. STALE: the documented "Current defaults" table

`data-provenance.md` says:

| documented | actual (2026-08-04) |
|---|---|
| OCR — Standard OCR **v10** | **v15** |
| Translation — Standard Translation **v8** | **v12** |

Five and four versions behind respectively. Nothing caught it: the doc is living, referenced, undated, and carries no date-plus-quantity pattern, so it passes every check in `doc-staleness.mjs`.

**Fix:** extend `scripts/audit/doc-enum-drift.mjs` to compare the documented defaults against `prompts` where `is_default: true`. Same shape as the enum check it already performs, and it would have caught both of these.

## Why this matters beyond tidiness

[...]

PR #3617, opened 2026-08-04
audit: catch documented enums drifting from production

JDerekLomas:
Follow-up to #3475. Adds `scripts/audit/doc-enum-drift.mjs` and corrects `data-provenance.md` with the real `pages.{ocr,translation}.source` enum.

## Why

`data-provenance.md` documented `page_revisions.source` as five values; production had twelve. The eight undocumented ones covered **111,685 rows**, including the label identifying which 29.5% of the "double OCR" corpus is a maintenance sweep rather than a second reading. Two sessions inferred that the hard way while `distinct('source')` answered it in one query.

Nothing caught it. The doc is living, referenced, undated, and carries no date-plus-quantity pattern, so it passes every check in `doc-staleness.mjs`. Schema drift is a distinct failure mode: the prose stays true-looking while the world adds a value underneath it.

## What the first run found — 14 undocumented values

`pages.ocr.source` (10) and `pages.translation.source` (4), in fields I had never touched by hand. Three matter beyond bookkeeping:

- **`corpus` (~5.7K rows on each field) and `wikisource` (21) are not model output.** Any metric treating `pages.ocr` as "what the model read" is wrong on those rows, and nothing but `source` says so.
- **Repair sweeps write to the live fields too**, not only `page_revisions`: `ia-ocr-repair`, `reocr-contamination-repair-3362`, `spread-split`, `ai-repair-sync`.
- **`songshi-juan56` and `AI generated`** (1 row each) are junk that leaked into an enum. Documented rather than ignored — the alternative is teaching the audit to lie.

## It also caught a flaw in itself

A sampled run reported `pages.translation.source` "all documented". The exact scan found four undocumented values there, the largest on 5,759 rows — five times the stated detection floor. MongoDB's `$sample` uses a pseudo-random cursor that is not uniform, so a value can be missed at any frequency.

A check that can report a false clean is worse than no check, so: **exact is the default**, `--fast` must be opted into, and a sampled run can only print `? INCONCLUSIVE` — never a pass, and it never satisfies `--fail-on-findings`.

## Known cost, stated rather than resolved

Exact means several minutes on `pages` (19.1M docs); `page_revisions` stays exact and fast (8s). I argued earlier that a slow check will not be run habitually, and that is still true — so if this ends up unrun, the fix is to schedule it, not to make it lie.

## NOT verified

The final confirmation run — that `data-provenance.md` now passes its own check — **died on a DNS drop mid-scan and has not been re-run.** The enum table was written from a completed exact scan, so the values are right, but the round-trip has not been closed. Worth re-running before merge:

```
set -a; source .env.production.local; set +a
node scripts/audit/doc-enum-drift.mjs --fail-on-findings
```

## Out of scope

- **#3614**, filed not fixed: two rows carry `is_default: true` for `{summary, Standard Summary}`, violating an invariant this same doc states, so `getSummaryPrompt()` resolves arbitrarily and one path records `prompt_version: undefined`. The doc's "Current defaults" table is also five and four versions behind (OCR v10 vs v15, Translation v8 vs v12).
- Extending the check to cover that defaults table — same shape, would have caught it.

### DIFF
--- a/.claude/docs/data-provenance.md
+++ b/.claude/docs/data-provenance.md
@@ -58,7 +58,7 @@ page.ocr.prompt_version    // "v10"
 page.ocr.prompt_id         // ObjectId-string of the prompt document, or 'hardcoded' / 'custom'
 page.ocr.prompt_hash       // md5 of prompt content (cryptographic verifier)
 page.ocr.prompt_name       // "Standard OCR" | "Latin OCR (Neo-Latin)" | etc
-page.ocr.source            // 'ai' | 'batch_api' | 'manual' | 'contributor' | 'skip'
+page.ocr.source            // see the full enum below — 16 values, not 5
 page.translation.prompt_version
 page.translation.prompt_id
 page.translation.prompt_hash
@@ -127,7 +127,8 @@ For batch jobs the four fields are stamped on the `batch_jobs` row at submission
   data: "the full previous text content",
   source: "batch_api" | "ai" | "pipeline_preview" | "manual" | "skip" | "system" |
           "maintenance" | "mineru" | "realtime_api_sequential" | "unknown" | null |
-          "<sweep-label>",   // ad-hoc, e.g. "shift-repair-erara-2026-07"
+          "<sweep-label>",   // ad-hoc, e.g. "shift-repair-erara-2026-07",
+                             // "reocr-download-failure-fix-2026-07"
   model: "gemini-3-flash-preview",
   prompt_version: "v10",
   job_id: "job_abc123",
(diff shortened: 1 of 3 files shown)


JDerekLomas:
**Verification passed** — the round trip this PR was missing is now closed.

```
✓ page_revisions.source       12 values, all documented   (full scan, 326,872 docs)
✓ pages.ocr.source            16 values, all documented   (full scan, 19,132,133)
✓ pages.translation.source    11 values, all documented   (full scan, 19,132,133)
```

Exact scans, no sampling, so these are real clean bills rather than the `? INCONCLUSIVE` kind.

Two bugs fixed in `4210b4db`, both found by reading the process list rather than trusting the script's own status output:

- **Hung forever on error.** Switching `process.exit()` → `process.exitCode` (correct in itself) meant the error path stopped force-exiting, and nothing closed the `MongoClient` — so node's event loop never drained. A DNS blip left one run alive **2h43m** holding an open connection pool. Now `try/finally`, and validated the hard way: the next run also hit a network timeout and **exited cleanly**, which is exactly what the zombie could not demonstrate.
- **Phantom detection matched field paths.** `pages.translation.source` was reported as a value "documented but absent from production" — the regex was scraping field names out of prose. Values never contain a dot.

Also documented: **run this on Hetzner.** Exact mode scans `pages` twice at 19.1M docs and outlasts a consumer connection — two of three local attempts died mid-scan. The scans are the point, so the fix is where it runs, not a weaker check.

Ready to merge.

PR #3623, opened 2026-08-04
docs: volunteer-queue handoff + the invariant that page text closes your script tag

JDerekLomas:
Session wrap-up for 2026-08-02/04.

## Handoff

`.claude/handoffs/2026-08-04-volunteer-queues-and-review-design.md` — **132 volunteer signups against 0 ratings**, and the three stacked mechanical causes rather than the tempting product conclusion:

1. Two of three queues 504'd, `$sample`-ing **19.1M** `pages` docs with an unindexable predicate behind a 15s limit (#3568). Now 23–61 ms from a precomputed pool.
2. `git grep` found **no inbound link** to the queues from anywhere on the site; the one card describing the work pointed at `/gallery`.
3. No qualitative channel — four buttons only answer the question we thought to ask.

Plus the pool-hygiene defects found by *working the queues as a volunteer would* (#3586), the review-programme design decisions (#3596), the OCR-adjudication sampling measurements, the PR sweep (41 → 26 open, stale tail 15 → 4), and the dependency audit (30 → 6 production vulnerabilities, #3616).

## The CLAUDE.md invariant

Per the rule that every handoff ends with "does CLAUDE.md need a new invariant?" — yes, and it is genuinely non-obvious.

**OCR output carries an HTML-shaped metadata envelope, and one of its tags is literally `<script>`** — it records the writing system:

```
<scan-quality>good</scan-quality> <language>Latin</language> <script>printed</script>
```

So `pages.ocr.data` is a string containing `</script>`, and the HTML parser ends a `<script>` element at the first `</script` it sees, inside a JS string literal or not. **Any surface interpolating page text into HTML or JS without escaping is broken by ordinary corpus data, not by an attacker.**

Not hypothetical: it broke an offline prototype this session, and the audit that followed found **all 11 JSON-LD sites in `src/` unescaped** while the reader page embeds a 500-char excerpt of page text (fixed in #3609).

The invariant records the fix (`jsonLdHtml()`), the guard, and **the tell** — a `ReferenceError` naming a word from the page's own metadata (`printed is not defined`), because the value inside the envelope tag escaped into the document as markup. That is a sufficiently strange symptom that recognising it is worth more than re-deriving it.

Docs only.


JDerekLomas:

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded by #3609, which landed the durable form: `jsonLdHtml()` in `src/lib/json-ld.ts` pinned by `tests/unit/json-ld-escape.test.ts` (the exact printed-script case, the img-onerror case, and an absence check that fails any component hand-rolling JSON.stringify). The handoff the body advertises was never added to the PR.

Issue #3556, opened 2026-08-02
The 041 item-language filter drops every facing-page bilingual edition (Loeb, I Tatti) from the reference set


JDerekLomas:
## One line drops the entire facing-page scholarly-edition class

`scripts/enrichment/ingest-loc-bulk.mjs`:

```js
const itemLang = (subfieldValues(rec, '041', 'a')[0] || f008.slice(35, 38) || '').toLowerCase().trim();
if (!itemLang.startsWith('eng')) return null;
```

MARC lists a bilingual edition's languages **in order of predominance**. A facing-page scholarly edition puts the *ancient* language first, because the ancient text is the primary content:

- `041 0# $a lat $a eng $h lat` → `[0]` is `lat` → **rejected**
- `041 0# $a lateng $h lat` (concatenated, common in older records) → `startsWith('eng')` false → **rejected**

Both encodings fail. That is precisely the **Loeb / I Tatti / Dumbarton Oaks Medieval Library / Clay Sanskrit Library** form — the dominant vehicle for scholarly English translations of classical and early-modern works, i.e. exactly the priors that defeat our first-translation claims.

### The signature is visible in what survived

**4,622 rows passed with a multi-language item code — 424 distinct forms** (`engsan` 459, `engspa` 439, `englat` 386, `engfre` 331, `engger` 262, `engheb` 228 …). Every one of those survived *only because `eng` happened to be listed first*. The mirror population — same editions, ancient language first — was dropped silently.

### Confirmed downstream

`Rhetorica ad Herennium` (Caplan, Loeb 1954) returns **zero rows** in the extract. Lemay's *Women's Secrets* (SUNY 1992) likewise absent. Both are modern, both squarely inside the 2016 snapshot's period, both verified real priors that the search reported `none_found` for.

## Fix

Test whether `eng` appears in **any** `041$a` (and handle the concatenated form by chunking into 3-character codes), not just the first:

```js
const codes = subfieldValues(rec, '041', 'a')
  .flatMap((v) => v.toLowerCase().trim().match(/.{1,3}/g) ?? []);
const itemLangs = codes.length ? codes : [(f008.slice(35, 38) || '').toLowerCase().trim()];
if (!itemLangs.includes('eng')) return null;
```

Keep the full list on the row (`item_language` becomes an array, or add `item_languages`) so a consumer can tell a monolingual English translation from a facing-page bilingual — that distinction matters for the completeness screen in the demote packet.

[...]

Issue #3524, opened 2026-08-01
55 books badged 'first English translation' while tagged English — three different bugs, not one


JDerekLomas:
Found while evaluating the FT system on 2026-08-01. `{is_first_translation: true, visible: true, pages_translated: >0, language: /^en/i}` returns **55 books**. It looks like one bug. It is three, and they need opposite fixes — **a bulk demote would have destroyed ~22 true claims.**

Evidence: modal `<language>` tag over up to 12 *content* pages per book (≥700 chars of OCR, so covers and flyleaves are excluded) plus a non-Latin script count on the body text. Worksheet: `scripts/output/ft-english-badged-adjudication-2026-08-01.txt`.

**Why page 1 lies:** the first page is nearly always a cover or a flyleaf, and its OCR language tag describes the *modern pencil cataloguing annotations*, not the text. That is how a 12/12-Hebrew book ends up tagged `language: English`.

## Class 1 — the `language` field is wrong; the badge is probably RIGHT (~22 books)

Do **not** touch `is_first_translation`. Fix `books.language`.

| book | modal language | evidence |
|---|---|---|
| Cordovero, *Sefer pardes rimmonim* | hebrew 12/12 | 63,645 Hebrew chars |
| Zevi Hirsch, *Shiv'im tikunei ha-zohar* | hebrew 12/12 | 155,166 Hebrew chars |
| Alkabets, *Ayelet ahavim* | he 12/12 | 50,874 Hebrew chars |
| Poppers, *Sefer me'orot natan* | he 9/12 | 74,895 Hebrew chars |
| + 4 more Hebrew, 9 Latin (Fernel *Cosmotheoria*, Borrichius *Docimastice metallica*, *Ortus sanitatis*, Breviarium Romanum, Harley MS 2407, Bibliotheca Askeviana ×2, Moninckx Atlas), 1 French (*Annales maçonniques*), 1 Italian (*Vita di san Giosafat*), 1 Malayalam (*Jyothisha Marga Deepika*, 6,915 non-Latin chars), 1 German (Andreae, *Geistliche Kurtzweil*) | | |

This is the same class as the known IA/BPH language-mistag problem, surfacing on the FT surface.

## Class 2 — the text really is English; the badge is definitionally wrong (~16 books)

A *first English translation* of a work written in English is not a claim that can be true. **No external catalogue search is required to settle these** — which makes them the cheapest correct demotes in the system.

[...]

PR #3628, opened 2026-08-04
docs(ft): session handoff + the two invariants it earned (#3524, #3556)

JDerekLomas:
Wrap-up for 2026-08-01 → 08-04. Handoff plus two CLAUDE.md invariants.

## The finding worth not losing: a negative result

The `041$a` bilingual fix (#3563) recovered **8,206 rows** and moved catalogue recall **27.0% → 27.5%**. Half a point.

And the two works the entire diagnosis rested on — Caplan's Loeb *Ad Herennium* and Lemay's *De secretis mulierum* — are **still absent** after it, confirmed by querying `title`, `uniform_title` **and** `author`.

That independently corroborates the third explanation already recorded in CLAUDE.md: **the real gate is that `041$h` is required at all.** Three explanations offered, two now measured wrong. Negative results decay fastest, so it's written down.

## Two invariants

**A long corpus loop fails silently in two directions, and `noTimeout` only fixes one.** The 300s watchdog truncates (fourth occurrence this session: demote packet 31/33, screen 2,000/17,072, LoC load part 30/43 — all exit 0, all a prefix). Removing it converts that into a **silent hang**: the cohort search sat 2h48m at 0.0% CPU having written 1,500 of 10,126 efforts. The truncation is the worse of the two, because `latestEffortPerBook()` reads a partial generation as the current one. Diagnose with CPU time — **never `pgrep -f <script>`**, which matches your own ssh command line and reported RUNNING for a job that had stopped 7,500 books earlier.

**JSON round-tripping a Mongo document destroys its Date types.** Finishing a wedged upsert from a saved `.json` wrote `run_at` as a **string** on 10,126 docs where the writer stores a `Date`. BSON orders the two as separate type classes, so `$gte: new Date(...)` matches none and a type-ordered sort never interleaves them — silently breaking the mechanism the whole `search_efforts` design depends on. Caught only because a verification query returned 0 instead of 10,126; all 10,126 repaired. The shortcut around a hung process introduced a subtler bug than the hang.

## Also in the handoff

Five instances **in one session** where the instrument lied rather than the thing measured: a self-matching `pgrep`; sampler tests that passed against a deliberately broken sampler (every fixture had the body at the front, so the wrong strategy scored right by accident); a projection that made all 5,947 badges look unconfirmed; an ingest that reported 43-of-43 success and wrote nothing because `--load` was missing from its own runbook; and a partial load indistinguishable from a complete one.

Plus the state as it stands: both screens complete over all 17,071 eligible books, 440 flagged/held (2.6%), reference set at 126,558 verified against disk, and three bounded review lists (117 translator holds · 26 screen-contradicted badges · 158 classifier-contradicted).

## Note on the handoff file

`git add -f` — `.claude/handoffs/` is gitignored, and CLAUDE.md permits only genuinely public-worthy **technical** postmortems here. This is one: no PII, no business material, no reader data.


JDerekLomas:

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. The `noTimeout` invariant is on main at `first-translation-reference-set.md:479` with the same evidence, the partial-generation caveat at :459 and in `first-translation-claims.md`, and the FT lane has since been rebuilt twice (#3881, #4536) so `search_efforts` is now an archive behind `first_translation_attempts`. Branch conflicts.

Issue #2976, opened 2026-07-04
TTL-expire raw telemetry collections (api_usage, gemini_usage, analytics_events, audit_log) — ~7GB data + ~1.7GB indexes of cache pressure


JDerekLomas:
## Why

The Atlas working set (95GB data + 5.8GB indexes across 117 collections) vastly exceeds cluster cache; measured random-fetch throughput is ~1.6K docs/s, which is what made the /explore/timeline outage possible (#2973). The biggest *removable* cache pressure is append-only telemetry sitting in the same cluster as reader-facing content (sizes 2026-07-04):

- `api_usage`: 4.5M docs, 2.0GB data + **827MB indexes** (6 indexes)
- `gemini_usage`: 3.8M docs, 2.2GB + 518MB indexes
- `analytics_events`: 4.0M docs, 720MB + 282MB indexes
- `audit_log`: 2.5M docs, 783MB + 67MB indexes

## Plan

1. **Decide retention per collection** (proposal: 180d api_usage/gemini_usage — they feed the budget + cost dashboards; 90d analytics_events; 365d audit_log since it's the security trail). ← Derek sign-off.
2. **Verify consumers first**: grep for each collection in `src/` + `scripts/` and confirm no reader needs rows older than the window (e.g. `backfill-usage-daily.mjs` rollups, the /platform analytics routes, api-budget 24h windows). Document findings in this issue.
3. **Ensure daily rollups cover historical reporting** before any expiry (usage_daily etc.). Backfill if gaps.
4. **Create TTL indexes** on the timestamp field (verify exact field name per collection — they differ). TTL deletes trickle gradually, no stall. Add to the index manifest (see companion index-governance issue).
5. **Verify**: collection counts plateau, Atlas cache hit ratio improves, dashboards still render.

## Safety

- TTL is destructive-by-design: do NOT create the TTL index until step 2's consumer audit is posted here and approved.
- `audit_log` retention is a policy question, not just a size one.

## Acceptance
- All four collections have TTL indexes with agreed windows, documented in the manifest; rollups verified; total data size trending down.

Part of the post-#2973 database-lifecycle plan. Related: #2974.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JDerekLomas:
Taking the step-2/3 consumer audit — working in a Claude session.

JDerekLomas:
## Step-2/3 consumer audit (2026-07-05)

[...]

PR #3644, opened 2026-08-05
feat(mcp): log the initialize handshake so we can count clients, not IPs

JDerekLomas:
## Why

Asked "is anyone actually using the MCP server?", the current telemetry can't answer it.

`mcp_tool_calls` knows *which tools* get called (1,982 calls / 30d, led by `search_images`, `search_library`, `get_book_text`). It does not know *who* is calling:

- the server is stateless, so each tool call is its own HTTP request carrying no client identity;
- all `Claude-User` traffic leaves Anthropic egress IPs, so distinct `ip_hash` badly **under**counts installs there;
- crawlers and consumer ISPs rotate addresses, so it **over**counts elsewhere.

Meanwhile `/api/mcp` took ~88k requests / 30d from ~860 IPs — but the single loudest caller is a monitor whose user-agent literally reads `liveness-only, never invokes tools`. Raw request volume reads as adoption when it is mostly directory discovery.

## What

`initialize` is the only MCP message that carries `params.clientInfo`, and the transport consumes it before any of our handlers run — so `POST` reads it off the already-parsed body and writes one row per handshake to a new `mcp_clients` collection: `client_name`, `client_version`, `client_title`, `protocol_version`, `ip_hash`, `user_agent`. Handles batched JSON-RPC (array body). Fire-and-forget, like the existing tool log; a malformed body is skipped, never thrown.

`scripts/analytics/mcp-clients.mjs` reads it, and deliberately reports two numbers per client: handshakes, and how many of those client IPs went on to actually invoke a tool. That's the crawler/user split the request counts hide.

Retention follows the #2976 decision — no TTL, manual pruning only. `mcp_clients` is registered in the `ensure-indexes` manifest and added to the `prune-telemetry` allowlist.

## Verification

`npx tsc --noEmit` clean. Handshake logging verified against the preview deployment with a distinctively-named client, then read back out of the collection (see comment below).

## Out of scope

- **Attributing individual tool calls to a client.** Statelessly that would need session correlation (or a client id echoed on every request); the handshake gets us install counts, which is the actual question.
- **Backfill.** Nothing before this deploy has client identity to recover.
- **Any dashboard surface.** CLI script only for now — worth a panel once there's a few weeks of data to look at.


JDerekLomas:
## Verification

Probed against preview deployments with a positive **and** negative control — an `initialize` carrying a uniquely-stamped `clientInfo.name`, then a read of `mcp_clients` for both that name (must be found) and a name never sent (must be absent). Without the negative control, "found a row" wouldn't prove the query discriminates.

**First run failed, and the failure was the interesting part.** The row was absent after 12s of polling — but present on a re-check a minute later, timestamped at the moment of the original request. A control test through the *pre-existing* `logMcpToolCall` path never landed at all.

Cause: `void writeEntry(...)` is not fire-and-forget on serverless. Vercel freezes the instance once the response is sent, so an insert still in flight is suspended and resumes only if a later request thaws that same instance. Warm production instances usually win the race, which is why 20,890 rows look healthy. **The losses concentrate on cold starts — and a cold start is disproportionately first contact from a client we've never seen**, so the log drops new clients preferentially. Pre-fix `mcp_tool_calls` counts should be read as a floor.

Fixed in bda83ea7: both writers go through Next.js `after()`, bounded at 3s, with a non-request-scope fallback — same shape as `deferDbWrite()` in `src/app/api/track/route.ts`. This makes the existing tool-call log more reliable as a side effect.

**Second run, against a freshly-built (cold) preview:**

- initialize handshake → row present within the polling window, all fields correct (`client_name`, `client_version`, `client_title`, `protocol_version`) — **PASS**, negative control correctly absent
- tool call → `mcp_tool_calls` row present — no regression from the `after()` change
- `scripts/analytics/mcp-clients.mjs` runs and reports its empty state correctly against the real collection

All three probe rows were deleted from production by explicit `_id` afterwards (2 in `mcp_clients`, 1 in `mcp_tool_calls`) so the new instrument doesn't start life with test data in it.

`npx tsc --noEmit` clean. Invariant written up in `.claude/docs/invariants/measurement-instruments.md`.

PR #3651, opened 2026-08-05
feat(feedback): tag submissions with the tenant they came from

laubaumau:
Phase 1 of separating BPH feedback from general Source Library feedback. This change alone ships nothing user-visible; it makes the two streams **separable**, which everything after it depends on.

## The problem

Feedback rows carry no tenant. The only way anything currently answers "is this BPH?" is a string prefix in `scripts/analytics/feedback-triage.mjs`:

```js
if ((r.page || '').startsWith('/catalog/')) return 'partner-cms';
```

`page` is `window.location.pathname`, captured client-side. That heuristic is wrong in both directions:

- **False positives** — the global site has its own `/catalog` and `/catalog/scholar` routes, so ordinary sourcelibrary.org feedback is filed as partner-CMS.
- **False negatives** — a librarian writing from `/catalogue` (no trailing slash) or from the tenant home is not counted at all.

You cannot build a librarian-facing view on top of a guess, so this replaces it with a recorded fact.

## What changed

**`src/lib/feedback-origin.ts`** (new) — pure, I/O-free helpers that derive `tenant_slug`, `surface` (`catalog` | `reader` | `global`) and `embedded`. Unit-tested in isolation.

**`POST /api/feedback`** — stamps `tenant_slug`, `tenant_id`, `tenant_source`, `surface`, `embedded` on every new row. The tenant comes from the proxy's `x-tenant-*` headers, never from the request body: a client-supplied tenant would let anyone file into a partner's queue.

**`GET /api/feedback`** — adds an admin-only `?tenant=<slug>|none` filter so the split is visible immediately. `none` matches missing as well as null, since pre-backfill rows have no such field.

**`scripts/maintenance/backfill-feedback-tenant.mjs`** (new) — tags historical rows. Dry-run by default.

**`scripts/analytics/feedback-triage.mjs`** — prefers `tenant_slug`; the path guess survives only as a fallback for un-backfilled rows, and now excludes `/catalog/scholar`.

## The embed case, and why `proxy.ts` is untouched

`src/proxy.ts` resolves an API call's tenant from the Referer's **first path segment**. For a reader in the Webflow iframe the Referer is `/embed/bph/…`, so that segment is the literal `embed`, not `bph`, and the row would have arrived untagged.

Widening the proxy would put this change inside the tenant-lockdown invariant for no good reason. Instead the feedback route re-reads `/embed/<slug>/` from the Referer itself and resolves it through the `tenants` collection, so an arbitrary Referer cannot invent a tenant. Rows tagged this way record `tenant_source: 'embed-referer'` and are distinguishable from header-derived ones.

## Safety

- **The write cannot fail because of this.** Origin resolution is wrapped; on any throw the row is saved untagged and the response contract is unchanged. `/api/feedback` is the write path for the widget inside the partner iframe, and that widget only checks `res.ok`.
- **The backfill is timid on purpose.** It is the same heuristic this change replaces, so it only touches `/catalog*` rows, excludes `/catalog/scholar`, never overwrites an existing `tenant_slug`, re-asserts "untagged" at write time, and leaves anything ambiguous alone. Rows it writes are stamped `tenant_source: 'backfill-page-heuristic'` and the undo command is printed on completion. Leaving a row untagged is cheap; tagging a stranger's feedback as BPH is not.
- **No PII change.** `ip` and `email` handling is untouched. Stripping them is Phase 2's job, on a separate tenant-locked route. The `?tenant=` filter added here stays behind `withAdminAuth`.

## Verification

- `npx tsc --noEmit` clean
- `npx vitest run` — 1891 passed, 137 files (16 new)
- New tests pin both historic failure cases: `/catalogue` now classifies as `catalog`, and `/catalog/scholar` is excluded from the backfill

## Out of scope

- The librarian-facing feedback page (Phase 2) and the kanban board (Phase 3)
- Running the backfill — deliberately manual, dry-run first
- Any change to how feedback is displayed in `/admin/feedback`

## Reviewer note

`src/app/feedback/page.tsx` is a public route calling the admin-only GET, so it errors for everyone but admins. It is untouched here, but it should not become the librarians' page in Phase 2.


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek** (blocked label), not merged and not closed. The bug is still live — `scripts/analytics/feedback-triage.mjs:39` still guesses tenant from `page.startsWith('/catalog/')` — and main split feedback on a different axis (`channel: mcp|web`, #4560-era), so nothing supersedes this. But it is 41 files against a catalog surface that #3985 reworked, and it sets a tenant-from-proxy-header trust model on an admin filter over a multi-tenant table (auth-adjacent). Needs a decision: is BPH feedback separation still wanted, and is the header trust model reviewed? If yes, it needs a rebase before CI can say anything.

JDerekLomas:
Closing (backlog decision 2026-09-22): 41 files against a catalog surface #3985 reworked, plus a tenant-from-proxy-header trust model on an admin filter that was never reviewed. The live bug it targeted (feedback-triage.mjs guesses tenant from a page-path prefix) is small enough for its own fix; filing it as an issue. Branch kept.

Issue #3653, opened 2026-08-06
MCP: canonical locus addressing + contains_works — the two asks that would change what the library is for (from MCP feedback)


JDerekLomas:
Filed from four MCP `submit_feedback` reports (2026-08-05, `SourceLibrary-MCP/4.3`) written during an agent session checking ~25 circulating Aristotle quotations against the corpus. The reporter's own summary: *"The tools themselves are well-shaped; the friction is almost entirely in metadata, ranking, and the absence of a canonical-locus layer."* Concrete data bugs are split into a companion data-quality issue; this one is API shape only.

Extends #3083 (earlier AI-client ergonomics feedback), which stays open — none of this supersedes it.

**Context that makes these cheap:** the connector went live in the Anthropic directory on 2026-08-04 (<https://claude.ai/directory/connectors/source-library>). The directory stores a pointer to the remote URL, not a tool snapshot — clients call `tools/list` on connect. So **every item below is additive and ships on deploy with no re-publish and no re-review.** Renaming or removing an existing tool is the only breaking move; don't.

## 1. Canonical locus addressing — the highest-value missing feature

There is no way to go from "Metaphysics 1045a8" to a page in any held edition. Scholars think in loci, not scan pages: Bekker for Aristotle, Stephanus for Plato, Diels–Kranz for the Presocratics, Migne for the Fathers, sura:ayah, chapter:verse.

The reporter found the passage they needed by *noticing* that Bekker vol. 2's scan pages map 1:1 onto Bekker numbers, deriving `page ≈ 310 + (Bekker − 1094)`, and guessing.

**The numbers are already extracted.** They sit in the OCR as `<page-num>1103</page-num>`. Proposed:

- `get_quote(book_id, locus="Bekker 1460a26")`
- `get_by_locus(work, reference)` → the matching page across **every** held edition (Greek, Latin, each English translation)

That second form collapses 5–6 calls into 1 and is, in the reporter's words, "the feature that would change what the library is for."

## 2. `contains_works` derived from OCR running headers

Four volumes in a row misled the session because the catalogue title named works the scan does not contain (details in the companion issue). The running headers in the OCR were perfectly reliable throughout.

Proposed field, computed mechanically from header frequency and exposed in `get_book` and `list_books`:

```
contains_works: [{ work, first_page, last_page }]
```

[...]

Issue #3652, opened 2026-08-06
Aristotle Opera sets: titles don't match contents; translation_percent inconsistent corpus-wide (from MCP feedback)


JDerekLomas:
Filed from four MCP `submit_feedback` reports (2026-08-05, `SourceLibrary-MCP/4.3`) written during an agent session verifying ~25 circulating Aristotle quotations against the corpus. Every claim below was re-checked against production Mongo before filing; where the report was wrong or imprecise, that is noted.

## A. Multi-volume *Opera* sets: titles do not describe contents

Confirmed titles/slugs as of 2026-08-05:

| id | catalogued as | reported actual contents |
|---|---|---|
| `6956953e8c9559f6c2db0b6d` | "The Rhetoric, Poetic, and Nicomachean Ethics of Aristotle" (1818) | only the Nicomachean Ethics, Books I–X |
| `69ae633c5d11d232640c382c` | "Aristotelis Opera (Bekker Edition, Vol. 2): Metaphysica, Ethica, Politica, Rhetorica, Poetica" (1831) | the **Scholia** — Porphyry's *Isagoge*, David the Invincible's *Prolegomena*, Simplicius/Dexippus on the *Categories*. i.e. Bekker vol. **4** |
| `69b220b356715b0e324732e1` | "Aristotle, Opera Vol. 6 — Metaphysica (Aldine Editio Princeps)" (1495) | Nicomachean Ethics from p.10, Politics from p.200+ (headers ΗΘΙΚΩΝ ΝΙΚΟΜΑΧΕΙΩΝ, Politics VIII) |
| `69b21bc3429e087c6f863493` | "Aristotelis Opera, Vol. 5 (Rhetoric, Poetics, Politics)" (1495) | Problems, Mechanics, and the **Metaphysics** (Book Ζ at p.469, Θ at p.481) |

The pattern is volume-number/contents misalignment across ingested *Opera* sets, so these should be audited as a batch, not one at a time.

**Correction to the report:** it claimed `69b21bbc429e087c6f8632bc` is *titled* "Enquiry into Plants" while its slug says Vol. 4. The `title` and `slug` actually agree ("Aristotelis Opera, Vol. 4 (Metaphysics, Ethics)"). The mismatch is in **`english_title`, which reads "Enquiry into Plants"** — a Theophrastus work, on an Aristotle *Opera* volume. Same class of bug, different field. Worth sweeping `english_title` on multi-work volumes too.

[...]

PR #3654, opened 2026-08-06
MCP directory guard, page-break continuity, and the BPH records that had no address

JDerekLomas:
Two things, both prompted by the connector going live in the Anthropic directory on 2026-08-04 (<https://claude.ai/directory/connectors/source-library>).

---

# 1. Page breaks — the retrieval unit doesn't match the writing unit

The corpus is fetched a page at a time. It was never *written* a page at a time.

**Measured over 1,209 adjacent prose page-pairs: 18.6% have a passage spanning the break**, 1.7% with a word split by a hyphen — *"…the guide of our move-"* / *"movements, skill in…"*.

**Access was never the problem.** `/api/books/[id]/quote` has accepted `include_context` since long before the connector shipped, and `get_book_text` reads ranges. The problem is that a caller cannot *tell*. A page ending "our move-" reads as ordinary prose to a model that never saw the next leaf, and it arrives wrapped in a perfectly valid citation — so a fragment gets quoted with every appearance of integrity. That is the same family as the wrapper-block misquote in `quote-and-snippet-integrity.md`: confidently citing something that isn't the whole of what the author wrote.

So `get_quote` now returns:

- **`continuity`** — `continues_from_previous` / `continues_on_next` / `hyphen_split_at_end`, computed from text already in hand. No extra query, no model call.
- **`continuity_hint`** — one line telling the model what to do about it.
- **`context: true`** — opt-in, finally passing the parameter the API was always waiting for. Opt-in rather than default because 18.6% is not 100%, and it triples the payload of the most-called citation tool.

Two things the predicate deliberately refuses to do: it makes **no case-based claim about caseless scripts** (Arabic, Hebrew, CJK get the hyphen signal only — a confident wrong answer is worse than silence, and a flag that cries wolf gets ignored), and it declines anything under 200 characters, which is plate captions rather than prose.

### Two corrections worth recording

**The first measurement said 12.6%, and it flattered the status quo.** It judged a page's opening without stripping the running head, so every page whose prose resumed under a `BOOK ONE` line scored as a clean start. Most pages in most scans carry a running head — that was the common case, not an edge one. Which is precisely the mistake this module exists to stop a caller from making.

**Verifying on a preview instead of trusting a green suite caught two bugs, one of which no unit test could have caught.** The API serves `quote.translation`; the code read `quote.text`, which isn't a field, so every flag came back false — silently. And production serves that text through `markForExport`, which threads **zero-width characters** through it as the provenance mark. They're invisible, so a page ending `"our move-"` arrives ending in a zero-width joiner and `/[\p{L}]-$/` sees a clean break. The tests would have stayed green forever: they run on pre-mark text straight out of Mongo. A test-versus-production gap where the fixture is real but captured *upstream of the transform that breaks the thing under test*.

### Verified live on this PR's preview

```
p46: continues_on_next ✓  hyphen_split_at_end ✓   → "continues on p.47"
p47: continues_from_previous ✓  continues_on_next ✓ → "runs from p.46 through p.48"
p61: continues_from_previous ✓  (the running-head case)
context:true → next_page opens "movements, skill in intentions…"
```

Fixtures are real full pages captured from production **before** the assertions were written. An earlier draft paraphrased them short, three tests failed on the 200-char floor, and the tempting fix was to lower the floor to fit invented evidence. Captured the real pages instead. Every guarded behaviour confirmed red when its code is removed.

---

# 2. The directory contract guard

The connector was listed in late April 2026, **silently de-listed in mid-May with no notification**, and republished only in August. Two defects were fixed in between — tool titles only under the deprecated `annotations.title` (#2618), and `.well-known/*` advertising OAuth that `initialize` never enforced (#2621) — and **nothing in this repo stopped either from coming back**. Both are invisible from inside the codebase; the app works perfectly with either present.

`scripts/audit/mcp-directory-contract.mjs` speaks MCP to a live server and asserts what the validator reads. It also pins tool **names** against a committed manifest, encoding the rule that governs every future change: adding a tool or a parameter ships on deploy with no re-review, because both listings hold a pointer to `/api/mcp` and clients call `tools/list` on connect — but renaming or removing one silently breaks saved Claude Projects.

`npm run audit:mcp:self-test` runs the negative controls, including the exact 2026-05 shape where `title` is absent but `annotations.title` looks fine. No network, so it blocks every PR; the live audit runs nightly. **A preview URL is a valid target** — this PR's own preview passes 14/14, which is how the `context` parameter was confirmed not to break the contract.

---

## Also in scope

- `.claude/docs/mcp-registry-publish.md` — described one listing we control. There are two, they drift, and only the registry one is ours; the Anthropic listing isn't org-linked, so its copy changes by email.
- `server.json` 4.3.2 → 4.4.0, matching the route. The audit now pins it.
- `memory/mcp-server.md` — described the remote server as "with OAuth" (wrong since June) and conflated it with the npm package, whose tool list has since diverged.

## Out of scope

- Republishing to the MCP registry (needs `mcp-publisher` + a GitHub login).
- Linking the directory listing to a Claude org — needs a Team/Enterprise org ID; recorded as Q11 in the Poundwise brief.
- The rest of the Aristotle feedback: #3652 (catalogue data quality) and #3653 (API shape). This PR closes item 4 of #3653 only.

`npx tsc --noEmit` clean. 1,897 unit tests pass.


JDerekLomas:
## Added: the BPH manuscript/photograph records had no URL at all

Reported twice by BPH staff and never picked up — José Bouman 2026-07-31 ("not possible to click on titles with a shelf mark M (+number), nor on those with shelfmark Fot (+ number)") and Natalie Koch 2026-08-05 ("the manuscript records aren't clickable yet").

Memorix issues no UBN for manuscripts or photographs. The catalogue detail route is keyed on `ubn` and the browser's `detailUrl()` returned null without one, so those rows rendered as dead plain text. Not a styling bug — **the records had no address anywhere on the site.**

**Measured: 2,012 rows carry no `ubn`, including all 110 `Fot` photographs and 442 `M ` manuscripts.** Every one has a `uuid`, so the route now accepts either key and picks the column by shape. Safe because the shapes cannot collide: `ubn LIKE '%-%-%-%-%'` returns 0 rows in production.

Second defect found on the way: manuscripts keep their title in `full_title` (`title` is null on all 812), and the API selected neither `uuid` nor `full_title` — so even once clickable they'd have read "(untitled)".

Typing `ubn` as nullable surfaced four more places that had quietly assumed it: a `digitizedUbns` index, the public path, the editor nav, and the edit link.

The key rule lives in `src/lib/bph-catalog-key.ts` rather than inline, so the collision property is pinned by a test — if a future scheme introduces hyphenated UBNs, that fails loudly instead of silently 404ing every affected record. Negative control confirmed.

### Verified on this PR's preview

```
manuscript (uuid)      HTTP 200 | soft-404: no | Miroir de Salut Du Nom de Jésus - BPH catalogue
print (ubn, control)   HTTP 200 | soft-404: no | De manifesten der Rozenkruisers…
list row "M 341"       ubn:null uuid:5f8c42fc-… full_title:"Miroir de Salut Du Nom de Jésus"
```

Soft-404s return 200 with a not-found body here, so the check greps the body rather than trusting the status code.

PR #3656, opened 2026-08-06
eval: corpus dataset + OCR trustworthiness instruments

JDerekLomas:
Flat tables for the double-OCR corpus, and the instruments that came out of
analysing them.

## The dataset

`scripts/eval/build-corpus-dataset.mjs` — resumable, free (Mongo reads only).

| table | rows |
|---|---|
| `books.csv` | 2,135 |
| `pages.csv` | 779,409 |
| `revisions.csv` | 191,221 |
| `book_terms.csv` (TF-IDF) | 83,776 |

It adds the three things the existing pair-level export lacks: book metadata,
a **within-book control** (pages OCR'd once, so claims about the re-OCR'd
population are falsifiable), and **re-OCR provenance**.

Provenance is the point. 29.5% of OCR revisions carry
`source='shift-repair-erara-2026-07'` — #3357 moving text between pages, where
the "prior" side is the neighbouring page's transcription. Removing those by
their own label drops the residual leaf-shift rate from the ~40% that
`revision-image-shift.mjs` reports on the raw corpus to **4.2%**, leaving 69%
as clean same-image re-OCR. That test abstains on the 49.9% of pairs with no
readable printed page number, so 4.2% is a lower bound.

## Instruments

- `corpus-dataset-report.mjs` — offline; reproduces every figure from the CSVs.
- `corpus-signal-audit.mjs` — scores ten disagreement signals under two targets.
- `ocr-self-agreement.mjs` — cross-model containment.
- `typeface-divergence-artifact.mjs` — three model tiers, per-page, image URLs.
- `ocr-quality-screen.mjs` — free corpus-wide screen + blinded stratified
  sampler, emitting tasks for the existing volunteer page-check queue.
- `tibetan-split-pilot.mjs` — splitting multi-leaf scans.

## Findings

**Re-OCR is not uniformly an improvement.** It repairs commentary-as-transcription
(589 fixed / 303 introduced) but introduces more repetition loops than it fixes
(352 / 614). Overwhelmingly Tibetan and CJK on `flash-lite-preview`: Tibetan ×
lite loops at 25.1%, and is 341 of all 614 newly-broken pages. Excluding Tibetan
the net effect is +45 across 99,223.

**The failure axis is letterform familiarity, not language.** Within Hebrew,
same model and prompt: printed square script 96-98%, Mikraot Gedolot (square
text + Rashi commentary on one page) 28%, cursive manuscript 9%, Zohar 0% —
while Arabic reads at 80% and Sanskrit at 65%. This reproduces `/blog/rashi-ocr`
from an independent instrument, and covers the case that post flagged as
unsolved: semantic alignment passed it, because OCR and translation were
generated from the same hallucination.

**`gemini-3.1-pro-preview` fails too**, so this is not a routing fix. The three
tiers fail differently — lite fabricates fluently, flash truncates, pro leaks
its reasoning into the transcription.

## Method notes

Three of my own metrics produced confident wrong answers before the data
corrected them, each documented in the commit that fixed it: Jaccard labelled
German "NOT READING" at 31% when containment was 93% (one model simply
transcribed further); whitespace tokenizing discarded every Chinese page as
"thin"; and `body_emptied`, the strongest independent signal, turned out to
fire mostly on **repairs** — Micrographia p289's prior text was Mersenne's
*Harmonicorum Libri*, a different book entirely.

Consequently the review queue's claim is narrow: it finds pages where two
passes disagree about what the page **is**, not which side is right. Its job is
to build a labelled set. Every precision figure here rests on proxies that have
now been shown to point the wrong way, which is why the volunteer sampler is
blinded and equal-per-decile with known-good controls mixed in.

No production code paths touched — all additions under `scripts/eval/` plus one
doc.


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek** — ships a public blog post (`reading-or-generating`) and edits the blog index, which is public copy. The six eval scripts and the flat corpus dataset underneath are unlanded and uncontested, but the PR also commits 23.7K lines of CSV into the repo, which is a repo-weight call. Nothing on main supersedes it.

JDerekLomas:
Closing (backlog decision 2026-09-22): superseded by the #4916 dataset work (portal with datasheets at ocr-quality-paper.vercel.app), and committing 23.7K lines of CSV into the repo is not the shape we want. The six eval scripts are recoverable from the branch if any is still wanted. Branch kept.

Issue #3657, opened 2026-08-06
Analytics dashboard shows two differently-deduped traffic figures without labelling either


JDerekLomas:
## Summary

The `/analytics` page carries two traffic instruments that disagree by **5.9×** over the same window, and the one that looks authoritative is labelled `human` on **99.999%** of rows by construction rather than by measurement.

Reported as "the analytics numbers look wrong." They are at least inconsistent, and the label that would tell you which to trust cannot.

## Measured, last 7 days

| source | count |
|---|---|
| `analytics_pageviews` (written by `/api/track`) | **104,904** |
| `analytics_events` `event: page_read` (written by `/api/analytics/event`) | **615,358** |
| — of those, `traffic_class: human` | **615,354** |
| — of those, any bot class | **0** |

`traffic_class` is 100% populated in this window (classification landed 2026-07-29), so this is not a coverage gap.

### 1. The `human` label is tautological

Both routes are **client-side JS beacons**. Anything that reaches them ran JavaScript, so it is classified human — the classifier restates its own entry condition. `src/app/api/track/route.ts:182` makes this explicit:

```ts
recordClass(db, host, 'human'),   // hardcoded
```

and `src/proxy.ts:873` states the assumption: bots "run no JavaScript."

That assumption is **defensible but known to fail exactly where it matters**. This project's own record: the AS132203 fleet used 28 forged Chrome UAs, and PostHog once showed 204,270 one-hit distinct_ids with an impossible 50/50 Chrome-Windows/Chrome-Mac split. Headless fleets execute JavaScript.

**A classifier that has never returned a non-human value in 7 days is not evidence of clean traffic.** It should either report "not measurable" or be screened against `userAgent`/`ip`, both of which `analytics_pageviews` already stores — so unlike the historical `analytics_events` problem (#3405), this **can** be audited retroactively.

I did audit it and found **no fleet signature**: top /24 is 1.8% of views, top 12 /24s are 20.5%, countries look organic (US 24%, BR 15%, NL 6%, FR 5%, ES 5%). So this is a latent instrument weakness, not evidence of live contamination.

### 2. The 5.9× divergence — a denominator question

Most likely `page_read` fires per page-turn in the reader while `analytics_pageviews` fires per navigation. If so the two are measuring different units under similar names, and any tab quoting either without saying which is misleading. **This needs pinning before either number is quoted** — CLAUDE.md: *a metric's name is a claim about its denominator.*

[...]

PR #3664, opened 2026-08-06
docs: measurement invariant was backwards (pointed readers at the unclassified collection)

JDerekLomas:
Documentation only. Zero risk — no code path changes.

## The defect

`measurement-instruments.md` told the reader:

> `analytics_pageviews` classified at ingestion and stayed clean; `analytics_events` did not … and **could never be cleaned**, because the field needed to classify was not stored.

Measured 2026-08-06, that is **backwards**:

| collection | `traffic_class` today | backfillable? |
|---|---|---|
| `analytics_events` | **100%** since 2026-07-29 (0% before — the fix landed mid-window) | n/a, fixed forward |
| `analytics_pageviews` | **0%**, on every day measured including today | **yes** — stores `userAgent` + `ip` |

So the doc was pointing the next reader at the collection with **no class on it at all**, and away from the one that had been fixed. That is the worst possible failure for a doc whose whole job is "which instrument can you trust."

Corrected in place, with the measurement shown, and an explicit warning that **this bullet has now been wrong once in each direction** — re-measure rather than trusting either sentence. The July incident text is preserved; it is still valid history.

## Two invariants added, found while checking it (#3657)

**A classifier that never returns a negative is not classifying.** `page_read` labels 615,354 of 615,358 events `human` over 7 days — zero bots. That is not a clean corpus: both `/api/track` and `/api/analytics/event` are **client-side JS beacons**, so anything reaching them ran JavaScript and is called human *by construction*. `src/app/api/track/route.ts` hardcodes `recordClass(db, host, 'human')` for exactly that reason and `src/proxy.ts` states the assumption. It is defensible, and it fails precisely where it matters — the AS132203 fleet forged 28 Chrome UAs, and headless fleets run JavaScript. Read the label as *"reached a JS beacon"*, never as *"screened"*.

**Two instruments that sound alike are not comparable until someone states the unit.** `analytics_pageviews` (104,904/7d) vs `analytics_events.page_read` (615,358/7d) differ **5.9×**:

| | pageviews | page_read |
|---|---|---|
| mounted | **root layout** — every route site-wide | reader only |
| fires on | `pathname` | `page.id` |
| debounce | 300ms, **cancels** | none |
| excludes | `/admin/`, `/api/` | nothing |

Scope alone predicts pageviews should be the **larger** number. It is 6× smaller — which proves `page_read` fires more than once per page actually read, without needing to know the mechanism. `analytics-ingest.ts:143` already documents the double-fire; the dashboard does not carry the caveat to the reader.

## Deliberately not in this PR

The stale `system_config.dashboard_snapshot` fallback (126 days old, `usage/route.ts:143`, no staleness indicator). It is a code change and needs a design call — surface the date to the UI, or drop the fallback entirely — so it gets its own PR rather than riding along with a docs fix.


JDerekLomas:
Closing without merging — **the doc was right and this PR is wrong.**

## What I got wrong

I read the absence of a `traffic_class` field on `analytics_pageviews` rows as the absence of classification. It is the opposite: it is the **signature of a filter**.

`src/app/api/track/route.ts:134-146` classifies with the full Cloudflare signal set (`cf-verified-bot`, `cf-threat-score`, per-IP rate cap) and then:

```ts
if (cls !== 'human') {
  deferDbWrite(async () => { await recordClass(await getDb(), host, cls); });
  return OK;                 // <- no pageview row is written at all
}
```

Non-human traffic is **dropped before the insert** and counted in `analytics_traffic_class`. Every stored pageview is human *by filtering*, so a per-row label would be redundant. `recordClass(db, host, 'human')` further down is inside the human-only branch, so it is correct too — not a hardcode bug.

Verified against the counters, last 7 days:

| collection | human | other_bot | search_crawler | ai_agent |
|---|---|---|---|---|
| `analytics_traffic_class` (pageviews) | 116,858 | 485 | 83 | 1 |
| `analytics_events_class` (dropped events) | — | 42 | 55 | 1 |

`/api/analytics/event` does the same thing via `recordDroppedEvent`. So the `page_read` stream is also a filtered human stream.

## Which makes both "invariants" in this PR false

- **"A classifier that never returns a negative is not classifying"** — it does return negatives. They are routed to the two `*_class` counters and dropped from the main collections. 99.999% human is what a working filter looks like from inside the filtered set.
- **"The doc is backwards"** — it is not. *"`analytics_pageviews` classified at ingestion and stayed clean"* is precisely accurate: it classifies at ingestion, and keeps only the clean rows.

Closing per the precedent in `measurement-instruments.md` itself — *a plausible non-fix is worse than no fix* (#3418 was closed the same way).

## What survives, and is now in #3657

- The **5.9× divergence** between `analytics_pageviews` and `page_read` is real, and the trigger asymmetry behind it is real: `PageTracker` is root-layout/site-wide, fires on `pathname`, 300ms debounce that cancels, skips `/admin/`; `page_read` is reader-only, fires on `page.id`, no debounce, no exclusion. Scope predicts pageviews should be larger; it is 6× smaller. **`page_read` over-counts.**
- `system_config.dashboard_snapshot` is **126 days stale** and is an unlabelled fallback in `usage/route.ts:143`.
- `analytics_events` rows gained `traffic_class` on 2026-07-29 (0% before, 100% after). Worth one line in the doc eventually — but as an addition, not a correction.

## The actual lesson

Four times in this audit I read a missing field, a missing email, or a missing config line as a missing behaviour, when the behaviour was implemented somewhere I had not looked. **Absence of a marker is not absence of the mechanism** — find the code path before concluding from the shape of the data.

Issue #3661, opened 2026-08-06
Canonical loci (Bekker/Stephanus): what is actually derivable — scoped from measurement


JDerekLomas:
## Why

Canonical loci — Bekker for Aristotle, Stephanus for Plato — are how scholarship actually addresses these texts. They were agreed centuries ago precisely so a citation survives re-typesetting. Source Library addresses everything by scan page, which is a property of one copy and shareable with nobody.

The gap surfaced through the MCP connector (#3653 item 2), where an agent verifying ~25 attributed Aristotle quotes had to reconstruct the Bekker mapping by hand and then guess: *"I found the Poetics passage by noticing Bekker vol. 2's scan pages map 1:1 onto Bekker numbers, deriving page ≈ 310 + (Bekker − 1094), and guessing."*

This issue scopes what is actually derivable from data we already hold, **before** anyone commits to building it.

## Correction to an earlier claim

I previously said roughly two-thirds of Aristotle/Plato books were "one derivation away." **That was wrong.** It came from checking a single volume by hand and generalising from a count of books carrying `<page-num>` at all.

Measured properly across all 276 live Aristotle and Plato books (109 + 167), parsing every `<page-num>` anchor and fitting scan-page → printed-number:

| | Aristotle | Plato |
|---|---|---|
| Live books | 109 | 167 |
| With ≥10 parseable anchors | 87 | 144 |
| **A — clean linear fit (pagination)** | 15 | 45 |
| **B — marginal canonical refs** | 3 | 2 |
| Partial fit (0.7–0.9) | 3 | 12 |
| No usable structure | 66 | 85 |

So ~22% have a clean fit, not two-thirds, and only a handful of those are canonical.

## The measurement changed the design

The first pass asked one question — does scan→printed fit a line? — and that **conflated two different mechanisms and scored the more valuable one worst**:

**A. `<page-num>` holds the book's own pagination.** Useful as a locus only when that edition's pagination *is* the citation standard. Needs a clean linear fit.

**B. `<page-num>` holds marginal canonical references** printed beside the text — the Oxford/OCT convention. These *are* loci already, need no fitting, and score **worst** on a linear test because Bekker numbers do not advance at the rate of pages. Historia Animalium scored `fit=0.009` and is one of the best books in the corpus for this purpose.

## The pilot set — 7 books

Two of them are the editions that *define* the systems:

| id | | fit / letters | why it matters |
|---|---|---|---|
[...]

Issue #50, opened 2026-02-20
Processing outage Feb 18-19: 60k image fetch failures, pipeline degraded 12+ hours


JDerekLomas:
## Summary

Between Feb 18 20:00 UTC and Feb 19 ~14:00 UTC, the processing pipeline was severely degraded. ~60,000 OCR attempts failed with "Failed to fetch image: fetch failed", while successful processing dropped to near-zero for ~12 hours. No user-facing data was lost (verified: 0 pages lost existing OCR), but significant processing time was wasted and API quota was burned on failed requests.

## Timeline

| Time (UTC) | Success/hr | Failed/hr | Notes |
|---|---|---|---|
| Feb 18 20:00 | 0 | 5,332 | Failures begin — zero successes |
| Feb 18 21:00 - 23:00 | 0-1 | ~8,000 | Total outage — all attempts fail |
| Feb 19 00:00 - 01:00 | ~1,800 | ~8,000 | Partial recovery starts |
| Feb 19 02:00 - 08:00 | 18-90 | ~8,500 | Mostly dead — <100 successes/hr |
| Feb 19 09:00 - 11:00 | 300-1,600 | ~8,500 | Slow ramp-up |
| Feb 19 12:00 - 13:00 | 13-18k | 7-8k | Major recovery begins |
| Feb 19 14:00 | 22,171 | 1,402 | Approaching normal |
| Feb 19 17:00+ | 17-22k | <40 | Fully recovered |
| Feb 20 01:00 - 04:00 | 24-30k | <70 | Peak throughput (overnight batch) |

## Root Cause

**59,875 OCR failures** with error `"Failed to fetch image: fetch failed"` — the bulk of the 62k total failures. The OCR pipeline couldn't download page images from their source URLs.

Likely causes (needs deeper investigation):
1. **External image source downtime** — IA/Gallica/MDZ servers may have been rate-limiting or down
2. **Vercel Blob outage** — archived images stored on Vercel Blob may have been unreachable
3. **DNS/network issues** — the processing scripts (on Mac + Hetzner) may have had network problems

The failed records have `source: undefined` and `error_type: undefined` — **the logging is incomplete**, making root cause analysis harder.

### 1. Stale `realtime_ocr` jobs stuck in `processing`

4 jobs with status `processing` and 0 progress:
- 3x limit=5000, 1x limit=2000, 1x limit=10000
[...]

Issue #51, opened 2026-02-20
Create missing MongoDB indexes on analytics collections


JDerekLomas:
## Problem

The analytics dashboard (`/analytics`) is unusable — the Usage tab triggers 504 timeouts because `gemini_usage` (1.4M+ docs) has no standalone `timestamp` index. Time-range queries (`{ timestamp: { $gte } }`) do full collection scans.

Three other collections also lack basic indexes: `cron_runs`, `pipeline_snapshots`, `loading_metrics`.

## Current index state (production)

```
gemini_usage (3 indexes):
  _id_
  gemini_usage_book_ts_idx: { book_id: 1, timestamp: 1 }     ← compound, doesn't help time-only scans
  gemini_usage_type_status_book_idx: { type: 1, status: 1, book_id: 1 }

cron_runs (1 index):
  _id_                                                         ← no timestamp index at all

pipeline_snapshots (1 index):
  _id_                                                         ← no timestamp index at all

loading_metrics (2 indexes):
  _id_
  loading_metrics_ts_idx: { timestamp: 1 }                     ← has one, needs received_at + name compound
```

### Critical (gemini_usage — fixes 504s)
1. `gemini_usage_ts_idx`: `{ timestamp: -1 }` — enables all time-range analytics queries
2. `gemini_usage_status_ts_idx`: `{ status: 1, timestamp: -1 }` — failed-call aggregation

### Important (cron_runs — pipeline observability)
3. `cron_runs_ts_idx`: `{ timestamp: -1 }`
4. `cron_runs_cron_ts_idx`: `{ cron: 1, timestamp: -1 }`
5. `cron_runs_status_ts_idx`: `{ status: 1, timestamp: -1 }`

### Important (pipeline_snapshots)
6. `pipeline_snapshots_ts_idx`: `{ timestamp: -1 }`

### Nice-to-have (loading_metrics)
7. `loading_metrics_received_at_idx`: `{ received_at: -1 }`
8. `loading_metrics_name_ts_idx`: `{ name: 1, received_at: -1 }`

## What's already done

- `src/app/api/admin/ensure-indexes/route.ts` updated with indexes #1 and #2 (deployed)
- `scripts/create-critical-indexes.mjs` created with all 8 indexes
- The ensure-indexes API route 504s when trying to build on 1.4M docs (Vercel 60s limit)
- Local script also blocks for 30+ min on gemini_usage — Atlas M10 index builds are slow

[...]

PR #3680, opened 2026-08-07
fix(mcp): search ranking, continuity flags, and tool discoverability

JDerekLomas:
Five reports arrived through the MCP connector in one day from a reader
verifying ~25 circulating "Aristotle quotes" against primary sources. This
addresses them. Each finding was measured before and after; two of their
diagnoses turned out to be wrong in instructive ways, and one of mine did too.

## 1. `search_within_book` returned pages in ascending page order

The Atlas branch re-sorted Atlas's results by `page_number` and then cut at 50,
so on a long book the cap was spent inside the front matter before the body was
reached. It read as a ranking problem; it was a retrieval problem. The same fix
went into the regex fallback earlier this week and **missed this branch, which
is the one that runs**.

Measured on all five of the reporter's regression cases: **Atlas already ranked
the wanted page #1 in every one.** Removing the sort is the whole fix.

Worth recording what did *not* work. A local BM25 rescoring pass looked
obviously right — the reporter had diagnosed "raw term frequency, no stopwords",
which is accurate about the displayed `match_count` — and it moved Taylor's
p264 from #1 to #366, because Taylor's 1801 wording contains none of the query's
words and Atlas was using corpus-wide evidence a single book's pool cannot see.
The scorer is kept for the regex fallback, which has no relevance signal at all.

| case | before | after |
|---|---|---|
| Aldine 1551 → p705 (`Poetics` 1460a26, verbatim) | last of 8 | **#1** |
| Diogenes Laertius → p159 | ~#50 of 59 | **#1** |
| Taylor → p264 | absent | **#1** |
| Taylor "Speusippus" → p347 | #1 | #1 |
| nonsense control | 0 | 0 |

## 2. The two legs concatenated rather than merged

Confirmed against the live API: scoped to Sextus Empiricus, results 52–59 are
pages 396, 397, 209, 350 … at `match_count` 1, and p397 is exactly what
`search_concept` returns for the same query. **The semantic leg does run** — its
hits just sat behind fifty lexical ones. Each leg is now normalised against its
own best hit (the only honest way to compare an Atlas BM25 score with a cosine
similarity) and a page found by both is marked `found_by: "both"`. `score` is
surfaced so a caller can tell a real hit at rank 20 from noise.

## 3. `hyphen_split_at_end` could never fire

Computed from the **translation** — and a translator resolves hyphens. The
line-break hyphen is a fact about the printed original, so the flag was being
tested against the one text where it cannot appear. Diogenes p.33 (Greek ends
`ἐγέ-`, English ends "the Greeks.") now reads hyphen `true` / `continues_on_next`
`true`, where before it read false and false — a page that looked self-contained
and was not.

Two of their other continuity cases were already correct: Taylor p268 was never
broken (they predicted it rather than measuring), and p269 was fixed by the
tag-decapitation repair earlier in this branch.

## 4. Front matter was invisible on early modern books

Both heuristics keyed on roman-numeral pagination and English structural
headers — conventions of Anglo-American printing. The 1551 Aldine (Greek and
Latin front matter, foliated by signature) flagged **nothing**, which is exactly
the material where the apparatus is longest. Structural `<page-type>` values are
now a third, subordinate signal.

| book | flagged before | after |
|---|---|---|
| Aldine 1551 | 0 | 45 of 720 |
| Taylor 1801 | 67 | 66 of 536 |
| Bekker 1831 (*has no front matter*) | 11 | 11 of 684 |

The **prose** page-types (`preface`, `dedication`, `advertisement`) are
deliberately excluded — a body page mistyped `preface` would otherwise be
demoted on no other evidence. A test pins that exclusion.

## 5. A commentary on Aristotle outranked Aristotle

Searching "Aristotle Metaphysics" put Bekker's 1831 Greek critical edition 7th
and Taylor 9th, behind commentaries by Aquinas, Asclepius and Tartaretus. All
carry both query words in their titles, so the word-count rung tied — and the
tie fell through to "older editions rank higher", handing the top of the page to
whichever commentary was oldest. That rule is right *between editions of one
work* and meaningless *across works*.

Adds one rung: does the query name this book's author? It sits **above** the
exact-phrase rung, because Asclepius's volume is literally titled "…Commentary
on Aristotle Metaphysics" and won that rung by coincidence.

Replayed against production's own result sets for eight queries: "Aristotle
Metaphysics" top five become five Aristotle editions; "Aquinas" surfaces three
actual Aquinas works; "Newton Principia" puts Newton above Descartes; five other
queries unchanged.

**Known limitation, pinned by a test so it is not read as a regression later:** a
commentary naming both the author and the work in its title ("Averroes'
Paraphrase of Plato's Republic") clears the word-count rung above this one, so
"Plato Republic" still returns Proclus at #2. Fixing that needs the
work/commentary relation in the data rather than inferred from title strings.

## 6. `search_concept` was invisible from inside a session

Ranked by the reporter above every bug they filed:

> I spent a full working day on a quote-verification task which is exactly and
> only what semantic search is for. I never called `search_concept` once. Not
> because it failed, but because nothing in the MCP surface told me it was the
> right instrument.

`search_within_book`'s description now says what it is for and routes paraphrase
queries to `search_concept` explicitly. `llms.txt` is exposed as an **MCP
resource**, so an agent can read the corpus guide mid-session instead of needing
to already be browsing the website.

## Out of scope, deliberately

- **Locus addressing** (`get_by_locus("Pol. 1287a28")`) — their highest-value
  feature request, tracked in #3661.
- **Span-level `text_type`** separating author text from a translator's
  commentary. Real and important — an agent can currently quote a Victorian
  translator as Aristotle — but it is a data-modelling job, not a ranking fix.
- **Coverage map and lemma indexing.** Both worth doing; neither is a bug fix.
- **Malformed `<margin>` markup.** The trailing-furniture regex absorbs it by
  accident, so it is currently harmless; a validator pass over
  `<margin>`/`<note>`/`<unclear>` is owed.

## Verification

`npx tsc --noEmit` clean; 2019 tests across 151 files pass; the
directory-contract audit passes 14/14 including its negative controls. Per
`.claude/docs/invariants/search-filters-and-lanes.md`, the ranking change wants
a curl matrix against the preview before merge — the offline replay above used
production's own result sets, which exercises the comparator but not the lanes.


JDerekLomas:
## Preview verification (curl matrix, prod vs preview)

Per `.claude/docs/invariants/search-filters-and-lanes.md`, ranking changes need a
curl matrix against a deployed preview rather than an offline replay. Run against
`sourcelibrary-v2-316jsdgwg` with a browser UA (a default UA trips the bot limiter).

### 1. Library ranking — does the work outrank commentaries on it?

| query | production | preview |
|---|---|---|
| Aristotle Metaphysics | Aristotle \| **Asclepius** \| Aristotle \| **Aquinas** | Aristotle \| Aristotle \| Aristotle \| Aristotle |
| Newton Principia | Newton \| **Descartes** \| Newton \| Newton | Newton \| Newton \| Newton \| Newton |
| Aquinas | …\| …\| Renwick \| Thomas Aquinas (#4) | …\| …\| **Thomas Aquinas (#3)** \| Pseudo-Thomas Aquinas |

### 2. `search_within_book` — the reporter's regression cases

| book | wanted page | production | preview |
|---|---|---|---|
| Taylor *Metaphysics* | p264 | **absent** | **#1** |
| Aldine 1551 | p705 | **absent** | **#1** |
| Diogenes Laertius | p159 | #51 of 59 | **#1**, `found_by: "both"` |

`found_by: "both"` on the Diogenes case is the merge working as intended — the
keyword and semantic legs independently agree on that page, which is the
strongest signal the tool can report and was previously unreachable behind fifty
lexical hits.

Two side effects worth noting:

- **`total` is now honest.** It read 50/50 and 49/49 on production because the
  cap was applied during retrieval; it now reports what actually matched
  (400, 394) alongside `returned: 50`. A caller that saw 50 and assumed that was
  everything was stopping too early.
- **`score` is exposed** (1 → 0.06 on the Aldine case), so a caller can tell a
  real hit at rank 20 from noise the way `search_concept`'s calibration already
  allows.

### 3. Controls

Nonsense query returns `total: 0` on both, so matching still runs and the
improvement is not from loosening the filter. Front matter orders last in every
case on both sides.

### CI

`test`, `search-eval`, `Negative controls` and `DCO` pass. The Vercel check shows
a failure from an earlier build in the series; the deployment for the head commit
of that series reports `state: success` and is the one curled above.

PR #3681, opened 2026-08-07
docs(invariants): CDN path rules beat force-dynamic; referrers can't attribute in-site clicks

JDerekLomas:
Two failure classes from the `/give` session (#3646 / #3649 / #3668). Both shipped silently, and both read as facts about users when they were facts about instruments.

## deploy-and-caching

`CDN-Cache-Control` in `next.config.ts` is keyed on **path** and overrides a route's `force-dynamic`. `/support` rendered per request and was edge-cached anyway — production answered `cf-cache-status: HIT` with `age: 1202` on a supposedly dynamic page, so every visitor got whichever country warmed the cache, held for 24h, plus a per-request Mongo call that bought nothing.

Adds:

- The tell — `cache-control: private, no-cache, no-store` and `cdn-cache-control: public, max-age=86400` are present in the **same response**. Reading only the first inverts the truth.
- Personalise on an uncached route; never uncache a route to personalise. Those TTLs are load-bearing (June 2026 scraper flood, ~1.3M req/day at exactly those paths).
- **Last match wins** — the ordering the `/book/:id/preview` `private, no-store` overrides depend on. Reordering them could serve a cached editor 200 to anon users.
- The standing guard (`tests/unit/dynamic-routes-not-edge-cached.test.ts`) and its **10 untriaged pre-existing collisions**.
- Trigger line updated so the doc fires on any `export const dynamic` on a page.

## measurement-instruments

`analytics_pageviews.referrer` cannot distinguish an in-site click from a direct arrival, for two independent reasons that stack: `/api/track` collapses self-referrals to `'direct'`, **and** Next's client-side navigation preserves the original external referrer.

I misread 7 `/give` visits as "none came from the header" when up to 4 of them may have been exactly that.

Adds:

- Fire the event **at the control**, not reconstructed downstream.
- Use `sendBeacon` for anything fired on a control that navigates — a plain `fetch` is cancelled by the navigation itself, so the metric reads zero for a reason unrelated to users.
- Verify a new conversion event end-to-end in production before trusting its zero, and **write down the test timestamp** so the row can be excluded.
- The generalisation: a zero from a field that cannot express the thing is not evidence of absence.

Docs only — no code, no behaviour change. `CLAUDE.md` stays at 202 lines; both lessons are subsystem-triggered, so they belong in `invariants/` rather than the always-loaded body.

### DIFF
--- a/.claude/docs/invariants/deploy-and-caching.md
+++ b/.claude/docs/invariants/deploy-and-caching.md
@@ -2,7 +2,8 @@
 
 *Read this when you touch: `src/app/api/deploy-warm/route.ts`, `scripts/deploy-prod.sh`,
 `.github/workflows/post-deploy-warm.yml`, any `CDN-Cache-Control` header in `next.config.ts`, any
-`revalidatePath` / `revalidateTag` call, or Vercel project settings.*
+`revalidatePath` / `revalidateTag` call, **any `export const dynamic` on a page**, or Vercel project
+settings.*
 
 ## There are two independent caches, and a deploy can empty both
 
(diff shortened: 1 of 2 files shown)


JDerekLomas:
**Triage finding (read-only backlog sweep, 2026-08-28): main is ahead of this PR's version.**

Both target files (\`deploy-and-caching.md\`, \`measurement-instruments.md\`) have moved past what this PR would add. Current \`main\` already includes: the \`export const dynamic\` caveat on the CDN-cache preamble, the emergency Cloudflare-purge curl one-liner, the \`deploy-warm\` curl step, and a whole "Did my merge actually ship?" section this PR's version doesn't have. This branch was built against an older snapshot of those docs; its content reads as already covered (and exceeded) by later edits on main. Recommend closing.

JDerekLomas:

PR #3682, opened 2026-08-07
docs: localized-surfaces invariant (a hardcoded €1,000 reads as one euro in Spanish)

JDerekLomas:
Gnite doc pass from the giving session. One new conditional invariant plus its routing line — no deletions, nothing demoted.

`/es/support/business` (#3673) surfaced three locale rules with nowhere to live: `git grep -il "locale\|hreflang"` over `.claude/docs/invariants/` returned **nothing**.

**1. Never hardcode a formatted number in a shared string table.** `es-ES` uses `.` for thousands and `,` for decimals, so a literal `€1,000` reads to a Spanish reader as **one euro** — and it got as far as review on the page whose entire job is showing what a gift costs. Store numbers, format with `Intl.NumberFormat` per locale, and force `useGrouping: 'always'` because es-ES otherwise drops the separator below 10,000 (`5000 €`) while the prose beside the table writes `5.000 €`.

**2. A layout's `alternates` covers every child route.** `src/app/support/layout.tsx` sets `canonical = '/support'` for everything beneath it, so a child that doesn't declare its own self-deindexes.

**3. The twin pattern's two forgotten steps** — `sitemap.ts` enumerates static routes by hand (a new `/es/*` page is otherwise invisible), and internal links must branch on the locale you already have, never on `usePathname()`.

Also records why Dutch tax terms (`BV`, `box 2`, `periodieke schenking`) stay untranslated: they're the words the reader's accountant will say back to them.

CLAUDE.md stays at 203 lines, inside the ~250 budget, so nothing needed demoting this pass.


JDerekLomas:

PR #3684, opened 2026-08-07
fix(metering): every worker now logs Gemini usage, plus a standing audit

JDerekLomas:
Closes the worker half of the metering gap in #3576, and makes the rest visible instead of silent.

## The three workers

| worker | what was invisible |
|---|---|
| `embed-gemini.mjs` | **~4.9M embeddings** across five tables — the single largest unmetered phase |
| `ocr-correct-grounded.mjs` | grounded second-pass OCR correction |
| `transliterate-greek.mjs` | Greek transliteration |

All three now log via `supabase-usage-logger` (Supabase primary, Mongo fallback) rather than inserting to `gemini_usage` directly.

**Two deliberate details:**

**Logged BEFORE any early return on empty output.** An empty candidate is still a billed call — the same trap `batch-collector` documents for RECITATION blocks, where a job whose every page was blocked recorded $0.00.

**`batchEmbedContents` returns no `usageMetadata`.** So tokens are read if the API ever supplies them, and estimated (chars/4) otherwise — with the method recorded in `prompt_version` as `tokens=estimated(chars/4)`. An unlabelled estimate sitting in a meter is precisely the failure this work is about; a labelled one is honest. A reader can filter on it.

## The standing audit

`scripts/audit/gemini-usage-logging.mjs`. This class of hole is created by **forgetting**, and forgetting produces silence rather than an error — the same shape as the five `entities.books[]` writers and the six search boxes where one logged nothing for four months.

```
Gemini call sites scanned: 50 (pipeline paths)
  logged:   15
  accepted: 1
  UNLOGGED: 34      ← none are workers
```

It **does not fail the build by default** — many call sites are legitimately unmetered, and a detector that cries wolf gets bypassed. `--strict` fails only when a **worker** is unlogged, which is now zero. Accepted omissions require a written reason, so a real hole cannot be grandfathered in unexplained.

## Not in this PR: the other 34

All 34 are API routes and `src/lib` modules. Patching them one at a time would repeat the mistake the R2 key guard's own header warns about — *"40+ scripts with their own `uploadToR2`, and guarding them one at a time is how the same bug shipped twice."*

The leverage point is the boundary: **`src/lib/ai.ts` is the shared client for 26 modules** and exports `performOCR` / `performTranslation` / `generateSummary` / `performTransliteration` / `processPageComplete`. Instrumenting there plus `src/lib/gemini-batch.ts` (8 importers) covers most of the tail in two files.

Worth noting while in there: `src/lib/ai.ts` carries **its own duplicate `MODEL_PRICING` table**, which is exactly the drift `scripts/lib/model-pricing.mjs` was created to end.

## Verification

`node --check` on all three workers · audit runs clean in `--strict` · logging calls verified to have `page` in lexical scope (a syntax check would not catch that).

### DIFF
--- a/scripts/audit/gemini-usage-logging.mjs
+++ b/scripts/audit/gemini-usage-logging.mjs
@@ -1,0 +1,91 @@
+#!/usr/bin/env node
+/**
+ * Standing audit: which Gemini call sites do NOT log usage?
+ *
+ * WHY. Spend metering failed the same way the entity writers and the search
+ * boxes did — a new caller was added and nobody remembered to log it, and
+ * forgetting produces SILENCE, not an error. In Aug 2026 that meant ~4.9M
+ * embeddings were invisible to every cost surface, and total recorded spend
+ * read $7.5K against a corpus that plainly cost far more (#3576).
+ *
+ * This makes the omission visible instead of silent. It does NOT fail the
+ * build: many call sites are legitimately unmetered (one-off analysis scripts,
+ * validate-key probes). The point is that the list is SHORT and REVIEWED, not
+ * that it is empty.
+ *
+ *   node scripts/audit/gemini-usage-logging.mjs            # pipeline paths
+ *   node scripts/audit/gemini-usage-logging.mjs --all      # every caller
+ *   node scripts/audit/gemini-usage-logging.mjs --strict   # exit 1 if a WORKER is unlogged
+ */
+
+import { execSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+
+const ALL = process.argv.includes('--all');
+const STRICT = process.argv.includes('--strict');
+
+const CALL_RE = /generateContent|batchEmbedContents|generativelanguage\.googleapis/;
+const LOG_RE = /logUsage|logUsageAsync|logGeminiUsage|completeBatchUsage|buildUsagePayload|gemini_usage/;
+
+// Paths that run repeatedly at scale. An unlogged caller here is a real hole;
+// elsewhere it is usually a one-off and only worth listing.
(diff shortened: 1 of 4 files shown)


JDerekLomas:
**Triage finding (read-only backlog sweep, 2026-08-28): partially superseded — one of three workers already covered.**

\`scripts/workers/embed-gemini.mjs\` on \`main\` already logs Gemini usage (via #4162 — \`gemini_usage\` rows, referenced in its header comment). \`scripts/workers/ocr-correct-grounded.mjs\` and \`scripts/workers/transliterate-greek.mjs\` on main do **not** have logging, and \`scripts/audit/gemini-usage-logging.mjs\` (the standing audit this PR adds) doesn't exist on main either. So the real remaining scope here is narrower than the PR title suggests: the other two workers + the audit script. Also see the comment on #3574 — it independently needs Gemini-call logging on three page routes (\`detect-split\`/\`modernize\`/\`transliterate\`) that overlaps with this PR's audit script. Worth consolidating into one rebased PR covering the actual gap rather than reviving this one as-is.

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. All three workers now log usage another way — `transliterate-greek.mjs` and `ocr-correct-grounded.mjs` import `supabase-usage-logger` and call `logUsage()` on main (the latter with this PR's own rationale at line 121), and `embed-gemini.mjs` has recorded spend since #4162. The standing audit landed today as `scripts/audit/gemini-usage-perimeter.mjs` (#4658, a5062769); none of these workers is in its baseline of unlogged sites, so they are covered rather than grandfathered. This PR's `gemini-usage-logging.mjs` would be a second auditor over the same ground.

Issue #3522, opened 2026-08-01
The reference set is 1% early-modern and our corpus is not: add ESTC as a source


JDerekLomas:
## The finding

We have been asking a **modern-imprint catalogue** about **early-modern translations**.

The reference set behind every first-translation claim is the LoC MDSConnect 2016 extract: 118,352 English translations, of which **1,230 are pre-1800 imprints — 1.04%**. Our corpus is overwhelmingly early-modern. The instrument and the question do not overlap.

Measured consequences, 2026-08-01:

- Catalogue recall against known priors is **27%** (`scripts/eval/ft-reference-set-recall.mjs`).
- A systematic sample of 14 works from the visible translation queue found **roughly half have a prior English translation the search missed**. So the positive predictive value of `none_found` is about **50%** — a coin flip.

Crucially, **these are not matcher failures.** I scanned the full extract for each miss:

| work | verified prior | in the LoC extract? |
|---|---|---|
| Agrippa, *De incertitudine et vanitate scientiarum* | Sanford, London 1569 | 12 Agrippa rows, **all** *De occulta philosophia*. Absent. |
| Maier, *Lusus Serius* (1616) | John Hall, London 1654 | 44 "Maier/Mayer" rows, **none** the alchemist. Absent. |
| Böhme, *De Signatura Rerum* | Ellistone 1651 | 10 Böhme rows incl. "Works.", "Clavis." Absent. |
| *Rhetorica ad Herennium* | Caplan, Loeb | **0 rows.** |
| ps-Albertus, *De secretis mulierum* | Lemay, SUNY 1992 | 1 hit, an unrelated 1876 novel. Absent. |

Matcher tuning cannot recover any of these. The #3459 work already lifted the MARC 240 ceiling (22% → 27%) and that was the last available gain from matching. **The remaining loss is corpus coverage.**

## The fix

Add the **English Short Title Catalogue** as a reference-set source. ESTC covers every English imprint 1473–1800 — precisely the period where our priors live and where LoC is empty. Every miss above is an ESTC record.

### Access is solved: it is free, open, and JSON

ESTC moved to CERL and the beta is public (no auth, no key):

```
https://datb.cerl.org/estc/_search?query=agrippa&size=100&from=0
Accept: application/json
```

- **487,406 records.**
- Returns `id` (ESTC S-number), `dates`, `language`, `search_titles`, `search_names`, `subjects`, `holdings`, `digital_copy_present`.

[...]

Issue #3459, opened 2026-07-29
The reference-set principle: 'first translation' is an unprovable negative — assert it against a stated catalogue instead


JDerekLomas:
Framing issue. Sets out what `translation_catalogs` is **for**, and why #3455, #3428 and #3425 are subordinate to it. Arrived at with Derek 2026-07-29, working backwards from why the badge system keeps producing indefensible claims.

## The problem: "first translation" is an unprovable negative

"This is the first English translation of X" asserts that **no prior translation exists anywhere** — a universal negative over all catalogues, all languages, all time.

**No search establishes that.** A catalogue query can return *nothing found*, which is a materially weaker statement. We demonstrated this twice this week against our own data:

- Heath's 1908 *Elements* — the standard modern English Euclid — is **absent from `translation_catalogs`**. A query would have reported "no modern translation" for one of the most-translated works in the corpus.
- Copenhaver's *Hermetica* (Cambridge 1992) was **absent until added by hand** on 2026-07-29.

This explains several things previously treated as separate problems:

| Symptom | Cause |
|---|---|
| **63% of Gemini "prior found" claims fabricated** | asked to ground an unbounded negative; nothing to ground on, so it confabulates |
| `evidence_strength` / `independence_score` exist in the schema | correct instinct — measuring *how many independent looks failed*, not truth |
| Tier-1 "prior found" is never a final demote | presence needs one confirmation; absence needs many |
| 1,770 badges on `confirmed_first_legacy` | a flag standing in for a claim nobody could actually establish |

**The asymmetry:**

| Claim | Cost | Certainty |
|---|---|---|
| "A prior exists" | **one hit** | conclusive |
| "No prior exists" | N searches × M catalogues | **never conclusive** |

So the cheap and reliable work is *finding* priors (`not_first`, `first_complete`, `first_modern`, translator credits). The expensive, never-finished work is *defending* firsts — which is what 6,096 badges are currently doing.

## The fix: assert against a stated reference set

Stop claiming absence from the world. Claim absence from a **named, dated, bounded** set:

> *"Not found in the union of LoC, HathiTrust, K10plus, Wikidata and the UNESCO Index Translationum, as searched 2026-07-29."*

That is checkable, falsifiable, and improvable. It is a measurement rather than an assertion.

**This makes `translation_catalogs` the thing that makes the claim sayable at all** — not a lookup table, but the reference set the negative is asserted against.

## The census paper already does this

[...]

PR #3686, opened 2026-08-07
fix(ft): state the search, not the negative — and stop confirming on evidence we call weak (#3459)

JDerekLomas:
Implements the first item in `.claude/docs/first-translation-reference-set.md` §8: *"Badge copy. ~5,839 live claims rest on evidence with ~50% PPV from a set that is 1% early-modern… Swapping to the search statement makes all of them true immediately and is gated on nothing."*

The state layer for this shipped in #3587 and rendered nowhere. This wires it — and fixes a defect in it that would have made the wiring a no-op.

## The defect: `confirmed` meant "we badged it"

`classifyFirstTranslationClaim` gated `confirmed` on `isFirstTranslation()`. That is the **render** rule — first-family verdict, visible, some translated pages — and it says nothing about evidence. Measured over the 5,932 badged + visible books on 2026-08-07:

| | |
|---|---|
| `confirmed` under the old gate | **5,684 (95.8%)** |
| badged books with `strong` evidence | 101 |
| with `moderate` | 588 |
| with `weak` | 2,286 |
| **with no verdict object at all** | **2,957** |

So the state was circular: "confirmed" resolved to the claim it exists to qualify, and this file's own promise — *"earned by evidence, never defaulted into"* — was false. Wiring it as-was would have changed nothing for 95.8% of books.

`confirmed` now also requires `strong` or `moderate` evidence: the same cut `isPublicFirst` makes for the headline count, and the same one `StrengthChip` already renders. After the fix:

```
candidate       5,217  87.9%
confirmed         627  10.6%
defeated           59   1.0%
not_applicable     29   0.5%
```

`scripts/audit/ft-claim-register.ts` produces that table and re-runs after any evidence sweep.

## Nothing is demoted

`is_first_translation` is untouched, the reconcile stays its single writer, and every badged book still carries a claim. 5,217 of them now state what we did instead of what no search can establish. A `candidate` is the honest default, not a penalty — and it is **monotone**: when ESTC lands and the evidence improves, books move up without anyone re-checking them by hand.

## The default is the safety property

`firstTranslationBadge` and `firstTranslationDescription` default to `candidate`, **not** `confirmed`. The collection, catalogue, author and exhibition cards render from `books_catalog`, which carries `ft_disposition` but no evidence strength — none of them can tell a cross-checked first from a legacy shim. An assertive default would let every one of those surfaces assert a universal negative by omission.

That is precisely how this area fails. Per `.claude/docs/invariants/first-translation-claims.md`: *"Every bug in this area fails toward a confident clean negative. Fourteen defects in one session, not one of which produced a false positive."* A test pins the direction, because a future refactor flipping that default would be silent and site-wide.

## In scope — where the claim is stated in prose

- **Evidence panel**: pill label, description, and colour. Gold asserts; stone reports. The candidate description states the search *and its limit* — "a record of the search, not proof that none exists" — because stating the limit is the claim, not a hedge.
- **Book page**: the publication-timeline label ("First English translation published" → "English edition published") and the stat-line chip.
- **`<meta description>`**: a candidate gets the plain `"English translation. "` prefix, not a search statement. 155 characters is not room for a search *and* its limits, and a truncated caveat is worse than none — it would put the bare assertion in front of search engines and AI crawlers with the qualifier cut off. Saying less is the honest option at this length.
- A candidate never gets the complete/modern/from-source gradations. Those are shades of a *conclusion*, not of the evidence behind it; drawing them dresses up a coin flip.

## Out of scope, deliberately

- **The card surfaces keep one label.** Giving them the real register needs an evidence-strength column on `books_catalog` plus a re-sync of ~100K rows — a separate PR. The safe default holds until then, so cards understate rather than overstate.
- **The 59 `defeated` books** — badged `is_first_translation: true` while `first_translation.verdict` reads `not_first`. The evidence panel already renders them correctly as "Existing translations", but the flag still drives the cards. That is a screening queue for #3524/#2933, not a rendering fix, and demoting a badge is sign-off-gated.
- **No new reference-set sources.** The ESTC harvest (#3522) is running separately; it moves books between registers without touching this code.

## Verification

- `npx tsc --noEmit` clean
- 148 test files / 1,910 tests pass
- New coverage: the evidence gate (strong/moderate confirm; weak, unrecorded, and the legacy-disposition shim do not), the candidate register's copy, and the `candidate` default on both label helpers

### DIFF
--- a/.claude/docs/invariants/first-translation-claims.md
+++ b/.claude/docs/invariants/first-translation-claims.md
@@ -21,6 +21,32 @@ predictive value at ~50%**: a coin flip. So:
 - **A null means different things in different traditions.** French has 23,035 English translations in the set, Syriac 119, and CJK is reachable only via MARC 880 (present on 2.3% of rows). Read reference-set *depth* beside every verdict; a flat badge cannot be honest across all of them.
 - **Keep "we could not ask" separate from "we asked and found nothing."** Conflating them turns an unasked question into a confident negative — the single most common way this system lies.
 
+**That distinction breaks at the PAYLOAD layer too, not just the evidence layer** (#3686,
+2026-08-07). Book surfaces are served from two sources for the same object: the Supabase
+`books_catalog` mirror (a projection, ~50ms) and the Atlas doc (complete, 1–5s). A pure
+classifier cannot tell which it received. `classifyFirstTranslationClaim` run inside
+`generateMetadata` — which resolves through `getCachedBookLookup`, i.e. the catalog row —
+returned `candidate` for **every** book, because the row carries `is_first_translation`
+but not `first_translation.evidence_strength`, and the classifier's "field absent" branch
+*was* its "evidence weak" branch. The assertive claim silently vanished from the meta
+description of the 627 books that had earned it. **Before classifying, assert the payload
+can answer** (`book.first_translation !== undefined`), fetch when it cannot, and fail
+toward the weaker claim. Check what `BOOK_SELECT` / `BOOK_DETAIL_SELECT` in
+`src/lib/books-catalog.ts` *omit*, not just what they carry.
+
+Two corollaries that cost real work here:
+
+- **A default is indistinguishable from a measurement**, so pick the direction that cannot lie. `firstTranslationBadge` and `firstTranslationDescription` default to `candidate`: every card surface renders from the catalog and none can evidence a first, so an assertive default would let all of them assert a universal negative by omission.
+- **A one-sided check on a two-sided change is a coin flip you will read as a pass.** The bug above and a successful fix produced identical output on the candidate direction. It surfaced only on the positive control — the book that should *still* assert. Test both registers.
+
+**And a state's name is a claim about its gate.** `classifyFirstTranslationClaim` gated
+`confirmed` on `isFirstTranslation()` — the *render* rule (first-family verdict, visible,
+some translated pages), which says nothing about evidence. That made 5,684 of 5,932 badged
+books (95.8%) `confirmed` while only 689 carried strong or moderate evidence, so the state
+resolved to "we badged it": the very claim it existed to qualify. Same family as a metric's
+name being a claim about its denominator. When a state promises "earned by evidence", the
+gate must read evidence.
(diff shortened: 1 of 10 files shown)


JDerekLomas:
## Verified on the preview, both directions

A second commit fixes a regression the **positive control** caught — worth recording, because the first check passed and meant nothing on its own.

Checking only the candidate direction showed "No prior translation found" and no "First English translation" anywhere, which looked like success. It wasn't: `generateMetadata` resolves its book through `getCachedBookLookup`, which serves the Supabase `books_catalog` row in the common case. That row carries `is_first_translation` and `ft_disposition` but **not** `first_translation.evidence_strength` or the screens — so classifying it returned `candidate` for *every* book, and the meta prefix had silently vanished for the 627 that earned it too. "The field is absent" had been read as "the evidence is weak", which is the one distinction this area exists to preserve.

`generateMetadata` now asks when the payload cannot answer: one projected lookup with `maxTimeMS: 3000`, taken only when the flag is set (~17% of visible books) *and* the payload actually lacks the field. Timeout or failure keeps `candidate` — understate, never overstate. The page body was never affected (`getBook` already refetches the full Atlas doc).

Against `sourcelibrary-v2-hmzlv280m` (● Ready):

| book | evidence | badge | `<meta description>` |
|---|---|---|---|
| `6a14e1dd…` *Thor bu bKa' brgyad 'don cha'i skor* | moderate | `First translation` | `First English translation. This collection…` |
| `6a601af0…` de Bary, *Untersuchungen über die Brandpilze* | weak / none | `No prior translation found` | `English translation. This monograph by Ant…` |

The candidate chip renders in stone (`#948d80`) rather than gold, with the tooltip *"a record of the search, not proof none exists"*.

Checks: `test` ✅ · DCO ✅ · Vercel ✅ (the first Vercel result was the usual pre-retry fail; the retry is Ready). `npx tsc --noEmit` clean, 148 files / 1,910 tests pass.

Related: filed #3687 for the 59 books badged while our own verdict names a prior — measured while wiring this, out of scope here because demoting a public claim is sign-off-gated.

Issue #3692, opened 2026-08-07
No detector asks whether a derived store is still being written (60-day embedding outage)


JDerekLomas:
## The instance

The page-embedding cron sat commented out behind a `#PAUSED-GEMINI` marker from
**June 9 to August 7** — 60 days. Measured when it was finally noticed: 2,462
live books with zero page vectors and 4,420 more under 90%, i.e. semantic search
blind on roughly 45% of the corpus.

Nothing alerted. Nothing could have: an unembedded book and a book with no
semantic match return the same empty list. It surfaced only because a reader
spent a working day concluding the corpus was thin on passages it holds, and
chasing *that* led to the vectors.

Fixed in #3691 (the pipeline now writes page vectors itself) and #3690 (the API
now says when a book is unembedded). This issue is about the **class**.

## The class

This repo has a strong culture of standing detectors — `scripts/audit/` holds
20+, and they are good ones: `r2-key-book-scope.mjs` catches a specific write
bug, `headerless-pages.mjs` catches missing chrome, `doc-staleness.mjs` catches
stale docs, `mcp-directory-contract.mjs` guards a published contract.

**Not one of them asks whether a derived store is still being written.** Every
detector checks correctness of what IS there; none checks that anything arrived.

Derived stores currently in play (all Supabase, all fed from Mongo):
`page_translations`, `book_embeddings`, `artwork_embeddings`,
`gallery_text_embeddings`, `clip_embeddings`, `books_catalog`.

## The trap, which is why this needs care rather than a one-liner

The obvious implementation — "alert if `max(updated_at)` is older than N days" —
**would have been wrong**, and measurably so.

`page_translations.updated_at` does not record when the row was written. It
mirrors the Mongo source's timestamp (that is the point of the neighbouring
`mongo_updated_at` column, which `--restale` diffs to detect re-OCR without
re-embed). Probed on 2026-08-07 *while six shards were actively writing to it*,
the column read **42 days old**. A naive monitor would have reported a healthy,
actively-filling table as stale, and — worse — would have been reporting the
same thing during the outage for a different reason, so the signal carries no
information either way.

The stores also disagree about their own freshness column:
`clip_embeddings` uses `created_at`, `gallery_text_embeddings` uses neither name.

So the detector has to be built per-store, from something that actually moves
when a write happens. Candidates worth weighing:

1. **Row count over time.** Snapshot counts to `system_config` or a small table
[...]

PR #3693, opened 2026-08-07
docs(invariants): a derived store outside the pipeline fails silently

JDerekLomas:
Up-ratchet from the session that found the page-embedding cron commented out
since June 9 — two months dark, 2,462 live books with zero vectors, semantic
search blind on ~45% of the corpus, and nothing detecting it because an
unembedded book and a book with no match return the same empty list.

Subsystem-triggered, so it lands in `.claude/docs/invariants/` with a routing
line in `CLAUDE.md` (204 lines, under the ~250 budget — nothing needed demoting).

Records the rules and, as importantly, the traps that produced confident wrong
answers while fixing it:

- **`updated_at` on `page_translations` mirrors the Mongo source, not the write
  time.** It read *42 days old* while six shards were actively writing to it. A
  monitor built on it lies in both directions — silent during the outage,
  screaming during the repair.
- **`--missing-only` cannot reach a book with zero rows.** It scans books already
  present, so a never-embedded book is invisible to it. That is what
  `--books-file` is for, and it was undocumented.
- **A flag can be accepted and ignored.** `--books-file` printed `(worker 3/8)`
  in its mode line and loaded the whole list anyway, so eight shards duplicated
  each other. The tell was byte-identical progress counters, which independent
  shards cannot produce.
- **A fixed staleness threshold flagged two healthy weekly jobs as dead.**
- **The crontab is not the whole schedule** — `enrich-worker` appears zero times
  in it and runs constantly, dispatched by `scheduler.mjs`.
- **A job with no log redirect cannot be observed even in principle**
  (`restic-backup.sh` is the sharp case).

Plus the session handoff — a public-worthy technical postmortem, no PII or
business material.

Related: #3690, #3691, #3692.


JDerekLomas:
**Triage finding (read-only backlog sweep, 2026-08-28): the lib/tests already landed byte-identical on main.**

\`src/lib/semantic-coverage.ts\`, \`tests/unit/semantic-coverage.test.ts\`, and \`tests/unit/page-embedding-text.test.ts\` on \`main\` are already present and diff to **zero lines** against this branch's versions — that piece already shipped. \`src/app/api/mcp/route.ts\` on main has diverged heavily since (many unrelated additions: server-version consolidation, edition-language apparatus, gallery viewer resource, etc.) and neither main's nor this branch's route.ts references \`semantic-coverage\` yet, so the "post listing" wiring itself doesn't look wired in on either side. Recommend closing and re-opening a fresh, small PR against current main if the MCP listing integration is still wanted — rebasing this branch's route.ts changes onto main's current version isn't worth the conflict given how much unrelated code moved.

JDerekLomas:

Issue #3716, opened 2026-08-07
books.published and books.year disagree on 1,035 visible books — and different APIs read different ones


JDerekLomas:
`books` carries two date fields — `published` (string) and `year` (number) — and on **1,035 visible, processed books they disagree**. Different API surfaces read different ones, so the same book is dated two ways depending on which endpoint you ask.

## Repro

`69937973b0a84a5763964d43` — the Bekker 1831 Aristotle, which is the reference frame for every Aristotle citation in the corpus:

```
title:      "Aristotelis Opera, Vol. 2 (Bekker Edition, Reimer 1831)"
published:  "1831"
year:       1890
```

`get_quote` returns `published: "1831"`. The newly shipped `get_locus` (#3703) returns `year: 1890`, because `locus_books` copies `year`. Both are faithfully reading the record; the record disagrees with itself.

## Measurement

Restricted to the unambiguous case — `published` exactly four digits, `year` positive — over visible books with `pages_count > 0`:

```
unambiguous pairs:  14,954
agree:              13,919
DISAGREE:            1,035   (6.92%)

by gap:  1-4 yrs: 305 | 5-19: 324 | 20-99: 262 | >=100: 144
```

**Method note, because the first pass was wrong and the error is instructive.** A naive `\d{3,4}` parse over all books reported 1,714 disagreements — but most were artifacts of the parse, not the data. `published: "1985 BC"` against `year: -1985` is the two fields *agreeing*, with `year` correctly signed. `published: "189-"` against `year: 1890` is uncertain-date notation being normalised, also correct. Filtering to `^\d{4}$` removes that whole class. The 1,035 above are cases where both fields state a plain year and state different ones.

## Which is right

In every case I checked by hand, `published` matched the title and `year` did not:

| id | published | year | title says |
|---|---|---|---|
| `69937973…` | 1831 | 1890 | Reimer **1831** |
| `69af0deb…` | 1566 | **2007** | "De Revolutionibus … (**1566** alt ed.)" |
| `69aeaa9a…` | 1519 | 1450 | Facetiae |

The 2007 is the tell — that is a scan or ingest year sitting in a publication field. But this is a handful of cases, not a survey, so treat "published is canonical" as a hypothesis to confirm rather than a finding. The 305 pairs differing by 1–4 years are a different population again and may be legitimate edition-vs-printing ambiguity.

## Why it matters beyond tidiness

[...]

Issue #3714, opened 2026-08-07
Two sessions can silently overwrite each other's derived collections — locus_anchors near-miss


JDerekLomas:
## What happened

On 2026-08-07 two sessions independently implemented #3661 within the same hour. Both wrote a corpus sweep that rebuilds the derived `locus_anchors` collection the same way:

```js
await col.deleteMany({});
if (docs.length) await col.insertMany(docs);
```

One sweep applied ~23:10Z (4,279 anchors, 7 editions). The other applied at 23:19:43Z (6,324 anchors, 10 editions, richer schema). **Ordering alone decided which survived.** The later one won and is correct, so nothing was lost — this is a near-miss report, not an incident.

Had they run in the other order, the shipped implementation's data would have been silently replaced by a poorer superset-of-nothing, and — because the two schemas differ (`ref_page`/`ref_section`/`ref_line`/`basis` vs `page`/`section`) — the shipped `/api/locus` route would have queried fields that no longer existed and returned **zero results for every locus**. No error, no exception, no failed write. A feature that had just passed CI would simply stop answering.

## Why the existing guards don't cover this

- `deleteMany` + `insertMany` is not a "batch delete" in the sense CLAUDE.md's Data Protection section means — no books, pages, or source material are touched. It is a legitimate rebuild of a derived collection, and rebuilding wholesale is the *right* call there (a partial update leaves rows from a previous parser version with no way to tell which is which).
- The worktree isolation that protects working trees does nothing here. Both sessions had correct, isolated checkouts. **The shared mutable resource was the production database, which no worktree isolates.**
- `git` shows nothing: neither session's branch conflicts with the other's, because the collision is in data, not code.

This is the same shape as the detached-job hazard already documented in CLAUDE.md ("A detached job writing into a worktree makes every git operation there destructive") — but inverted: there the shared resource is a file and the danger is git; here the shared resource is a collection and the danger is another session's sweep.

## Suggested guards, cheapest first

1. **Stamp and check a writer identity.** A rebuild script writes `extractor_version` (both of these already do) plus a run id. Before `deleteMany`, read the existing rows: if they carry a *different* `extractor_version` than the one about to be written, refuse unless `--force`, and print what is about to be replaced. This costs one query and would have caught this exactly.
[...]

PR #3728, opened 2026-08-08
docs: three lessons from the MCP feedback session, and one correction

JDerekLomas:
Session wrap-up from the MCP feedback triage. Both directions of the ratchet, plus the handoff.

## Up — two new invariants

**`tests-that-are-not-guards.md` gains a fourth failure mode**, and it is the only one where the test itself has no defect. The assertion is real, it exercises deployed code, its input is genuine — and it is still blind, because it only asks whether the result set is non-empty.

Every assertion in `tests/smoke/mcp-search.test.ts` passed throughout the period `search_within_book` returned 48 pages of front matter with the answer ranked ~50th (#3653 → #3680). The results were never empty; they were **plausible and wrong**, which for a corpus of historical text is the normal shape of a bug — a real page of real Aristotle, correctly cited, that simply is not the passage asked for. The entry records the golden-passage shape that does catch it (#3702), its two required controls, and why a known bug is pinned with `it.fails` rather than `it.skip`.

**`CLAUDE.md`'s Multi-Session Awareness gains the mirror of the detached-job rule**, directly beneath it. That one protects a file from git; this one protects a collection from another session. Two sessions built #3661 within the same hour and both rebuilt `locus_anchors` with `deleteMany({})` — ordering alone decided which survived, and the loser's schema differed, so the just-shipped `/api/locus` would have returned zero for every locus with no error and green CI. Worktrees isolate checkouts; nothing isolates production Mongo. #3714.

## Down — one correction in place

The Vercel note described a first-build-then-retry. **The commit statuses do not show that.** Three consecutive PRs today (#3702, #3713, #3715) each received *two* statuses with the same `created_at` second: a `failure` pointing at a generic `vercel.link/<id>` **identical across unrelated PRs**, and a `pending` pointing at the real deployment, which builds 5–6 min and passes.

So there is no retry — the fail is spurious and posted alongside the real build. Practical advice unchanged (trust `test`/DCO, wait for the deployment); mechanism rewritten, with the two commands that check the deployment rather than the check name.

Corrected in place rather than appended beside the thing it corrects, per the knowledge-layer rule.

## Budget

`CLAUDE.md` 205 → 206 lines, inside the ~250 cap. Nothing needed demoting.

## Handoff

`.claude/handoffs/2026-08-07-…` added with `git add -f`, deliberately — that directory is gitignored here and defaults to the private ops repo. This one is a purely technical postmortem with no PII, secrets, or business strategy.

It also records two near-miss corrections caught before publishing, both the same shape (reasoning from a number without checking what produced it):

- `contains_works` read as "0 books" — it is an object with a `.works` array, not an array, so `contains_works.0` matches nothing. Real figures: 2,090 books, 1,507 with ≥1 work.
- Date conflicts read as 1,714 books — but `published: "1985 BC"` vs `year: -1985` is *agreement* with `year` correctly signed, and `"189-"` vs `1890` is uncertain-date notation. Restricting to unambiguous four-digit pairs gives 1,035 (#3716).


JDerekLomas:

JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded. The load-bearing lesson — two sessions clobbering a shared Mongo collection, #3661, assert the SHAPE not the count, guard on the READ side — is on main at CLAUDE.md line 71 with the same incident cited, and `tests-that-are-not-guards.md` has moved through five commits since (#4177 → #4645). CLAUDE.md changed 49 times since this opened, so the branch cannot be revived as-is. **One fragment did not land:** the Vercel spurious-failure mechanism (two same-second statuses sharing one `vercel.link` id). If still true, it belongs in `.claude/docs/invariants/pr-checks.md` as a fresh 10-line PR.

PR #3742, opened 2026-08-08
feat(import): acquire Epicurus, Longinus, Sappho, Sextus Empiricus, Polybius

JDerekLomas:
Acquires 40 editions (15,722 pages) across the five requested authors. All imported **hidden** — flip visible only after an OCR/translation QA pass.

## How they were chosen

Followed the canonical loop in `.claude/docs/import-workflow.md`: enumerated from Internet Archive with `enumerate-dedupe-source.ts` (one query per author), deduped on `source_fingerprint` against the whole catalog, then subject-filtered by hand. That last step was load-bearing here — keyword enumeration returned Daudet's novel, Bret Harte's short stories and four printings of Grillparzer's tragedy under "Sappho", and Anatole France plus a shelf of garden essays under "Epicurus". None of it is in this list.

Selection leads with the Greek, then the editions that carried these texts into later thought. Modern scholarship *about* these authors is excluded, as is anything after 1929. All 40 probed against `archive.org/metadata` before import: none lending-restricted, all with downloadable page images.

| | new | notable |
|---|---|---|
| Epicurus | 6 | Charleton's 1656 first English Epicurus; Rosini's 1818 *De Natura* recovered from the Herculaneum papyri |
| Longinus | 8 | Boileau's *Traité du Sublime*; Rhys Roberts's Greek text after Parisinus 2036 |
| Sappho | 10 | Wolf 1733 and Neue 1827 (the standard 19th-c text); Diehl's *Supplementum Lyricum* papyrus fragments |
| Sextus Empiricus | 5 | Hervet's 1569 Latin *Adversus Mathematicos* — ancient scepticism as Montaigne and Descartes met it; Bekker's Greek |
| Polybius | 11 | Casaubon 1609; Book VI on the mixed constitution printed alone in 1539; Schweighäuser; Büttner-Wobst |

Also completes the Shuckburgh Polybius — we held only volume 2.

## The bug this PR also fixes

The batch's first run passed each book's **edition** language as the route's `original_language`. `resolveLanguage()` nulls that field when it equals `language` (FRBR work == manifestation), which left `is_translation` false, and `classifyTextRole()` returns `'original'` for any non-English scan without that flag. Result: Boileau's French Longinus, Gori's Italian Longinus, Hervet's Latin Sextus and Huart's French Sextus were filed as original-language Greek sources.

The import script now passes `language` and `original_language` separately, and `fix-greek-five-language-roles.mjs` repairs the affected rows — 24/24 applied and verified. It distinguishes three shapes, because they differ downstream: TRANSLATION (Greek not in the scan), APPARATUS (Greek text with Latin/vernacular commentary — stays `original`, `original_in_scan: true`), and FACING (Greek printed alongside a translation).

Gassendi's *Philosophiae Epicuri Syntagma* is deliberately excluded from the Greek-origin sweep — it is his own Latin composition, not a rendering.

## Out of scope

- **`work_id` is unset on all 40.** Work-level clustering is not automatic at import (#2318), so these join the existing unclustered tail rather than being hand-assigned here.
- **No OCR.** The pipeline is deliberately paused; nothing in this PR unpauses it or spends on these pages.
- **No collection tagging or visibility flips** — both wait on QA.

### DIFF
--- a/scripts/import/greek-five-authors-batch.mjs
+++ b/scripts/import/greek-five-authors-batch.mjs
@@ -1,0 +1,151 @@
+#!/usr/bin/env node
+/**
+ * Acquisition batch: Epicurus, Longinus, Sappho, Sextus Empiricus, Polybius.
+ *
+ * Sourced 2026-08-08 via the canonical loop in .claude/docs/import-workflow.md:
+ * enumerated from Internet Archive with scripts/import/enumerate-dedupe-source.ts
+ * (one query per author), deduped against the whole catalog on source_fingerprint,
+ * then subject-filtered by hand. Keyword enumeration was extremely noisy here —
+ * "Sappho" alone returned Daudet's novel, Bret Harte's short stories and four
+ * printings of Grillparzer's tragedy; "Epicurus" returned Anatole France and a
+ * shelf of garden essays. None of that is in this list.
+ *
+ * Selection rule: the Greek original first, then the editions and translations
+ * that carried the text into later thought — Hervet's Latin Sextus (which put
+ * ancient scepticism into Montaigne and Descartes), Boileau's Longinus (which
+ * made the Sublime a European idea), Casaubon's Polybius, Charleton's first
+ * English Epicurus. Modern scholarship *about* these authors is excluded; so is
+ * anything published after 1929, for copyright.
+ *
+ * All 40 items were probed against archive.org/metadata before listing: none is
+ * lending-restricted, all have downloadable page images.
+ *
+ * Imports land HIDDEN (the route sets visible:false + hidden:true). Flip to
+ * visible only after an OCR/translation QA pass — see the import-workflow doc.
+ *
+ * `lang` is the EDITION's language and `orig` the SOURCE WORK's — they are two
+ * different fields and the route wants both. Do not pass the edition language as
+ * `original_language`: resolveLanguage() nulls original_language when it equals
+ * language (FRBR work == manifestation), which also leaves is_translation false,
+ * and classifyTextRole() then returns 'original' for ANY non-English scan. The
+ * first run of this script made that mistake and filed Boileau's French Longinus,
+ * Gori's Italian Longinus, Hervet's Latin Sextus and Huart's French Sextus as
+ * original-language sources. Repaired by
+ * scripts/maintenance/fix-greek-five-language-roles.mjs.
+ *
+ * Usage:
+ *   set -a; source .env.production.local; set +a
+ *   node scripts/import/greek-five-authors-batch.mjs --dry-run
+ *   node scripts/import/greek-five-authors-batch.mjs --commit
+ */
+
+const COMMIT = process.argv.includes('--commit');
+const CRON_SECRET = process.env.CRON_SECRET;
+if (COMMIT && !CRON_SECRET) { console.error('CRON_SECRET not set'); process.exit(1); }
+const BASE = 'https://sourcelibrary.org/api/import/ia';
+
+const BOOKS = [
+  // ── EPICURUS ───────────────────────────────────────────────────────────────
+  // Held already: Usener's Epicurea, Bailey's Extant Remains, Vogliano's
+  // Herculaneum scripta, Gassendi's De Vita et Moribus, Charleton's Physiologia,
+  // Gassendi's Animadversiones on Diogenes Laertius X.
+  { id: 'bim_early-english-books-1641-1700_epicvrvss-morals-_epicurus_1656', title: "Epicurus's Morals, Collected and Faithfully Englished", author: 'Epicurus (tr. Walter Charleton)', year: 1656, lang: 'English', orig: 'Greek' },
+  { id: 'bim_early-english-books-1641-1700_philosophi-epicuri-synt_gassendi-pierre_1668', title: 'Philosophiae Epicuri Syntagma', author: 'Pierre Gassendi', year: 1668, lang: 'Latin', orig: null },
+  { id: 'epicurussmorals00descgoog', title: "Epicurus's Morals, with a Life by Saint-Evremond", author: 'Epicurus (tr. John Digby)', year: 1712, lang: 'English', orig: 'Greek' },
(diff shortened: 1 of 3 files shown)


JDerekLomas:
## Follow-up: work identity (c617c5c4)

All 40 now carry `work_id` **and** `author_id`, and the pre-existing holdings for the same five authors were re-clustered alongside them — 63 books, 6 author docs.

### Two author anchors were wrong in production

| | was | now | |
|---|---|---|---|
| `authors/sextus` | `Q1270100` | `Q236594` | Q1270100 is the ***Sentences of Sextus***, a 2nd-c collection of maxims — `P31` = text, not a human. Four **visible** books hung off it. |
| `authors/longinus` | `Q436634` | `Q744540` | Q436634 is **Cassius Longinus**, the Neoplatonist to whom *On the Sublime* was traditionally but wrongly ascribed. Every Longinus book we hold is the treatise, i.e. Pseudo-Longinus. Wikidata carries both on the work (`P50 = Q744540, Q436634`), so this is a genuine scholarly conflation, not a typo. |

Also added **Epicurus** — absent entirely from a 4,825-author thesaurus — plus Sappho, Metrodorus of Lampsacus and Anacreon, and tombstoned the editor-as-author doc `sappho-ed-henry-thornton-wharton` into `sappho`.

### Why not the resolver

`resolve-work-ids-wikidata.mjs` is author-anchored and needs `author_id`, which no IA import sets. And even with the authors fixed it would misfire on two of the five: Wikidata models Sappho as **~208 individual numbered fragments** ("Sappho fr. 174 Voigt") while every book we hold is the whole corpus, so containment matching would snap editions onto arbitrary fragments; Metrodorus has zero English-labelled works.

So the QIDs are hand-adjudicated, each verified for correct author (`P50`) and for **work-level rather than edition-level** granularity (`P31` ≠ `Q3331189`). That second test is load-bearing: it is what keeps the Histories on `Q250816` rather than `Q16038921` / `Q53748127`, which are *editions of* it and would have produced a wrong cluster.

### Clusters

| work_id | n | |
|---|---|---|
| `Q6245111` On the Sublime | 11 | Greek, Latin, English, French, Italian |
| `Q250816` The Histories | 19 | was split across **four** work_ids incl. one split by volume number |
| `Q3058641` Outlines of Pyrrhonism | 4 | |
| `Q20379739` Against the Mathematicians | 2 | |
| `local:a:sappho:fragmenta` | 8 | hand-minted; see above |
| `local:a:sextus:opera` | 4 | container |
| `local:c:greek-lyric:*` | 8 | containers (anthologies) |
| Epicurus: `Q7091086`, `Q740527`, `:morals`, `:extant-remains` | 5 | |

Casebook rules applied: volumes cluster under one id (#5); Sammelbände get their own and are never badged (#4) — Sextus's *Opera* contains **both** the Hypotyposes and Adversus Mathematicos, so filing it under either would be wrong.

The overwrite guard refuses any `work_id` whose source is not `local-mint`; `promote_from` names per cluster the local ids verified to mean the same work, so promoting the `llm-verified` Outlines cluster onto `Q3058641` stays a deliberate decision rather than a blanket override.

### Verification

Ran the system's own `scripts/analysis/work-coverage.mjs`. Every cluster resolves, and `local:a:epicurus:morals` reports `hasOriginal: false` — two English renderings of the ethical remains with no Greek original held. A real gap, and one the fragmented local-mint ids could not express.

### Also worth knowing

The import route calls `queuePreviewOcr` (`route.ts:362`), which OCRs the first 25 pages of every import. It ran automatically on all 40 — ~1,000 pages, ~$2.73 computed. It earned its keep once: IA titled `gri_33125008560068` with the generic series name, and the OCR'd title page read **TOMUS II**, which is how that volume got filed correctly.

Out of scope, deliberately: the 23-book Herculaneum series is minted **per volume** by an earlier curated backfill; our new volume joins that convention rather than the series being collapsed under casebook #5.

JDerekLomas:
## Correction to the two claims above

I read `resolve-work-ids-wikidata.mjs` and `mint-local-work-ids.mjs` properly afterwards and replayed the matching logic offline. Two things I asserted were wrong.

**1. Sappho's fragments would NOT have snapped.** I said containment matching "would snap editions onto arbitrary fragments." It would not. The author-name strip, the explicit `fragments`/`fragment` skip, and the ≥4-character token filter between them reduce every `Sappho fr. 174 Voigt` candidate to `{voigt}` or drop it, so `inter = 0` and all four of our Sappho titles return **NO-MATCH**. The resolver is safe here; it simply yields nothing. Hand-minting was still the right call, but because the resolver *cannot produce anything*, not because it would produce garbage. Softer, and less flattering to the decision.

**2. The Polybius volume split was deliberate, not junk.** I called the local-mint ids fragmentation. `mint-local-work-ids.mjs` **holds any multi-edition cluster whose members carry a volume/part marker by default** (`PART_RX`, `--include-parts` to override), on an explicit "under-cluster over mis-cluster" rule — because "one work in N volumes" and "N works in a series" are not distinguishable from the title. So `…:1-history-megalopolitan-vol` / `…:2-history-megalopolitan-vol` was the minter correctly declining to fuse, awaiting exactly the human-gated merge this PR performs. Collapsing them onto `Q250816` is still right; describing the prior state as sprawl was not.

## The one real resolver bug this did surface

Three Wikidata items share the label **"The Histories"**: `Q250816` (the work), `Q16038921` (an edition of it) and `Q53748127` (Paton's 1920s translation). The SPARQL has no `P31` filter, so all three are candidates; containment ties at **1.00** for all three; and the tie-break is `cont > bestCont` (strict), which keeps whichever SPARQL returned **first**. Replaying with the candidate list in two orders:

```
HIGH [order A] "The Histories of Polybius, Vol. 1" → Q250816     cont=1.00
HIGH [order B] "The Histories of Polybius, Vol. 1" → Q53748127   cont=1.00
```

Same book, same code, different `work_id` depending on SPARQL row order — written at **HIGH confidence and auto-applied** under `--apply`. The hand-grown label blocklist already in the code (`^opera |^the works of|collected works|…`) looks like prior scar tissue from the same problem. A `P31` filter is the principled version and is a one-line SPARQL change.

Filed as a separate concern rather than bundled here.

PR #3760, opened 2026-08-08
fix(archive): record the archiving on the book, not only on its pages

### LINKED ISSUE #3712: archive-images route never updates books.pages_archived, so archived books stay selectable for re-archiving
## What happens

`POST /api/books/[id]/archive-images` downloads page images and writes `archived_photo` onto each `pages` doc — but it never writes to `books`. It reads the book twice (lines 35 and 254) and updates nothing there.

`books.pages_archived` is a **cached** count of pages with `archived_photo` (`src/lib/types/book.ts:80`). Archiving a book entirely through this route therefore leaves it reading `pages_archived: 0` while all its pages are on R2.

## Why that costs real work

Selectors key on the counter, not on the pages. `scripts/workers/archive-erara.mjs` picks work with `pages_archived: { $not: { $gte: 1 } }`; `archiving-watchdog.mjs` buckets a book as "progressing" only when `(b.pages_archived || 0) > 0`. So a book fully archived by this route:

- stays permanently eligible for re-archiving, and a re-run re-downloads every page from the source
- is reported as un-started by the watchdog
- undercounts in any archived-coverage figure

Observed 2026-08-07 on `6a761f56af6d06b393a52a44` (Aristotelis Opera, Vol. III, Bekker 1831): 762/762 pages carried `archived_photo`, `books.pages_archived` read `0`. Corrected by hand to 762.

This is the mirror image of the `archived_photo` invariant in CLAUDE.md ("a bad write can erase its own repair path"). Here a *correct* write fails to record itself, so the work is simply repeated — silently, and at full bandwidth cost against the source library, which is the part that matters for providers that rate-limit or block us.

## Fix

After the page loop, recount and set the counter in the same request:

- count `pages` where `book_id` matches and `archived_photo` is a non-empty string
- `$set` that onto `books.pages_archived`

Recount rather than increment — the route is called repeatedly with `limit`, and an increment would drift on retries and on the Cloudflare 524s this route routinely returns (the edge times out at ~100s while the function keeps working to its 300s `maxDuration`, so the caller retries work that in fact completed).

## Also worth deciding

The route writes `archived_photo` but no `display_photo`. The reader reads `display_photo`, so a book archived this way has no R2 display variant and falls back to the source URL. Vol. III has 0/762 `display_photo` against 762/762 `archived_photo`; Vol. V, archived by the normal worker, has both. Either the route should generate variants or that gap should be documented as expected.

## Blast radius

Unmeasured. Bounded to books archived via this route — the watchdog's `--rearchive` path and manual calls — not the Hetzner/local workers, which set the counter themselves.

JDerekLomas:
Closes #3712.

## The bug

`POST /api/books/[id]/archive-images` writes `archived_photo` onto every page and never touches `books`. But `books.pages_archived` is a **cached** count, and the selectors key on the counter, not on the pages:

- `archive-erara.mjs` picks work with `pages_archived: { $not: { $gte: 1 } }`
- `archiving-watchdog.mjs` treats a book as progressing only when it is `> 0`

So a book fully archived through this route still reads `0` — permanently eligible for re-archiving, and a re-run **re-downloads every page from the source**. That is the expensive half, and not in dollars: it is full-bandwidth refetching against providers that rate-limit or block us outright (e-rara already 403s our datacenter IPs).

Observed on Aristotelis Opera Vol. III (`6a761f56af6d06b393a52a44`): 762/762 pages carrying `archived_photo`, book reading `pages_archived: 0`.

Same family as the `archived_photo` invariant in CLAUDE.md, inverted — there a *wrong* write made a page look done and hid it from repair; here a *correct* write isn't recorded, so the work is simply repeated.

## The fix

The handler already recounts archived pages to build its response. This persists that number.

**Set from the recount, never incremented.** The route is called repeatedly with `limit`, and the Cloudflare edge times out at ~100s while the function runs to its 300s `maxDuration` — so callers legitimately retry work that in fact completed (I hit exactly this driving Vol. III: two calls returned 524 and both had archived their pages). An increment would drift on every one of those.

## Also documented, not fixed

This route writes no `display_photo` variants — the worker archivers do, via `uploadPageVariants`. A book archived here therefore has `archived_photo` on every page and no display variant, and the reader falls back to the source URL. Vol. III: 0/762 display variants against 762/762 archived; Vol. V, archived by the normal worker, has both.

Left as-is deliberately: the display backfill sweep is live and cursor-driven, so the gap closes on its own. Recorded in the handler docblock so the asymmetry isn't mistaken for a reader bug.

## Verification

`npx tsc --noEmit` clean. The counter path is the same recount the endpoint has always returned in `archivedPages`, so the stored value and the reported value cannot disagree.

### DIFF
--- a/src/app/api/books/[id]/archive-images/route.ts
+++ b/src/app/api/books/[id]/archive-images/route.ts
@@ -18,6 +18,17 @@ const ARCHIVABLE_SOURCES_REGEX = /archive\.org|gallica\.bnf\.fr|digitale-sammlun
  * Download images from external sources and upload to Vercel Blob.
  * Supports: Internet Archive, Gallica (BnF), MDZ (Bavarian State Library)
  * This makes images available even when source sites are down.
+ *
+ * Writes `archived_photo` (plus a thumbnail) per page and syncs the book's
+ * `pages_archived` counter — see the recount near the end of the handler for why that
+ * matters.
+ *
+ * KNOWN GAP, deliberate: this route does NOT generate `display_photo` variants, which is
+ * what the reader actually reads. The worker archivers do (via `uploadPageVariants`), so a
+ * book archived through this route has `archived_photo` on every page and no display
+ * variant, and falls back to the source URL until the display backfill sweep reaches it.
+ * That sweep is live and cursor-driven, so the gap closes on its own; it is recorded here
+ * only so the asymmetry is not mistaken for a bug in the reader.
  */
 export const POST = withAuth(async (request, session, context) => {
   try {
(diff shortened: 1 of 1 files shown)


JDerekLomas:
**Triage finding (read-only backlog sweep, 2026-08-28): looks superseded.**

Current \`main\`'s \`src/app/api/books/[id]/archive-images/route.ts\` already performs a book-level \`updateOne\` and carries a comment about a book being "archived entirely through this route" — i.e. it already records the archiving on the book, not just on pages, which is this PR's stated purpose. Recommend a quick side-by-side read to confirm main's mechanism covers the same case this PR was written for, then close if so.

JDerekLomas:

Issue #3730, opened 2026-08-08
[Tracking] August 2026: make the identity stack load-bearing, then show it to a reader


JDerekLomas:
Execution home for August 2026. Architecture: #3258. Rationale, with the measurements behind it: `.claude/docs/identity-stack-rationale.md`.

**Restructured 2026-08-08 (v3)** after Derek's review question — *"is this in a pipeline? seems so complicated."* It wasn't, and that was the complexity. Identity is now **Phase 0 of the pipeline** (PR #3746), which replaces the old week 1's four mechanisms (shared helper + CI tripwire + nightly detector + re-sweep) with one worker and one alarm. Guiding constraints from Derek: keep it as simple as possible, high quality, clear purpose; Claude makes most judgment calls; scripts, systems and cheaper models do the work at scale — paid AI only with a sample-first estimate and sign-off.

## The one-sentence purpose

**A reader on any book should be told what else we hold of that work.** Identity correctness is the cost of shipping that sentence; everything below either makes the four keys true or shows them to a reader.

## The shape (simplified)

```
Phase 0 (pipeline, runs itself)   →   drain the queues (judgment, mostly scripted)   →   show a reader (the rail)
```

### 1. Phase 0 — identity runs itself ✅ (PR #3746)

One writer, `computeIdentityFields()`; two call sites (import route + `identity-worker.mjs`, 2-hourly Hetzner cron, runs even while the pipeline is paused because it costs nothing). Backlog cleared at ship time: 8,930 books stamped, queue 0. The alarm is the worker's `stale_missing` count in `cron_runs` — nonzero means the worker stopped. **Nothing here needs remembering; that's the point.**

Remaining in this lane:
- [x] Install the cron on the live Hetzner crontab ✅ 2026-08-08 — installed, worker verified on the box (stamped 0, queue 0).
- [x] `scripts/audit/identity-stack-gates.mjs` ✅ PR #3775 — prints the Done table live.
- [x] **Replay gate** ✅ PR #3775 (`dedup-replay-sample.ts`). Baseline 2026-08-08 with green controls: same-source ~100% both strata; cross-source Latin 89.5% → **100.0% after the drift restamp** (4,627 books stamped by older normalizer versions, re-stamped via `identity-worker --restamp`); cross-source non-Latin **0.0%** — the one remaining recall hole, fixed only by the edition-key tier (§2). Details: PR #3775 + comment.

### 2. Dedup reads the layer it writes

[...]

Issue #3102, opened 2026-07-09
Duplicate management: make dupes relations, not prose (bug / copy / edition model)


JDerekLomas:
## Problem

"Duplicate" currently collapses three different relationships, and resolutions are stored as free text. Measured 2026-07-09 against prod:

| Layer | Definition | Measured | Verdict |
|---|---|---|---|
| Same digital object ×2 | identical `ia_identifier` | 61 extra copies | bug residue (import dedup blocks new ones) |
| Same edition, both visible | same normalized title+author+**year**, `visible:true` | **270 clusters / ~460 extra visible books** | the reader-facing problem |
| Different editions, both visible | same normalized title+author, different years | 123 clusters | mostly legitimate editions, not presented as such |
| Hidden dupes, link in prose | `hidden_reason: "duplicate of <slug>"` | **4,916 books** | resolved but unqueryable |
| Hidden dupes, structured | `duplicate_of` field | 15 books | right shape, barely used |
| Multi-edition works | shared `work_id` | 3,128 works | meaningful — a feature, keep |

Consequences today: nothing links a hidden copy to its keeper (keeper deletions orphan silently), "other scans of this edition" can't be rendered, dedup audits re-derive everything from scratch, and bulk visibility sweeps can't distinguish a hidden *copy* from a hidden *problem book* (see hidden_reason sweep rule from the Kloss takedown).

## Model

Three relationships, three treatments:
1. **Bug** (same digital object): merge/hide, backfill pointer. Should trend to zero — `src/lib/dedup.ts` tier-1/3 blocks new ones at import.
2. **Copy** (same edition, different scan): keep ONE keeper visible (best scan quality/completeness), others hidden with `duplicate_of: <keeper id>` + `hidden_reason: 'duplicate'` (machine value). Copies stay credited and become a reader feature: "other scans of this edition."
3. **Edition** (same work): NOT a duplicate. Belongs to `work_id` clustering (#2318 stack) and the related-editions UI.

## Plan

1. **Backfill** `duplicate_of` from the 4,916 prose `hidden_reason` strings (slug is embedded in the string). Script with dry-run + backup, verify slugs still resolve, report orphans (keeper since deleted/renamed).
[...]

PR #3775, opened 2026-08-08
feat(audit): replay gate + gates table — and the third failure class they caught on day one

JDerekLomas:
Part of #3730 (lane 1). Two read-only instruments plus the repair they immediately justified.

## The instruments

**`scripts/audit/identity-stack-gates.mjs`** — prints #3730's Done table live, dated. "Done" becomes a command instead of issue prose (which this repo has measured to drift badly — #3102, #2567).

**`scripts/audit/dedup-replay-sample.ts`** — the end-to-end gate: sample live books, replay each through the real `checkDuplicate()` twice — identifiers kept (same-source) and stripped (cross-source) — stratified by tier-2 reachability, with positive and negative controls.

## The controls earned their keep twice before lunch

1. First run's positive control: **0/20**. The probe had sampled `duplicate_of` pointers, which are overwhelmingly artworks — whose identity lane is sha1/CLIP, not title dedup. Books-only: **18/18**. (Lesson already in memory: a probe's "not found" is worthless until it has returned "found.")
2. First stratifier filed «Чехова. Том 3» as Latin — the ASCII normalizer reduces it to `"3"`, non-empty but just as invisible to tier 2 as `''`. Stratum boundary now matches tier 2's own `length >= 5` guard.

## Baseline (2026-08-08, controls green)

| replay | Latin | non-Latin |
|---|---|---|
| same-source (identifiers kept) | 99.0% | 100.0% |
| cross-source (identifiers stripped) | **89.5%** | **0.0%** |

The non-Latin **zero** is the ASCII hole measured as behavior: a non-Latin book re-imported from a different provider sails past the title tier entirely. Fingerprint/IIIF have been carrying the whole load — which is why zero such dupes exist *so far*, and why the edition-key dedup flip (#3730 §2) is justified by measurement, not taste.

## The third failure class: normalizer drift

Most Latin misses weren't the known holes. They were books whose **stored** normalized fields no longer equal what the **current** code computes at query time — stamped by older or divergent implementations:

- old article list that kept leading "de": stored `de concilio liber…` vs computed `concilio liber…`
- unsorted author tokens from direct-insert scripts: stored `manuzio aldo` vs canonical sorted `aldo manuzio`

A book in that state is silently unfindable by tier 2 — recall loss with no error anywhere. **`identity-worker.mjs` gains `--restamp`**: recompute for ALL books, rewrite only diffs; never run by the cron. Run against production: **4,627 drifted books re-stamped in 101s**. Queue metrics in `cron_runs` now always report the cron's absence-queue so a restamp run can't read as an outage.

A post-restamp verification replay is running; result lands as a comment on this PR.

## Verification

- `tsc --noEmit` clean; parity + edition-key suites pass (22 tests).
- Gates table against prod: Phase 0 at 0 missing / `stale_missing` 0; 222 work/edition conflicts; 1,638 live books in multi-language work clusters; 1,993 dark pointers; 297 both-visible clusters.

### DIFF
--- a/scripts/audit/dedup-replay-sample.ts
+++ b/scripts/audit/dedup-replay-sample.ts
@@ -73,9 +73,15 @@ function fullMetadata(b: BookDoc) {
   };
 }
 
-/** Non-Latin stratum: the stored normalized_title is empty while a real title exists. */
+/**
+ * Non-Latin stratum: the stored normalized_title is too short to reach tier 2
+ * (< 5 chars) while a real title exists. Not merely empty: «Чехова. Том 3»
+ * ASCII-normalizes to "3" — non-empty, but just as invisible to the title
+ * tier as the empty string, and stratifying it as "Latin" understates the
+ * non-Latin miss population.
+ */
 function isNonLatin(b: BookDoc): boolean {
-  return !!b.title && (b.normalized_title ?? '') === '';
+  return !!b.title && (b.normalized_title ?? '').length < 5;
 }
 
 async function replay(db: Db, candidate: Parameters<typeof checkDuplicate>[1]): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


JDerekLomas:
## Post-restamp verification — the drift repair closed the Latin gap completely

Same probe, same controls (positive 18/18, negative 0/3), run after the 4,627-book `--restamp`:

| replay | Latin (n=200) | non-Latin (n=38) |
|---|---|---|
| same-source | 100.0% | 100.0% |
| cross-source | **100.0%** (was 89.5%) | 0.0% (unchanged, expected) |

Every remaining cross-source miss is in the tier-2-unreachable stratum — 武備志 volumes, Cyrillic collected works, plus short-title edge cases like "Hobo" (4 chars, under tier 2's length guard; the stratifier classifies by reachability, which is the right boundary). The non-Latin zero is now the *only* recall hole in import dedup, and it has a named fix: the edition-key tier (#3730 §2), whose shadow measurement is the next piece of work.

Note the non-Latin sample is 38, not 200 — the tier-2-unreachable population among *visible* books is small (most of the 20K unreachable books are hidden/warehouse). 0/38 is still unambiguous.

PR #3804, opened 2026-08-09
docs(invariants): blank leaves aren't translations, and published is free text

JDerekLomas:
Step 4 of `/gnite` — the upward ratchet for two bugs fixed and merged today.

Both go in `visibility-and-stats.md` rather than a new file: they are stats-correctness rules, and that doc already owns the canonical readable denominator `pages_ocr − pages_blank` that both bugs corrupted. **No `CLAUDE.md` change** — the routing entry already exists, and CLAUDE.md is at 222 lines against its ~250 budget.

**#3747 — the numerator must exclude what the denominator excludes.** Blank pages are subtracted from the denominator but carried the placeholder `[Blank page — no translatable content]`, so a non-empty check counted 87,777 flyleaves as translated work and pushed `translation_pct` over 100 on 6,228 live books (32% of the public library; the Blue Qur'an read 1000%).

The reason this is worth a doc entry rather than just a memory: **it recurred after 136 days.** The existing note stated the naive rule — "`pages_translated` = pages with actual `translation.data`" — as correct. The guidance wasn't missing, it was wrong, and wrong guidance is trusted. I corrected that note in the same pass.

**#3718 — `books.published` is free text.** `parseInt("12th century") === 12`, and `sort: { published: 1 }` is lexical so "550" follows "1750". Points at `editionYear` / `formatYearSpan` / `byChronology`, and flags `src/app/languages/page.tsx` as still carrying it.

Also pins the composition-date discriminator, including the tell that looks right and isn't: "published is Unknown" does **not** identify them — the Bodleian Gospels manuscripts at 895/927/950 also lack a publication statement and their dates are genuine.

Docs only; no code paths touched.


JDerekLomas:

Issue #3778, opened 2026-08-08
FT verification at scale: escalation ladder — memory bank + Gemini skeptics, Claude/human only for the flagged tail


JDerekLomas:
Design learned from the 2026-08-08 full-queue verification (#3687, #3726). Goal: run Stage-2 verification mostly on cheap Gemini agents + accumulated memory, reserving Claude subagents / humans for the residue where they're irreplaceable.

## The memory bank (3 layers exist, 1 missing)

| Layer | Store | Role |
|---|---|---|
| Verified positives | `translation_catalogs` (24K, +52 on 08-08) | every verified prior turns future checks into free membership tests |
| Catalogue absence | `reference_translations` (149K, recall 32%) | the bounded search we cite |
| Search history | `first_translation_attempts` (62.7K) | never re-run a search already run (`latestEffortPerBook`) |
| **MISSING: casebook** | docs/invariants prose today | the failure-mode detectors + policy rules + tradition-specific rules (CJK reachability, Kabbalistic title collisions, container scoping), machine-readable, injected as rubric |

The demote screen's five detectors (no_named_translator, prior_is_a_study, container_title, witness-vs-work, amalgamated-citation) predicted 13/16 verified outcomes with ZERO searches. They are the routing function.

## The ladder

1. **Registry lookup** — free. Prior already verified → done.
2. **Gemini grounded skeptic** (~$0.005/book) — same strict JSON contract as ft-verify (verbatim queries, per-source found-lines, structured priors with MANDATORY pub_year, refute-framing). Handles the easy 80%: does the citation exist, complete vs excerpt.
3. **Claude subagent** (subscription) — only for detector-flagged hard classes (work-identity disambiguation, container scoping, our-manuscript content-reading via get_book) and rung-2 disagreements. Cross-family independence is the point — Gemini checking Gemini re-correlates blind spots (measured: ~63% fabricated priors in early Stage-1).
4. **Human** — physical-copy checks (Cardano/Holden class), policy questions (practitioner PDFs).

## Non-negotiables carried over

- **Ingest is actuation** (#3776): evidence rows feed nightly derive+reconcile. Rung-2 output writes at catalog-tier trust; only rung-3+ earns the resolver tier that can move a badge.
- **prompt_version stamping** on attempt rows (July replicability gap, still open) — required before any automated loop.
- Persist rung-2/3 transcripts (or at least the full tool-call list), not just the structured extract.
[...]

PR #3805, opened 2026-08-09
fix(ft): rung-2 skeptic fails closed on ungrounded responses — two-phase prose-then-format (#3778)

JDerekLomas:
Follow-up to #3783, from the first at-scale ladder run (Derek-approved $10 budget).

## What run 1 found

634 books in, every attempt row had **zero API-level grounding metadata** while the model's JSON self-reported plausible `queries_run` — `gemini-3.1-flash-lite` was answering from memory and inventing the search trail. The run was halted; all 634 rows were deleted with Derek's approval (backup: `scripts/output/ft-ladder-run1-attempts-backup-2026-08-08.json`; transcripts and usage rows kept as the true record). The one thing that contained the damage: the contract computes coverage from API metadata, never the model's self-report, so 498 of the bogus absences were already inert to derive. The remaining hole — 109 "found" rows graded moderate because their priors carried year+URL — is exactly what this PR closes.

## Diagnosis (measured, not theorized)

- v2 added an explicit "you MUST search" mandate: **0/10 grounded**. Not an instruction problem.
- Matrix probe: compact prompt 0/3, search-first imperative 0/3, **prose output 3/3** (11–13 grounding chunks). The strict-JSON output demand itself suppresses tool use on lite.
- Preview model grounds 10/10 but costs ~$0.113/book (thinking tokens) — $590 for the queue.
- Lite prose grounding is stochastic: 8/12 per call. Two cheap retries → ~96%.

## The fix

- **Skeptic v3 is two-phase**: a grounded prose research call (tools on, no format demand), then a tools-off formatter into the JSON contract, instructed to add nothing.
- **Fail closed**: a response with zero grounding is recorded as `gemini_verifier` (model-knowledge family), weak, flagged `ungrounded_response` — never as a grounded search. Without grounding, a URL is just another token.
- **Retry the cheap rung**: up to 2 primary-model retries before any fallback model.
- Tests pin both: the ungrounded fail-closed mapping, and that an ungrounded-found-only ledger derives nothing.

Full run relaunched under this code (v3 prompt stamp, $8 cap, fallback off — the ~4% residual fails closed into the free rung-3 Claude queue). Projected ≈$4 for all ~5.2K books.

### DIFF
--- a/scripts/eval/ft-ladder.ts
+++ b/scripts/eval/ft-ladder.ts
@@ -1,0 +1,376 @@
+/**
+ * ft-ladder.ts — the FT verification escalation ladder (#3778).
+ *
+ * Runs each queued book up the cheapest rung that can settle it:
+ *
+ *   rung 0  history      — the attempt ledger already answers → skip (free)
+ *   rung 1  registry     — deterministic tier-0 catalog match → log it (free)
+ *   rung 2  skeptic      — grounded Gemini refutation attempt (~$0.005–0.03/book)
+ *   rung 3  claude queue — hard classes + rung-2 residue, emitted as a queue
+ *                          file for the ft-verify skill (NOT executed here)
+ *   rung 4  human queue  — policy holds (practitioner PDFs etc.), emitted
+ *
+ * ACTUATION (#3776): with --apply this script writes to
+ * `first_translation_attempts`, which the 05:30 nightly derive+reconcile reads.
+ * Every row written here carries a method that resolves to tier1_catalog or
+ * tier0_linked — resolvers the reconcile valve (--resolver=tier2_agent,human)
+ * does NOT admit — so nothing this script writes can move a public badge.
+ * The badge-moving tier is rung 3+ (ft-verify → ingest-ft-verify-results.mjs)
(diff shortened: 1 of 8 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek**, leaning close. The fix is unlanded (`skeptic.ts` on main is still the #3783 base, no two-phase path), but the architecture moved under it: #3885 (one ledger, one queue), #4524, and #4552 deleted the book-grain badge machinery; the file this PR targets (`derive-from-evidence.ts`) no longer exists — main has `derive.ts`. #3858 separately landed a model-aware rung-0 skip for the same ungrounded-response finding. Question for you: is the rung-2 grounded skeptic still a live path? If not, close; if so, this needs a rebase.

JDerekLomas:
Closing (backlog decision 2026-09-22): targets derive-from-evidence.ts which no longer exists (main has derive.ts), and #3885/#4524/#4552 replaced the ladder architecture; #3858 landed the model-aware rung-0 skip for the same ungrounded-response finding. Branch kept.

PR #3841, opened 2026-08-09
fix(dedup): artwork revalidate paths are slug-keyed, not id-keyed

JDerekLomas:
Follow-up to #3836, found while executing the approved lanes: the `/artwork` route is `[slug]`-keyed and 404s on an id (verified live), so `pagePath()` was emitting no-op revalidate paths for artworks. Fixed to use `slug || id` for anything with `resource_type`/`content_type: 'artwork'`; books stay `/book/[id]`.

The 603 affected artwork pages from today's runs were re-revalidated by slug out-of-band (all 200 now), so this fix is for future runs, not a repair.


JDerekLomas:
Recreated-branch CI trap: test workflow never fires on this branch name (close/reopen + new push both failed to trigger it). Superseded by a fresh-branch PR.

Issue #3823, opened 2026-08-09
Parsing reader: click-a-word Latin dictionary lookup (Lewis & Short + LatinCy)


JDerekLomas:
# Parsing reader: click-a-word Latin dictionary lookup

## Why (the human purpose)

The mission is putting primary sources into people's hands — and the single biggest barrier to *reading* our originals is vocabulary. A reader opens a 1617 alchemical treatise, hits `sublimatio` or `menstruum`, and today has to leave the page to find out what it means. The translation pane helps, but it answers "what does this passage say," not "what is this word" — and the readers we most want to serve (students, scholars, the Latin-curious) are exactly the ones trying to engage the original text word by word.

Prior art: https://ancientlibrary.net/ (Worth Media) built its entire site around this one feature — click any word in a classical text for its lemma, morphology, and the full Lewis & Short entry. It works, it's delightful, and every ingredient is free. We can adopt the idea for our corpus, where it's both harder (OCR of early modern printing, not clean critical editions) and more valuable (our texts mostly have no published translation at all).

## What the reader gets

In the book reader, on original-language Latin pages: tap/click any word → a small popover shows

1. **Lemma** (dictionary headword) + basic morphology ("ablative singular of `menstruum`")
2. **The Lewis & Short entry** (collapsed to the headword + first sense, expandable)
3. A link to the full entry page

No mode switch, no leaving the page. On mobile, the popover becomes a bottom sheet.

## Ingredients (all free)

| Piece | Source | License |
|---|---|---|
| Lewis & Short, *A Latin Dictionary* (1879) | Perseus XML (`lexica` repo, ~51K entries, TEI) | CC BY-SA / public domain text |
| Lemmatization + morphology | [LatinCy](https://huggingface.co/latincy) (spaCy pipeline) | MIT |
| Fallback lookup | Whitaker's Words (rule-based, handles medieval/early-modern forms) | Public domain |

Greek (LSJ + odyCy) is the same shape but out of scope for v1 — Latin is ~an order of magnitude more of our corpus.

## The hard part: our Latin isn't Perseus Latin

ancientlibrary.net parses clean critical editions. We parse OCR of 16th–17th c. printing:

- **Orthography**: `æ/ae/e`, `u/v`, `i/j`, `ß/ss`, long-s OCR artifacts (`ſ` → `f` errors)
[...]

PR #3851, opened 2026-08-09
feat(reader): tap-a-word Latin dictionary popover (#3823 Phase 2, stacked on #3840)

JDerekLomas:
Phase 2 of #3823 — the reader UI on top of the lookup API from #3840. **Stacked on #3840**; merge that first (this branch contains its commits until then, after which this diff reduces to two files).

## What readers get

On any Latin book's reader page, with the original-text pane open: tap a word → a popover shows the dictionary form (with vowel lengths), a morphology line (genitive, gender, declension), the first Lewis & Short sense, and a "Full entry" expansion. Desktop gets a floating card anchored at the tap; mobile a bottom sheet. Uncertain matches (suffix-heuristic / OCR-repair tiers) carry an explicit "treat as a suggestion" hedge, and a miss says honestly that abbreviations and contracted forms often miss.

Verified live on Drebbel's *Tractatus duo* (1628): tapping `ſubtilitate` — long s, straight off the page — resolves to *subtīlĭtas, gen. ātis, fem., 3rd decl.,* "Fineness, thinness, slenderness, minuteness," full 4-sense entry expandable.

## How (and the two traps from the issue)

- **Zero DOM mutation.** No per-word `<span>` wrapping: a click handler reads the tapped word out of the text node via `caretPositionFromPoint`. OCR text that transcribes literal markup (#3609) therefore cannot break the page, and `NotesRenderer`'s rendering is untouched.
- **Language gate is the edition's, not the work's.** The OCR pane always displays edition text, so `book.language === 'Latin'` is the correct gate (per the language-field lesson); disabled for paired Greek overlays and edit mode.
- Click delegation is document-level because the OCR panel mounts/unmounts as readers toggle panels (panel-bound listeners miss a panel shown after mount — found this live).
- Text selections and clicks on interactive elements are ignored; Escape/outside-click dismiss; ~300-entry client cache in front of the CDN-cached API.

## Out of scope

- Greek (LSJ), lemma-layer search, and linking entries to a scanned Lewis & Short (#3823 Phase 3).

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -117,14 +117,17 @@ const nextConfig: NextConfig = {
             value: [
               // Next.js requires 'unsafe-inline' for styles and 'unsafe-eval' for dev; in prod we need 'unsafe-inline' for React inline styles
               "default-src 'self'",
-              "script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: data: https://translate.google.com https://translate.googleapis.com https://www.googletagmanager.com https://analytics.ahrefs.com https://eu-assets.i.posthog.com",
-              "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://translate.googleapis.com",
+              // PROTOTYPE (#3823): jsdelivr allowed for the Alpheios embedded
+              // reading-tools evaluation. A production ship vendors the
+              // bundles and removes these entries.
+              "script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: data: https://translate.google.com https://translate.googleapis.com https://www.googletagmanager.com https://analytics.ahrefs.com https://eu-assets.i.posthog.com https://cdn.jsdelivr.net",
+              "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://translate.googleapis.com https://cdn.jsdelivr.net",
               "font-src 'self' https://fonts.gstatic.com",
               // Images: self + all configured remote image sources + data URIs for base64 thumbnails.
               // Host list lives in src/lib/csp-img-hosts.ts (shared with getBookThumbnailUrl's
               // renderability screen — edit it there, never inline here).
               CSP_IMG_SRC,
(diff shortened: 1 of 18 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek** — a new tap-a-word popover on the public reading page is reader-facing UX. Half of it is already on main: the lookup API, morph/normalize libs, import scripts, and both eval result JSONs landed via #3840, #3883, #3961, so the branch will conflict on 10 of its 18 files and needs a rebase to just the reader layer (`LexiconTapLayer`, `AlpheiosLoader`, `lookup-grc.ts`, the `next.config.ts` Alpheios loader) before CI can run.

JDerekLomas:
Rebased onto current main (7f5d19c7) and reduced to the still-unlanded reader layer. The Latin lookup chain, morph/normalize libs, lexicon import scripts and both eval scorecards landed via #3840/#3883/#3961 — main's version of each is kept, and main's newer `latin-morph.ts` (`irregularEntries`) and `morpheus-crunch.mjs` (import guard) are untouched.

Now in scope (7 files, +644/-11):
- `LexiconTapLayer` + `AlpheiosLoader` and their wiring into the reader's OCR pane
- `src/lib/lexicon/lookup-grc.ts` and the `lang=grc` dispatch on `/api/lexicon/lookup` (without it the Greek library is dead code)
- `scripts/lexicon/rank-grc-lemma-keys.mjs` — the ranking pass the lemma tier honors
- `next.config.ts` CSP: jsDelivr + `*.alpheios.net` for the prototype loader
- shorter browser `max-age` on the lookup route (CDN TTL unchanged)

One change beyond the original diff: main's modernized-English toggle now shares the OCR pane, so `alpheios-enabled` is gated off while modernized text is on screen.

Out of scope, deliberately: vendoring the Alpheios bundles (the loader pulls `@latest` from jsDelivr — prototype only), and turning `LexiconTapLayer` on (it ships `enabled={false}` pending the comparison).

Issue #3419, opened 2026-07-28
gallery_images.type holds raw model output on 99 rows (up to 4KB of reasoning stored as the enum value)


JDerekLomas:
## What

99 rows in `gallery_images` carry a `type` value that is not a type — it is a chunk of the extraction model's raw response, up to **4,152 characters**, including the model's own visible reasoning.

```
gallery_images total                206,401
distinct `type` values                  113
  legitimate enum-ish (<=40 chars)       16
  response blobs (>40 chars)             96
rows carrying a blob type                99   (0.048%)
```

One real value, verbatim and truncated:

> `diagramBase64: diagram; bbox: [0.01, 0.03, 0.98, 0.94] (covers all three leaves) - wait, the type must be exactly one of the list. I will use 'diagram'.`
>
> `Actually, I will provide the bbox as requested. x: 0.01, ...`

The model deliberated about which enum value to use, and the deliberation was stored *as* the enum value.

Three shorter malformed values sit in the same family and are easy to miss because they look almost valid:

```
"diagram,"
"diagramBase64EncodedImage:null,"
"diagramBase64EncodedImageData:null,"
```

## Why it matters

This is the **reasoning-as-output** class that `CLAUDE.md` already documents for OCR (`pages.ocr.data` holding `-> wait, ... is on the same line as`, PR #3273), appearing in a different writer. The lesson there applies here: a field that is *supposed* to hold one of N enum values needs validation at write time, because the model will occasionally narrate instead of answer, and nothing downstream distinguishes narration from a value.

Blast radius is small (0.048%) but non-zero, and these rows render their `type` in gallery surfaces. It also silently corrupts any `distinct('type')`-based UI or analytics — the type facet has 113 entries where it should have ~14.

## Scope

1. **Write-time validation** in the image-extraction writers: reject anything not in the enum rather than storing it. Live prompt is inline in `scripts/workers/image-extract-worker.mjs`, `scripts/workers/pipeline-orchestrator.mjs`, and `src/lib/image-extraction.ts` — same three-place duplication called out in the quality-systems section of `CLAUDE.md`, so a fix must land in all three.
2. **Repair the 99 rows.** Most blobs contain a recoverable intended value (`Type: diagram.` appears inline in several), so this is mostly parseable rather than needing re-extraction. Anything unparseable → `unknown`.
3. **Normalize the three malformed near-values** to `diagram`.
[...]

PR #3900, opened 2026-08-11
fix(scripts): two NUL bytes made a source file binary and silently disabled a guard

JDerekLomas:
Small but worth reading, because the second consequence hid the first.

`normalizeForMatching` in the gallery-type repairer protects `musical_score`
across the underscore-flattening pass by swapping it to a spaced form and back.
Both string literals held a raw `0x00` byte instead of a space, so the round trip
was **asymmetric**: the value went in as `musical\0score` and the restore looked
for `musical score`, never matched, and left the sentinel in place.

## Two consequences

**1. The guard silently did nothing.** No repair outcome actually changed — the
one affected row resolved through the run-on-field rule on the prefix before
`Status`, independent of this path. But that was luck, not design. Re-ran the
repair after fixing: 0 new changes, the same 4 rows still honestly undecidable.

**2. Git classified the file as binary**, so PR #3898 rendered it as
`Bin 6338 -> 8337 bytes` instead of a diff. The most important change in that PR
— stopping a re-run from overwriting `type_raw` and destroying the only copy of
the original model output — **was unreviewable**. `git grep` couldn't see the
file's contents either, which is how I nearly concluded the merge had dropped the
change (a plain `grep` for a string that was present printed nothing).

## What I deliberately did NOT add

A repo-wide "no control bytes in source" check. I scanned `src/` and `scripts/`
first: three other files contain them, and **all three are legitimate**.

| file | byte | use |
|---|---|---|
| `src/app/api/ngrams/route.ts` | U+001F | composite key joiner |
| `scripts/eval/lib/metrics.mjs` | NUL | unit-sequence joiner |
| `scripts/maintenance/additive-mint-authors-3780.mjs` | NUL | composite key joiner |

Each uses its delimiter **symmetrically** — join and split with the same
character — which is the standard technique for a joiner that cannot occur in the
data. A guard firing on three correct files and one wrong one is noise, and
`CLAUDE.md` already records where that leads: a noisy safety check doesn't fail
closed, it trains people to bypass it (the `--force` habit on the worktree
reaper). The tell for the real defect is the **asymmetry**, not the byte, and I
don't have a cheap reliable check for that.

Refs #3419


JDerekLomas:
Superseded by a PR from a fresh branch — GitHub Actions would not fire `unit-tests` for this branch after it had already been merged once (no run recorded for either pushed SHA, and neither close/reopen nor an empty commit triggered one). Reopening from a clean branch off main rather than merging something CI never saw.

Issue #3597, opened 2026-08-04
pdf-facsimile will time out and/or OOM on most of the corpus (measured: 252s / 707MB at 366 pages)


JDerekLomas:
## Summary

The July handoff noted *"End-to-end 300-page facsimile PDF never load-tested against prod."* It has now been load-tested locally, at 366 pages. **It does not have the headroom to serve most of the corpus.**

## Measurements

Kitāb al-Bulhān (`6953b56577f38f6761bd979d`), 366 pages, all with images. Replicating the route's own pipeline (`fetchPageImagesOrdered` concurrency 8 → sharp resize/grayscale/JPEG q75 → pdfkit):

| Phase | Result |
|---|---|
| Mongo fetch | 6.3 s |
| Image fetch + compress (366, conc. 8) | **230.9 s** |
| Buffered image bytes | 74.3 MB |
| RSS after fetch | 311 MB |
| PDF render | 15.3 s |
| **Total** | **252.5 s** |
| **Peak RSS** | **707 MB** |
| Output | **75.9 MB**, 865 PDF pages |

`maxDuration = 300` on the route. So a 366-page book completes with **~48 s of margin**, and peaks at 707 MB against Vercel's **1024 MB** default.

## Why this is not an edge case

| Cohort | Count (visible) |
|---|---|
| > 366 pages | **6,987** |
| > 800 pages | **1,156** |
| Largest | **4,198 pages** |

Nearly seven thousand visible books are larger than the one that already uses 84% of the time budget. Linear extrapolation puts an 800-page book around 500 s and ~1.4 GB — over `maxDuration` *and* over the memory ceiling.

## Two structural problems, not just "it's slow"

**1. Nothing streams until every image is fetched.** `fetchPageImagesOrdered()` awaits *all* images before the first page body is written. The title page flushes, then the connection is silent for ~231 s. The route is described as "streamed … (CF ~100s window)", but streaming does not help when the producer blocks for four minutes before emitting anything. Cloudflare will drop this well before `maxDuration` does.

**2. Peak memory scales with page count.** All compressed images are held simultaneously (74 MB) *plus* the pdfkit document (~76 MB output) plus overhead. Memory is not network-dependent, so unlike the fetch time it will not improve on Vercel.

## Caveat on the numbers

[...]

Issue #2709, opened 2026-06-22
Reader: visually distinguish AI editorial annotations from the book's own text


JDerekLomas:
## Problem
In the reading view, AI-written editorial annotations render indistinguishably from the book's actual transcribed/translated text. A reader can't tell what the **source** says vs. what **we** added.

Concrete example: https://sourcelibrary.org/book/the-reformed-and-enlarged-hermetic-museum-anonymous/page/6984727b8e424cf61970feff — the translation contains:
- `<meta>` — AI page summary ("This page serves as the opening theological frontispiece…")
- `<note>` — AI **describing** Figure 1, the celestial Trinity diagram ("At the very center is a triangle containing the Tetragrammaton…")
- …interleaved with the real transcribed text.

The `<note>` reads as if it were the book's words, when it's actually the AI describing an illustration.

## Why it matters
This is the reading-view analog of the **quote-integrity doctrine** (CLAUDE.md "Quote & snippet integrity"; PRs #2232/#2233). Editorial wrappers *describe* the page/scan and routinely name content from adjacent pages — they are **never verbatim source**. We already strip them from quotes/snippets/search/IIIF; the reading view should clearly **label** them, not blur them into the prose. Showing AI commentary as if it were the original undercuts the core "read and quote the original" promise.

## Current state
- `src/components/reader/NotesRenderer.tsx` (`extractMetadata`) already parses these tags and pulls some into a metadata object (`meta`, `summary`, `keywords`, `vocabulary`, `pageType`, …), gated by `showMetadata` / `showNotes` props — but the rendered result doesn't visually differentiate them enough, and inline `<note>` figure-descriptions in particular read as body text.
- `src/lib/strip-editorial-wrappers.ts` holds the canonical taxonomy:
  - **Editorial** (AI describes page/scan, NOT source): `meta, summary, keywords, vocab, language, scan-quality, script, page-type, columns, warning`
  - **Inline glosses** (sit on real body text): `note, term, margin, gloss, unclear, insert`
  - **Real printed marks** (keep as text): `header, catchword, sig, page-num`

## Proposed approach
1. Render the three categories with clearly distinct treatment in the reader (and split view):
   - Editorial page-descriptions → a **labeled, set-off block** ("Curator's note" / "About this page" — tinted, clearly not prose).
   - Inline glosses → visually marked as annotation (muted/italic, icon, or margin treatment) — unmistakably "ours, not the book's."
[...]

PR #3911, opened 2026-08-11
feat(downloads): facsimile PDF as facing-page spreads with colored annotations

JDerekLomas:
## Why

Reader feedback via /support (2026-08-10, a signed-in reader finalizing work with our PDFs): *"the layout is off. Are you able to make the required changes so they export as they are displayed on the website: One page with the scan beside one page of translation. I also like the colour coding of text when you are providing context within the translation - can we keep that?"*

Both asks are exactly right — the old `pdf-facsimile` squeezed a 600×900 **grayscale** (EPUB-sized) scan into the top 60% of each page with translation text crammed beneath, and flattened every annotation to plain black text.

## What

**Facing-page facsimile.** Each book page becomes a spread: full-page **color** scan (1600px, ~260 DPI) on the left, translation on the right. Blank filler pages keep parity when a translation overflows (the Loeb problem), and the PDF catalog's `/PageLayout TwoPageRight` opens the title page alone like a cover, then scan-left/text-right in any viewer's two-page mode.

**Annotation colors.** `note`/`image-desc` (gold italic), `margin`/`insert` (sage), `gloss`/`term` (violet), `unclear` (gray italic + ?) — same palette as the reader's `NOTE_TAG_STYLES`, in both `pdf-facsimile` and `pdf-translation`. Spans are normalized via `normalizeAnnotationSpans` (#2709) before run-parsing; legacy `[[bracket:]]` syntax converts; table cells unwrap tags (no color in cells).

**Refactor for testability.** Both generators moved from the two route files into `src/lib/pdf-export.ts` with image streaming injected per-route (routes keep their fetchers, per the existing pattern). The tenant route now streams images with bounded lookahead instead of buffering all of them (same #3597 rationale as the main route).

**QA harness.** `scripts/qa/render-book-pdf.ts` renders either format locally against prod data — reproduce a reader report or inspect a change visually in one command, no auth/deploy needed.

## Verified by rendering

- 16pp Latin disputation (1577): title page, spreads pair correctly, parity filler after a long translation, gold/violet/sage runs, colophon.
- 366pp Kitab al-Bulhan text PDF: Arabic script runs shaped correctly, GFM tables as real PDF tables, 441 output pages, no crash.
- Unit tests extended for run parsing (style across newlines, unclear '?', attributes, cell unwrapping) — 24 pass; `tsc --noEmit` clean.

## Out of scope

- EPUB facsimile keeps its 600×900 grayscale variant (e-reader constraints).
- No new download format keys; `pdf-facsimile` changed in place.

### DIFF
--- a/scripts/qa/render-book-pdf.ts
+++ b/scripts/qa/render-book-pdf.ts
@@ -1,0 +1,138 @@
+/**
+ * Local PDF-download QA harness.
+ *
+ * Renders the SAME generator code the download routes serve
+ * (src/lib/pdf-export.ts) against real production data, without auth or a
+ * deploy — so download-format changes can be inspected visually before they
+ * ship, and regressions reproduced from a reader report in one command.
+ *
+ * Usage:
+ *   set -a; source .env.production.local; set +a
+ *   npx tsx scripts/qa/render-book-pdf.ts <bookId> [pdf-facsimile|pdf-translation] [outPath]
+ *
+ * Output defaults to scripts/output/<bookId>-<format>.pdf (gitignored scratch).
+ * Open the PDF in a viewer that supports two-page view (Preview: View →
+ * Two Pages) to check facsimile spreads, or let Claude Read it page by page.
+ *
+ * The image path mirrors the route: canonical resolver (getPageImageUrl,
+ * display variant) → fetch → sharp resize/compress → ordered bounded stream.
+ */
+import { MongoClient } from 'mongodb';
+import sharp from 'sharp';
+import { createWriteStream, mkdirSync } from 'fs';
+import { dirname, join } from 'path';
+import {
+  generatePdfFacsimileStream,
+  generatePdfTranslationStream,
+  type PdfBookInfo,
+  type PdfExportPage,
+} from '../../src/lib/pdf-export';
+import { getPageImageUrl } from '../../src/lib/page-image-url';
+import { streamOrdered } from '../../src/lib/ordered-stream';
+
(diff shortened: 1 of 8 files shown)


JDerekLomas:
Heads-up from an adjacent merge — this PR will need a rebase, and the conflicts are structural rather than textual.

#3914 (merged `5929cbd6`, ~13 min after this branch's last push) restructured both download routes that this PR touches:

- The page-image machinery moved out of **both** routes into **`src/lib/export-page-images.ts`** — `fetchAndCompressImage`, `pageExportImageUrl`, `hasExportImage`, `fetchPageImagesOrdered`, `streamPageImagesOrdered`, `generateImagesZipStream`. Roughly 100 lines left each route. If this branch edits any of those in place, the edit needs to move to the lib (where it now applies to both routes at once, which is probably what you wanted anyway).
- **`generatePdfFacsimileStream` in the tenant route changed substantially**: it now streams via `streamPageImagesOrdered` instead of awaiting every image, and carries the time budget + truncation notice + `partial facsimile edition` colophon label the global route already had. Since this PR reworks facsimile PDF layout, that function is the likely conflict site.
- `markdownToHtml` also moved to `src/lib/export-markdown-html.ts` in #3908 earlier today.

Also worth knowing before you rebase: `tests/unit/download-route-parity.test.ts` now guards that neither route resolves images from raw `photo`/`compressed_photo` or redefines those helpers locally. If the rebase reintroduces either, that test fails by design — it is not a flake.

No action needed from me; flagging so the rebase isn't a surprise.

Issue #2185, opened 2026-05-30
Importers set language/date from caller input, ignoring source metadata — flip precedence + reconciliation gate


JDerekLomas:
## Problem

All six import routes set a book's `language` (and date) from **caller-supplied input first**, with the source's own structured metadata ignored or used only as a last resort. This lets a wrong caller value silently override what the source actually says — the root cause behind #2184 (Pico's *Oratio* imported as "Latin / 1486" when the scan is a 20th-c. Russian translation; IA's `metadata.language: rus`, `booksbylanguage_russian` collection, and `ocr_detected_lang` were all fetched and none gated the result).

This is an **engineering/precedence fix**, split out from the diagnosis + data remediation in #2184 (one concern per PR).

## Current state (still live as of `main`)

| Route | Line | language resolution |
|---|---|---|
| `src/app/api/import/ia/route.ts` | 253 | `language \|\| original_language \|\| iaMetadataRaw.language?.[0] \|\| iaMetadataRaw.language \|\| 'Unknown'` |
| `src/app/api/import/route.ts` | 98 | `language \|\| 'Unknown'` |
| `src/app/api/import/google-books/route.ts` | 170 | `language \|\| 'Unknown'` |
| `src/app/api/import/iiif/route.ts` | 494 | `language \|\| 'Unknown'` |
| `src/app/api/import/gallica/route.ts` | 159 | `language \|\| 'Unknown'` |
| `src/app/api/import/kyoto/route.ts` | 49 | `language \|\| 'Japanese'` |

- IA's route is the **only** one that consults the source at all — and even there the source is dead last behind caller input.
- Five of six have **no source-language signal**; `kyoto` defaults to a hardcoded `'Japanese'`.
- `ia/route.ts:253` was last changed 2026-03-17 (commit `9148eb21`, which *added* the caller aliases). Documented as the live path in the admin system-map, `/about/sources`, `api-client/import.ts`, and batch scripts.
- Dates have the same shape: caller-supplied `year`/`published` win over `iaMetadataRaw.date` (the Pico caller passed 1486; IA had no date).

## The core insight

In the Pico case the caller's "Latin" wasn't even *wrong* — it's the **work** language. IA's "rus" is the **manifestation** language. The bug is collapsing both into one caller-wins field. The fix is to separate them and let the source decide the manifestation language.

## Fix (checklist)

[...]

PR #3956, opened 2026-08-12
fix(mcp): surface edition language distinctly from work language (#3942)

### LINKED ISSUE #3942: Catalogue language vs page content mismatches: Hebrew collection tagged German; Ibn Khaldun 'original' serves de Slane's French
From MCP feedback 2026-08-11 (quote-harvest session report, item 4). Related: #2184 (scanned translations mislabeled as original language), #3742 (import original_language wiring), and the standing lesson that books.language is the EDITION language, not the source language.

Two confirmed cases:
1. **Hebrew Kabbalistic Collection** (69b418ae2b0edf3eaa2dd9aa) — catalogued as German, but p. 70 is Hebrew.
2. **Ibn Khaldun, Prolégomènes** (69ef3e3a72c1376fdf2b0d3d) — catalogued as Arabic, but the `original` field returns de Slane's FRENCH translation. This one matters most: a researcher citing it for non-European sourcing is unknowingly quoting English←French←Arabic.

Fix the two records, but the structural ask is the reporter's: **surface "original language of this edition" distinctly from "language of the work"** in MCP responses (get_book, get_quote), so a translation-of-a-translation is visible in the apparatus instead of silently flattened. The text_role / original_language machinery exists; the MCP surfaces don't expose the distinction.

Mayank-PPL:
Closes #3942.

## The defect

`books.language` is the **manifestation** language — what is printed on the leaves of this scan (the contract set by `resolveLanguage()`, #2185). `original_language` names the **work** when the two differ. Every MCP surface served the first scalar under the name `original_language`, and the tool tips asserted it was "the book's source language."

For a translated edition that is exactly backwards. de Slane's 1863 French *Muqaddimah* returned `original_language: "Arabic"` beside a page of French, so a caller citing it for non-European sourcing was quoting English←French←Arabic with nothing in the response saying so. That claim was generated by the response shape, so it held for every translated edition in the corpus regardless of how clean the record was.

The two reported records are the visible symptom. The flattening is the bug.

## Evidence for the record fixes

The OCR model declares a language per page (`<language>…</language>` inside `ocr.data`) — the same signal `english-source-detect.mjs` uses, and the only one that reads the leaf instead of the catalogue:

| Book | Catalogued | Page evidence | Now |
|---|---|---|---|
| Hebrew Kabbalistic Collection `69b418ae…` | German | **328/422 Hebrew**, 1 German | `language: Hebrew`, `text_role: original` |
| Ibn Khaldun, *Prolégomènes* `69ef3e3a…` | Arabic | **450 French-only, 145 French+Arabic, 0 Arabic-only** | `language: French`, `original_language: Arabic`, `text_role: modern-translation`, `is_translation: true` |

The Hebrew record had already been flagged by an automated detector on 2026-08-09 (`language_review_detail: {detected: "Hebrew", current: "German", confidence: "high"}`) and nobody acted on it. **1,520 live books carry an untriaged `language_review: true`.**

## What changed

- **`src/lib/edition-language.ts`** (new) — `languageApparatus()` derives `edition_language` / `work_language` / `text_role` / `translation_note` from a book record. Fields are *absent* rather than null when there is nothing to disclose, so a caller never renders "null" beside a citation.
- **`/api/books/[id]` nav projection** — now carries `original_language`, `text_role`, `is_translation`. It carried none of them, so the distinction could not reach the wire even with the data correct. This was the actual blocker.
- **`get_book`** returns the apparatus; **`get_quote` / `get_quotes`** add a citation warning when the page's `original` field holds a translator's words rather than the author's; the **quote API** serves the same three fields.
- **`search_translations` / `search_concept`** — field renamed `original_language` → `language`, both tips corrected. The search index carries no work language, so the honest fix is to stop claiming one and point callers at `get_book`.
- **`scripts/maintenance/fix-3942-edition-vs-work-language.mjs`** — corrects the two records. Dry-run by default, guarded on the wrong value still being present, writes a backup.
- **`.claude/docs/invariants/first-translation-claims.md`** carried a bullet stating the inverse contract ("`books.language` is the language of the WORK, not of the pages we hold") — which read as licence for this exact mislabel. Corrected, with the real observation behind it kept.

## Actuation, per CLAUDE.md

`language` and `text_role` are read by `sync-books-catalog.mjs` (the Supabase mirror behind every card surface) and feed `textRoleRank` in search ordering. Both records' `source_language_screen` recorded `basis: 'declared_language'` — they trusted the scalar this changes — but neither verdict flips (French and Hebrew are both non-English, so `foreign_source` still holds) and neither book carries `is_first_translation`, so **no public badge moves.**

The second commit is a defect found in the first: `sync-books-catalog.mjs` runs incrementally on `updated_at > lastSync`, so a correction that doesn't bump the watermark never reaches `books_catalog` — the card surfaces would have kept serving "German" and "Arabic" indefinitely while Atlas held the fix. Same family as the archiver that made its own repair path unreachable.

## Verification

- 9 new unit tests on the apparatus helper; `npx tsc --noEmit` clean; lint clean (the two `no-explicit-any` errors in `books/[id]/route.ts` are pre-existing on `main`).
- Records applied to production and the catalog sync re-run. `books_catalog` now reads Hebrew / French (`text_role: modern-translation`).
- Local `/api/books/[id]` and `/api/books/[id]/quote` verified end to end against both records — the French edition returns the full chain, the Hebrew MS correctly stays silent.
- **The MCP tools could not be verified locally**: `API_BASE` in `src/app/api/mcp/route.ts` is hardcoded to `https://sourcelibrary.org/api`, so a local (or preview) MCP call reads *production's* book API. The new fields appear on MCP only once the API routes here are live on production, i.e. after merge — not on the preview URL. Worth checking on prod after merge.

## Out of scope, deliberately

- **The `text_role` heuristic itself.** `src/lib/text-role.ts:86` short-circuits any non-English scan to `text_role: 'original'`, which is why both records read `original`. **169 live non-English books carrying an explicit translator signal are classed `original`** by it; the work-identity casebook independently measures `period-translation` used 19× against `original` 47,139×. Fixing that means touching a write path that feeds search ranking and FT badges, plus a re-classification sweep, and it needs its own PR and its own review.
- **`/api/books/[id]/editions` `original_language: book.language`.** Same shape, but there "original" means the source of *our* translation — the edition we hold — which is defensible in context, and the field is persisted into minted DOI citation payloads. Not touching it in a fix PR.
- The 1,520 untriaged `language_review: true` records.


Mayank-PPL:
Follow-ups from the out-of-scope section are now filed, so nothing is left only in PR prose:

- #3957 — `text_role` heuristic classes every non-English scan as `original`; 169 live books with a named translator are mislabelled. This is the write path that produced the Ibn Khaldun record.
- #3958 — 1,520 live books carry `language_review: true` with no consumer. The Hebrew MS in this PR was already in that queue, correctly diagnosed, three days before a reader reported it.
- #3959 — the edition citation block's `original_language`, which is defensible in context but travels into minted DOI payloads.

Issue #3951, opened 2026-08-12
Attribution: 24 held verdicts + 9 unresolved identity pairs needing a human


JDerekLomas:
The review queues left by #3894. All are genuinely undecidable from the evidence available to an automated pass — filing so they are a queue rather than lost context.

## A. 24 changing verdicts held because the evidence is not in the title

The Sonnet triage proposed a byline change and the reasoning is real bibliographic knowledge, but the proposed name **does not appear in the book's own title**, so it failed the grounding check. Plausible and unverified is the exact class this issue kept getting burned by.

- **Khunrath's two Frankfurt fair catalogues** (1610, 1614) → Georg Willer / Sigismund Latomus. Khunrath died 1605; the genre attribution is convincing but inferred.
- **Five *Breviarium* volumes** → "Catholic Church". Correct by cataloguing convention — no personal author on the page, and the named popes only *ordered* publication.
- **Hugo de Prato Florido**, *Sermones dominicales* — known incunable attribution, not on the page.
- **Leander van Ess** — a 1827 Greek NT, dated by inference against Tischendorf.
- **"Pyrophilus"** (1736 Hermetic tract) — pseudonym named only in the description.
- **Enimundus Bonefidius**, *Iuris Orientalis* — compiler named in the description.

Full rows in `_tmp-verdict-problems.json` / `_tmp-verdicts-all.json` (session scratch, worth moving somewhere durable if picked up).

## B. 9 identity pairs the merge pass refused

Four are **confirmed different people** and should stay apart — recorded so nobody re-proposes them:

| pair | evidence |
|---|---|
| 李杞 ⟷ 李琪 | *用易詳解* vs *春秋王霸列國世紀編* |
| 陸楫 ⟷ 陸璣 | *古今說海* vs *毛詩草木鳥獸蟲魚疏* |
| 王植 ⟷ 王質 | *皇極經世書解* vs *詩總聞* |
| 許謙 ⟷ 徐謙 | *詩集傳名物鈔* vs *仁端錄* |

Needing a decision:
- **俞琰 ⟷ 兪琰** — variant forms of one character; probably one man, refused because "probably" is not enough.
- **`alfonso` ⟷ `alfonso-2`**, **`persius` ⟷ `persius-2`**, **`s-n`** — resolved differently in the end (see #3894 comments); `johann-michael-faust-2` holds 0 books and can be left.

[...]

Issue #3894, opened 2026-08-11
Printer-as-author: the Aldine press is catalogued as the author of 41 books it only printed — and ai_metadata.author already knew


JDerekLomas:
Reported via MCP `submit_feedback` (feedback `6a7ac2d930665fc3449c1071`, 2026-08-11), from a research session compiling definitional passages on "magic". One record was named; it turned out to be the visible tip of a cohort, and there is an already-computed signal that finds the rest.

## The reported record (FIXED)

`6a09712fbf2196cdca905bb0` — *Demonomania de gli stregoni* (1592, Italian) was catalogued with `author: "Manuzio, Aldo"`. The Aldine press printed it; **Jean Bodin wrote it**, and the record's own title says so verbatim: *"Di Gio. Bodino francese tradotta dal K.r Hercole Cato."*

Corrected 2026-08-11 (Mongo write, keyed on this book id only, per the "key the fix on whatever unit was actually verified" rule in `.claude/docs/invariants/author-identity.md`):

| field | before | after |
|---|---|---|
| `author` | `Manuzio, Aldo` | `Bodin, Jean` |
| `author_id` | `manuzio-aldo` | `jean-bodin` |
| `translator` | — | `Cato, Ercole` |
| `work_id` | `local:n:aldo-manuzio:demonomania-witches` (singleton) | `Q10264553` |
| `text_role` | `original` | `period-translation` |
| `original_language` | — | `French` |

`work_title` / `work_slug` were set to the cluster's **existing** canonical pair (`De magorum daemonomania libri III` / `bodin-de-magorum-daemonomania-libri-iii`) rather than a new mint, so the other-editions rail doesn't split. Catalog synced, paths revalidated, CF purged; the book now renders on `/author/jean-bodin`.

The cost of the error was not cosmetic: we already hold **three other witnesses of this same work** (1581 Latin `69905e022fd6a039938a0803`, 1603 `695925eb15b4070cb1e7fe1f`, 1698 German `69b51e84ff09e4fe943b06f5`), all correctly attributed to Bodin and all clustered on `Q10264553`. The Italian sat in a singleton local work cluster, invisible from Bodin's author page and from its own edition rail — i.e. a printer-as-author error silently costs a work-graph edge.

## The cohort

[...]

Issue #3949, opened 2026-08-12
Attribution: enrichment contamination has a second signature — description-vs-title mismatch


JDerekLomas:
Found by the Sonnet triage pass in #3894.

`author-vs-ai-metadata.mjs` flags `ai_value_suspect` when one `ai_metadata.author` is asserted across books whose catalogued authors are mutually unrelated — it catches a repeated **name**. It misses a second, independent contamination shape: the AI **description describes a completely different book**.

Confirmed cases, none flagged by the current detector:

| book | AI description is actually about |
|---|---|
| Van Helmont, *Philosophia naturalis reformata* | a Dutch tragicomedy about Jacqueline of Hainaut |
| Rothmann, *Chiromantia medica* (1650) | a Varenius geography (coincidentally also 1650) |
| *Theatrum Sympatheticum* ×3 | della Porta's *Magia naturalis* (20-book, optics/agriculture/cryptography) |

## Proposed signal

Compare `ai_metadata.description` against the book's own `title` — shared distinctive tokens, or an embedding similarity if one is already computed. A description sharing nothing with its title is contaminated regardless of whether the author name repeats.

Report-only, same as the existing detector. It should feed the `ai_value_suspect` flag so downstream triage distrusts those rows.

## Why it matters

The whole attribution comparison rests on `ai_metadata` being an independent second opinion. Every contamination shape that goes undetected is a row where a human (or a triage model) is asked to adjudicate between a catalogue value and noise — and the 2026-04-13 Spinoza batch showed those rows read as confident, specific, and entirely wrong.

Related: #3894, `.claude/docs/invariants/author-identity.md`.

PR #3982, opened 2026-08-13
audit(authors): title-page attribution pilots — regex negative, model viable with citations

JDerekLomas:
**A negative result, committed so it's durable.** No behaviour change; adds one read-only audit script that says "don't scale this."

## Why it exists

Derek asked whether I was pulling information from OCR. I wasn't — a long attribution session ran entirely on catalogue metadata, `ai_metadata` and external authorities, and never read a page. Then I measured: **24 of 24** books in the #3951 held-verdict queue have page images *and* OCR. The primary evidence for "who wrote this" was in our own database for every row that had been held as unresolvable, because the grounding check tested the catalogue *title string* rather than the *title page*.

So the plan was: read the title page at corpus scale. Pilot first.

## Design

Control-first, because the question isn't "can it find names" but "when it finds one, is it right".

- **CONTROL** — books already anchored at T4. We know the answer, so agreement rate *is* precision.
- **TARGET** — the ~4,800 books that reach no author page. Reports yield.

Zero model calls; it reuses the existing grammar-based extractor.

## Result: negative

| run | control n | fired | agreed | target n | yield |
|---|---|---|---|---|---|
| first | 120 | 4 | 3 | 120 | 3.3% |
| + preprocessing fixes | 120 | 6 | 4 | 120 | **1.7%** |

**110 of 120 title pages yield no name at all.** The precision figures (75%, 66.7%) sit on 4 and 6 rows — that's noise, and reporting them as precision would be a lie. The measurement that matters is that the extractor fires on ~5% of pages.

The second run **fixed three real bugs and moved nothing**, which is the useful part:

1. Container tags carry attributes, so `<image-desc size="large" …>` leaked its prose — that's how *"Small printer's ornament"* got proposed as an author.
2. Title pages hyphenate across line breaks: `GVER- RA DI NICOLO MACHIAVEL-`.
3. Every rule is case-shaped — `GENITIVE_HEAD` has no `i` flag, so solid caps like `PONTANI` can never match a genitive ending. On mixed-case catalogue titles this is invisible; on page text it's a hard zero.

All three were genuine. None was the bottleneck.

## Why it fails — and this doesn't condemn the concept

The extractor was built for `books.title`: short, mixed-case, already edited by a cataloguer. It's genuinely good there (112 usable rows in #3894). A title page is none of those things — long, ornamented, set in capitals, and **full of names sitting in author-shaped positions that are not the author**. Every disagreement is that failure:

- `auctore` → captured **"Consiliario Aulico Hassiaco"** — Wolff's *office*, not his name
- `di` → captured **"Papa Leone"** (a dedicatee) and **"Illustriss. Signoria di Vinegia"** (a state)

A regex cannot separate an office from a person standing in the same grammatical slot.

## What I'd do next — not done here

Reading a title page is a reading task, so pay a model to read it, and require a **role** and a **quoted line** per name so every proposal carries its own checkable evidence instead of failing silently. Keep this file's control-first shape; the value of the pilot was the control, not the extractor.

Priced off this pilot's real denominator rather than a guess: **~$1–3** on flash-lite, **~$15** on flash-preview, for the whole ~4,800-book target population. That costs money, so it's Derek's call and it is deliberately not in this PR.

### DIFF
--- a/.claude/docs/archive/titlepage-benchmark50-2026-08-17.md
+++ b/.claude/docs/archive/titlepage-benchmark50-2026-08-17.md
@@ -1,0 +1,72 @@
+# Random-50 benchmark, blind adjudication — 2026-08-17
+
+Replaces the convenience-sampled 20-book benchmark, which could not carry a
+significance claim (McNemar p = 0.070 on 18 rows) and was scored against a truth
+table written by the same person who scored it.
+
+## Method
+
+- **Sample**: 50 drawn at random from the 161-book pool (Latin-script, no real
+  byline), **frozen to disk before any page was read** so it could not drift
+  toward whatever the readers turned out to be good at.
+- **Readers**: 50 independent Claude subagents, one book each, no shared context,
+  no knowledge of the other reader's answer — against gemini-3.1-flash-lite's
+  existing proposals from the full corpus run. Same pages, same task.
+- **Adjudication**: only the DISCORDANT rows (the only rows McNemar can use).
+  Each adjudicator saw the pages plus two unlabelled answers, **A/B order
+  randomised per case** (the reader was "A" in 9 of 20). The mapping was held in
+  a separate key file the adjudicator never saw.
+
+## Result
+
+|  | |
+|---|---|
+| scored rows | 48 (2 unparseable) |
+| concordant — both name the same person | 28 |
+| **discordant** | **20** |
+| …reader says nobody, flash-lite names someone | **20** |
+| …the other direction | **0** |
+
+Blind adjudication of the discordant rows (19 returned):
+
+|  | |
+|---|---|
+| subagent correct | **17** |
+| flash-lite correct | 1 |
+| neither correct | 1 |
+
+**McNemar exact, two-sided, on 18 decided pairs: p = 0.00014 — significant.**
+
+On **17 of 19** contested rows the adjudicator found that **the page names no
+author at all** and flash-lite had proposed one anyway. That is the same failure
+class the first benchmark identified, now measured on a random sample with the
+labels hidden.
+
+## What this does and does not establish
+
+**Does**: where the two readers disagree, the subagent is right far more often,
+and flash-lite's dominant error is inventing an author for a book whose front
+matter is genuinely anonymous. Taking the 28 concordant rows as correct — which
+they were not adjudicated to confirm — the implied precision is roughly 94% for
+the subagent against 60% for flash-lite.
+
+**Does not**:
+
+1. **Measure recall.** The pool is "books where flash-lite proposed an author",
+   so flash-lite names someone on every row by construction and can never be the
+   reader that declines. Books where it wrongly stayed silent are not in the
+   sample. The 20-0 discordance direction is therefore partly built in by the
+   frame — what the adjudication adds is *which* reader was right, and that part
+   is not built in.
+2. **Escape model-family bias.** The adjudicator is Claude, as is one reader.
+   Randomised order and unlabelled answers remove position and label bias; they
+   do not remove family bias. A human or third-family judge would be stronger
+   evidence, and the API key needed for a Gemini adjudicator is dead.
+3. **Score the concordant rows.** 28 rows where both readers agreed were never
+   checked. If they share a blind spot, this design cannot see it.
+
+## Cost
+
+~70 subagents at ~57K subscription tokens each ≈ 4M tokens. Not billed, but not
+scalable to 3,905 books either. The two-tier design stands: flash-lite for
+coverage, subagents for adjudication where a relationship judgement is at stake.
--- a/.claude/docs/archive/titlepage-benchmark50-scored-2026-08-17.json
+++ b/.claude/docs/archive/titlepage-benchmark50-scored-2026-08-17.json
@@ -1,0 +1,198 @@
+{
+ "readerWins": 17,
+ "flashWins": 1,
+ "neither": 1,
+ "mcnemar_p": 0.00014495849609375,
+ "detail": [
+  {
+   "n": "01",
+   "title": "Theatrum chemicum, praecipuos selectorum auctorum trac",
+   "subagent": "(nobody named)",
+   "flash_lite": "Lazarus Zetzner",
+   "adjudicator_says": "(nobody named)",
+   "winner": "reader",
+   "confidence": "high",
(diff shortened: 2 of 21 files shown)


JDerekLomas:
## Second commit: the paid pilot. The concept works — the tool was wrong.

Same control-first harness, same two populations, model instead of regex.

| | control fired | agreed | **precision** | target **yield** |
|---|---|---|---|---|
| regex (first commit) | 6/120 | 4 | noise (n=6) | 1.7% |
| **gemini-3.1-flash-lite** | 46/100 | 40 | **87.0%** | **24.0%** |

53 control pages correctly reported **no author named** — that's the model declining rather than filling a slot, which is what the prompt asks for and what the regex could never do.

### Adjudicating the 6 disagreements → real precision ~91%

**Two are the model being right and the catalogue being wrong**, which is the outcome that matters most:

- *Magia adamica* — catalogued `[Vaughan, Thomas]`, page reads **"By Eugenius Philalethes"**. That's Vaughan's own pseudonym; the page is more honest than the catalogue.
- *Mahabharata* — catalogued to `Pratap Chandra Roy; Kisari Mohan Ganguli`, page names **Krishna-Dwaipayana Vyasa**. Those two are the translators.

The other four are **one fixable class**: a name inside a quotation or scriptural citation read as the author — "Paulus" from a sermon's proof-text, "Terentius" from an epigraph, and ten biblical book names lifted straight out of a medieval commentary's citations. Next prompt version excludes names inside quoted or cited matter.

### ⚠️ The flash-preview arm is INVALID — don't read it as a comparison

**44 of 100 responses failed to parse.** Its "100% on 14" describes the calls that survived, not the model. It emits reasoning before the JSON and `maxOutputTokens: 1200` truncates it. Needs a raised cap and an explicit `thinkingBudget` before any model comparison means anything. I'm reporting it as broken rather than as a finding.

### What it actually finds

The founding defect of this whole workstream, with a citation attached:

- `Turrisan, Bernard` (Bernardo Torresano, a **printer**) → **Palladius**, p5: *«PALLADII DIVI EVAGRII DISCIPVLI»*
- `Manuzio, Aldo` → **Cicero**, p10: *«M. T. CICERONIS DE PHILOSOPHIA VOLVMEN PRIMVM»*
- `Unknown` → **Simeon bar Yochai**, p6: *«Collectanea De dictis & gestis R. Schimeon Filii Jochai»*

### Provenance — why this shape, not just a cheaper method

`.claude/docs/data-provenance.md` traces OCR, translation, summary, index and chapters back to a stored prompt with `prompt_id + prompt_hash + prompt_version + source`. **Attribution is not in that chain.** `ai_metadata.author` is a bare assertion with no pointer to what it was read from — which is exactly why #3949 needed a detector to *guess* which enrichment rows were contaminated, and why #3951's 24 verdicts were unresolvable.

Every proposal here carries `book_id + page_number + quoted_line + model + prompt_version + run_at`, and **any proposal whose quoted line is not verifiably on the page is discarded unread**. That guard dropped **25 proposals** in this run — fabricated citations are the failure a "cite your evidence" instruction invites, and it's cheaper to check than to trust.

### Projection

~4,846 target books → ~3.1M input / ~520K output tokens on flash-lite, roughly **1,160 candidates at ~90% precision**, each with a page and a quotable line.

### Also in this commit

- Extracted the page-finding half into `scripts/lib/title-page-ocr.mjs` — the part that worked in both pilots.
- Fixed a structural bug from the first full run: title pages resolve in one pass and **Mongo closes before the model calls**. Interleaving a fast store with hundreds of slow API calls got the idle connection reaped mid-run.
- Transient `fetch failed` now retries. Silent loss shrinks the denominator invisibly and makes a pilot look more decisive than it is.

**Writes nothing to Mongo.** Evidence is a gitignored JSONL.

JDerekLomas:
## Third commit: it was reading the wrong page. Fixed, and it moved everything.

Derek asked whether the picker sometimes thinks a page is the title page when it isn't, and whether it should look at other pages. Measured over 150 books — yes, on both counts:

- **41% carry more than one page tagged `title-page`** — half-title, engraved title, letterpress title. Taking the *first* usually took the **half-title**, which by definition carries a short title and **no author**.
- **18% have no `title-page` tag at all**, so the single pick fell back silently to "first prose page" and reported it as though the OCR had labelled it.

And the title page was never the only attribution surface. A **dedication** is normally signed by the author, a **preface** often is, a **colophon** names printer and author. For incunabula and manuscripts it's decisive — the title page is a 16th-century invention, so for earlier books the incipit and explicit are the *only* attribution there is.

| | v1 (one page) | v2 (window) |
|---|---|---|
| control fired | 46/100 | **61/100** |
| precision | 87.0% | **90.2%** |
| target yield | 24.0% | **27.0%** |
| "names no author" | 53 | **37** |
| quote-rejected | 25 | 50 |

**16 of the "names no author" verdicts were the picker's fault, not the book's.** And precision went *up* at the same time — more evidence, not looser rules.

### The guard got stricter, not just busier

A quote must now appear on the page the model **cited**, not merely somewhere in the window. A citation pointing at the wrong page is worse than none, because it still looks checkable. Rejections doubled to 50. Evidence rows now carry the cited page, that page's **type**, and the full window that was offered — so a reviewer can see what the model was choosing between.

### One more real class: academic disputations

Disputations name two people and they are not equals. The **praeses** presides and is conventionally the author; the **respondens** only defends the theses in the hall (*"defendere conabitur M. Hieronymus Heroldus"*). Without that rule the defender was being proposed as the author, and this corpus is full of disputations.

### Adjudicating the 6 disagreements → ~93% real precision

**Two are the model right and the catalogue wrong:** an Aldine volume catalogued to `Manuzio, Aldo` where the page names **Ovid**; and More's *Luciani Opuscula*, where **Lucian** wrote the Opuscula and More translated.

The residue is now **one class, and it isn't a prompt bug**: the *translator of this edition* read as the author — Yonge for Bohn's Philo, Musschenbroek for the *Tentamina*. That's the original-vs-edition distinction this corpus already struggles with elsewhere. It needs the work graph, not a better sentence in the prompt.

### Stopping the tuning here

Precision is measured against a 100-book control. Another prompt iteration judged on that same control is overfitting, not improvement. Any further gain should be measured on a fresh sample.

**Cost:** ~1,480 in / ~100 out per book → ~7.2M in / ~480K out for the ~4,846-book target population, yielding ~1,300 candidates at ~90% precision, each with a page number, a page type, and a quoted line.

PR #3985, opened 2026-08-13
BPH catalogue: manuscripts, the records that had no editable address, and search export

JDerekLomas:
Four things José Bouman (BPH) reported — two through the footer feedback widget, two by email — and one root cause under three of them.

## What was reported

| when | where | what |
|---|---|---|
| 2026-07-31 | feedback, `/catalog/11312/edit` | "not possible to click on titles with a shelf mark M (+number), nor on those with shelfmark Fot (+ number) **to edit them** or see more information" |
| 2026-08-12 | feedback, `/catalog/31084/edit` | "Is it possible to export a search selection? […] This is an important feature, that we often! use!" |
| 2026-08-12 | email, "Needs your attention" | answers to all four worklist questions, incl. "change it to 'yes'" and "I cannot access these by clicking on them" |
| 2026-08-13 | email, "Cataloguing manuscripts" | "the blank form should show other fields than the one for printed books. No UBN required." |

## The root cause

Memorix issues no UBN for manuscripts or photographs. **2,012 of 29,881 rows have `ubn IS NULL`** — 812 manuscripts, 959 photocopies — and that is correct data, not missing data.

#3654 taught `/catalog/{key}` to accept a uuid, which made those records **viewable**. Everything else still keyed on a non-null UBN:

- `bph_works_revisions.ubn` is `TEXT NOT NULL REFERENCES bph_works(ubn)`
- `applyWorkRevision()` fetched and wrote with `.eq('ubn', …)`
- the edit page fetched with `.eq('ubn', …)` and `notFound()`d otherwise

So all 2,012 still 404'd on `/catalog/{key}/edit`, and no manuscript could be created at all. That is the unfixed half of the July report, and the whole of the August 13 one.

Separately, the workspace worklist built its sample links from `bph_works.id` — **a different uuid from `bph_works.uuid`** — so every link in the three UBN-less buckets resolved to nothing. That is the August 12 "I cannot access these by clicking on them", reported a week after the browser's own links were fixed. Same symptom, second path.

Inventing a UBN was not an option: the BPH writes UBNs into the physical book by hand, so a synthetic one ends up in ink inside an object not supposed to carry one.

## What this does

**Manuscripts are a first-class record type.** `record_type` is now chosen on the form (printed / manuscript / photocopy) instead of left null, and it picks the field set. Measured across all 812 manuscripts, exactly ten fields are ever used — `shelf_mark` 100%, `full_title` 82%, `uniform_title` 70%, `author` 64%, `language` 48%, `object_size_cm` 47%, `remarks` 41%, `binding` 29%, `provenance` 22%, `bibliography` 15% — and `title`, `ubn`, `place`, `year`, `printer`, `publisher`, `keywords`, `number_of_copies` are **0.0%** populated, because those describe an edition off a press and a manuscript is a unique object. That is what "the same format as those already catalogued" means concretely.

**`full_title` becomes editable.** 663 manuscripts keep their title there and none use `title`, so a manuscript's title was uneditable anywhere in the UI.

**Search export.** New `/api/catalog/bph/export` plus an Export control on the results header. The filters moved into `src/lib/bph-catalog-filters.ts`, shared with the browsing API, so the two can never match different sets — an export that quietly differs from the screen it was launched from is worse than no export, because nothing reveals the drift.

**The worklist stops asking questions José answered.** "Records with no UBN" (2,012) was counting correct data as a defect; narrowed to records with no UBN *and* no shelf mark — 253 that genuinely cannot be found by any means. The 86 undescribed manuscripts became a known backlog ("Leave them as they are"). The "ja" bucket is now a re-import detector.

**"ja" → "yes", already applied** (31 rows, production). Per-row revisions carrying the real from-value plus a JSON backup, so each is individually reversible. The normaliser maps `ja`→`yes` at the **write** boundary too — without that half, a Memorix re-import would silently undo the sweep.

## Verification

- **Filter refactor is behaviour-preserving**: 23 queries run against production (old code) and localhost (new module), comparing total **and** ordered results — **23/23 identical**. Covers the diacritic lane (Bohme→Böhme, 741), bare-year (1545, 33), both UBN lanes (131→126, "BPH 131"→1), every sort, every digitised mode.
- **Export matches its search exactly**: `q=JRR` → 41 rows and 41 results (José's actual query); `shelf_mark=M` → 11,595 and 11,595, i.e. 12 pages, so the supabase-js 1,000-row cap is handled.
- **The 404 is fixed without any migration**: the old worklist link `000f89e8-…` now returns 200 "Mein Manuscript"; the canonical uuid works; a printed book by UBN is unaffected; a nonsense key still 404s.
- **Edit-page lookup**: old behaviour 404s the manuscript, new behaviour finds it, printed books unchanged. (Tested at query level — the auth gate precedes the lookup, so a signed-out HTTP probe cannot reach it.)
- 2,738 unit tests pass, including 16 new ones pinning the shared predicate. `tsc --noEmit` clean, eslint 0 errors.

## Migrations — NOT yet applied

Both are additive and the code **degrades gracefully without them**, so this is safe to merge and deploy first.

1. `scripts/migration/add-bph-uuid-keyed-revisions.sql` — lets a revision target a work by uuid. Required before a manuscript can be **created or edited**. Backfills 107 uuid-less rows (all printed, all with a UBN), adds a UNIQUE on `bph_works.uuid` (verified 0 duplicates, 0 rows with neither key), makes `revisions.ubn` nullable with a CHECK that one key is always present.
2. `scripts/migration/add-bph-worklist-rpc.sql` — re-run for the new bucket and for `uuid` in the sample links.

Until (1) runs, `applyWorkRevision` drops `work_uuid` and ubn-keyed edits work exactly as before; a uuid-keyed edit returns a message naming the migration rather than a confusing FK error.

## Out of scope

- **"please change catalog to catalogue"** — the visible labels already say "catalogue" everywhere; what José sees is the URL. The proxy already accepts `/catalogue/…`, so switching the canonical links is a safe follow-up, but it touches every internal href and belongs in its own PR.
- José's guess that the UBN-less printed records carry an `In: [author-title]` note **does not hold** — 0 of 241 do; 228 are empty stubs with no title, shelf mark, or UBN. They said they were guessing *because they could not click through to look*, which is the bug this fixes. For the 31 "ja" rows the note holds for 14 of 31. Worth telling them when we reply.
- A 32nd row reads `"ja, maar heeft geen eigen ICN nr"` — a real free-text note, correctly left alone by the sweep, but the exact-match bucket never surfaced it. One for José.

### DIFF
--- a/scripts/lib/bph-state-shelfmark.mjs
+++ b/scripts/lib/bph-state-shelfmark.mjs
@@ -11,11 +11,17 @@
 /** Values that mean "no", not a shelf mark. Compared case-insensitively. */
 const NOT_A_SHELFMARK = new Set(['neen', 'nee']);
 
+/** `ja` carries real information (the copy IS on loan from the state
+    collection) so it is translated, not dropped — José Bouman, 2026-08-12.
+    See the TS twin for the full note. */
+const TRANSLATE = new Map([['ja', 'yes']]);
+
 export function normalizeStateShelfMark(raw) {
   if (raw == null) return null;
   const kept = String(raw)
     .split('|')
     .map((part) => part.trim())
-    .filter((part) => part.length > 0 && !NOT_A_SHELFMARK.has(part.toLowerCase()));
+    .filter((part) => part.length > 0 && !NOT_A_SHELFMARK.has(part.toLowerCase()))
+    .map((part) => TRANSLATE.get(part.toLowerCase()) ?? part);
   return kept.length > 0 ? kept.join('|') : null;
 }
--- a/scripts/maintenance/translate-ja-state-shelfmark.mjs
+++ b/scripts/maintenance/translate-ja-state-shelfmark.mjs
@@ -1,0 +1,213 @@
+#!/usr/bin/env node
+/**
+ * Translate the "ja" answers in bph_works.state_shelf_mark to "yes".
+ *
+ * WHY
+ * ---
+ * Sibling of clear-neen-state-shelfmark.mjs, which emptied the Dutch "no"
+ * answers on 30 July and deliberately left "ja" ("yes") alone because deleting
+ * it would have discarded the only record that those copies ARE on loan from
+ * the state collection. That was left as a librarian's decision, and surfaced
+ * as a bucket in the catalogue worklist so it would actually get asked.
+ *
+ * It got asked, and answered — José Bouman (BPH), 2026-08-12:
+ *
+ *   "State Collection shelf mark still reads 'ja' — These items will also have
+ *    in the remark-field the note 'In: [+ author-title]'. […] This set of 31 is
+ *    owned by the State, therefore the 'Ja'. Better change it to 'yes'."
+ *
+ * So the value is kept and translated, not emptied. The write boundary is
+ * updated in the same change (src/lib/bph-state-shelfmark.ts and its scripts
+ * twin now map ja→yes), so a Memorix re-import cannot put the Dutch back —
+ * without that half, this sweep would silently undo itself on the next import.
+ *
+ * AUDIT TRAIL
+ * -----------
+ * bph_works is the BPH's catalogue of record: every mutation is traceable via
+ * bph_works_revisions (see src/lib/bph-catalog.ts). This writes a revision row
+ * per affected UBN — with the real `from` value, so each change is individually
+ * reversible — BEFORE touching the live rows, matching the ordering
+ * applyWorkRevision uses: if the run dies midway the failure direction is an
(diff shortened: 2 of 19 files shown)


JDerekLomas:
Closes the loop on these feedback rows (both marked read; they stay unresolved until the migrations run and José gets a reply):

Feedback-ID: 6a6ca91f848190b0609b4588
Feedback-ID: 6a7c2feddd6d74d71c5678af

Plus two emails from José: "BPH catalogue 'Needs your attention'" (2026-08-12) and "Cataloguing manuscripts" (2026-08-13, replied "On it!").

Issue #3969, opened 2026-08-13
A sweep must record a ROW, not a COLUMN — 477 fields on `books`, 140 of them dead


JDerekLomas:
## The pattern

Every sweep in this codebase has written a new **field** on `books` instead of a new **row** in a log collection keyed to the book. Nothing enforced otherwise, so it happened ~140 times.

Measured 2026-08-13 against production Atlas (`scripts/audit/field-sprawl.mjs`, added in this PR):

**`books`: 395–403 top-level fields (sampled), plus ~174 nested sub-keys — ~570 addressable fields.**

| populated on | fields |
|---|---|
| ≥99% of docs | **17** |
| 50–99% | 57 |
| 10–50% | 76 |
| 1–10% | 105 |
| **<1%** | **140** |

The comparison that shows this is accretion rather than necessary complexity: `books_warehouse`, the newest of the three book collections, carries **129 fields with 25 core**. `books` carries **~400 with 17 core**. `deleted_books` has drifted into its own 260-field shape with a *different* core set (12 fields).

## Why this is a correctness problem, not a tidiness problem

**A query against a field that exists but is 2%-populated returns a confident, well-formed, wrong answer.** It does not error and does not return null — it returns a small clean number that reads like a finding.

Confirmed instances:

- **`tenantId` (5,614) vs `tenant_id` (20,010) vs `tenant` (148)** — three spellings of the field behind a security boundary. See `.claude/docs/invariants/tenant-lockdown.md`.
- **`hidden_reason` (23,483) vs `hide_reason` (500)** — a takedown sweep reading the wrong one touches 2% of the corpus. We already have a scar here (#3099).
- **`pages_count` (104,690) vs `pageCount` (13,324) vs `page_count` (4) vs `page_count_source` (28,189)** — four spellings of page count, one of them camelCase.
- **`archive_reason` (865) vs `archived_reason` (12)**; **`deleted_reason` (15) vs `deletion_reason` (15)** — the second pair is a perfect duplicate on the same 15 books, so no fill-rate heuristic can pick a winner.
- **`author_id` (50,406) vs `author_entity_id` (32,019)** — two author-identity fields, neither complete.
- **`year` (50,000)** alongside `year_source`, `year_earliest`, `year_latest`, `original_work_year`, `year_estimated`, `year_reasoning`, `year_provenance`.
- **`commons_full_url` (24,607) vs `commons_url` (23,544)**.

### The instance with proof

[...]

PR #3999, opened 2026-08-15
refactor(import): build book/page docs via the whitelisted constructors (#3969)

JDerekLomas:
## Why

PR #3984 added `makeBookDoc` / `makePageDoc` in `scripts/lib/book-docs.mjs` — whitelisted
constructors that throw on unknown or retired keys — and then nothing used them. An
affordance with no adoption prevents nothing. This makes them load-bearing: every
direct-to-Mongo import script now builds its `books`/`pages` insert docs through them, so
field #478 cannot ship silently (issue #3969).

## Scripts converted (50)

- `scripts/import/*.mjs` (36): al-badri-direct, batch-import-istc-bsb, bncf-aldine-direct,
  bsb-from-file, ccag-vii-pdf-direct, early-america-direct, fetch-wikisource-javanese,
  founders-collected-direct, founding-chase, founding-docs-direct, founding-enrich-direct,
  founding-influences2-direct, founding-tail-direct, fragmenta-direct,
  harvard-iiif-direct-batch, harvard-philosophy-direct, harvard-wuzhen-direct,
  ia-bundle-import, iiif-direct-import, import-artwork, import-hashika-e-posters,
  import-idp-batch, import-kloss-collection, import-met-egyptian, import-oraec,
  import-pdf-books, import-rijksmuseum-artworks, import-sefaria, import-umn-health-texts,
  jefferson-canon-direct, marcianus-299-direct, met-shunga-albums-direct, ndl-daoist-import,
  oraec-paginate-translate, shwep-curator-pass-import, tartu-dspace-pdf-direct
- `scripts/import/import-thirukkural.ts` (1) — TS adopted cleanly; `scripts/` is excluded
  from `tsconfig.json`, and `tsx` resolves the `.mjs` import, so there is no interop cost.
- `scripts/import-*.mjs` (8): aic, cleveland, commons, getty, met, nga, rijks artworks +
  import-naj-sancai
- `scripts/iiif-discovery/import-*.mjs` (3): import-from-cache, import-leiden-artworks,
  import-leiden-books
- `scripts/migration/` (2): bph-s3-to-r2, bulk-import-to-r2

## Fields added to the whitelist

The 2026-08-13 scan behind the original lists read only **top-level literal keys in
`scripts/import/*.mjs`-shaped call sites**. Adoption surfaced the rest: keys reached through
a helper's `return {…}`, through a conditional spread, or written by a `.ts` importer or a
`scripts/`-root artwork harvester. Each of these was already being written to `books` /
`pages` in production; none is new behaviour, and nothing was removed.

**`BOOK_FIELDS` +40**

| Field(s) | Written by | Why it was missing |
|---|---|---|
| `curator_notes` | import-thirukkural.ts, direct-import-cosmogony.ts | `.ts` importers were outside the scan |
| `display_author` | import-naj-sancai | `scripts/`-root script, partial scan |
| `original_title`, `date_earliest`, `date_latest`, `translation_status`, `scholarly_notes` | ndl-daoist-import | catalog-coverage vocabulary; `scholarly_notes` has no reader in `src/` — Track B candidate |
| `provider`, `held_by` | import-umn-health-texts, import-hashika-e-posters | top-level duplicate of `image_source.provider` — Track B candidate |
| `current_location`, `dimensions_display`, `faceted_tags` | rijksmuseum, met-shunga, hashika artwork docs | read by `ArtworkInfo.tsx` / artwork search |
| `image_full`, `image_thumb`, `thumbnail_blob`, `archived_full_url`, `full_width`, `full_height` | artwork importers | artwork image family; `image_thumb`/`thumbnail_blob` existed on `PAGE_FIELDS` only, and artwork docs live in `books` |
| `attribution_note`, `enrichment`, `linked_art`, `wikidata_id` | commons, rijks, nga | reached only via conditional spreads, so the literal scan never saw them |
| `harvested_at`, `harvest_source`, `harvest_category` | all seven artwork harvesters | harvester provenance |
| `commons_title`, `commons_url`, `commons_license`, `commons_license_id`, `commons_license_url`, `commons_usage_terms`, `commons_copyrighted`, `commons_attribution_required`, `commons_restrictions`, `commons_credit`, `commons_assessment`, `commons_sha1`, `commons_mediatype`, `commons_upload_date`, `commons_uploader` | import-commons-artworks (+ the `commons_title`/`commons_url`/`commons_license` trio in all six museum harvesters) | the licence/attribution block the original scan missed entirely |

**`PAGE_FIELDS` +1**: `summary` (`migration/bph-s3-to-r2`, `migration/bulk-import-to-r2` write
it as a `null` placeholder; read by the book-index and chat routes).

**Retired fields:** none of the 50 scripts writes `tenant_id`, `tenantId`, or `hide_reason`.
Nothing had to be removed. (`scripts/import/etl-kloss-catalog-to-supabase.mjs` writes
`tenant_id`, but to a **Supabase** table where it is the correct key — that script is not
converted and is out of scope.)

## Verification

- `node --check` on all 49 modified `.mjs` files — exit 0, no diagnostics.
- **Behaviour-preservation proof**: an AST pass parses each converted file, splices every
  `makeBookDoc(...)` / `makePageDoc(...)` call back down to its object-literal argument,
  drops the added import line, and diffs the result against `HEAD`. **50/50 byte-identical.**
  The only two apparent mismatches are the arrow-implicit-return shape
  (`=> ({…})` → `=> makePageDoc({…})`), where unwrapping legitimately cannot restore the parens.
- **Runtime smoke**: the same pass extracts every static key set actually passed at each of
  the ~90 call sites and calls the real constructor with it. Zero throws after the whitelist
  additions — i.e. these scripts will construct, not explode, when next run.
- `npx vitest run tests/unit/book-docs.test.ts tests/unit/sweep-log.test.ts` — 17/17 pass.
- `npx tsc --noEmit` — exit 0.
- No script was executed and no database was contacted.

## Hunks worth a close look

1. **`scripts/iiif-discovery/import-from-cache.mjs:136`** — `if (doc.resource_type === undefined) delete doc.resource_type;` now runs on the constructor's shallow copy. Behaviour is identical (the key is whitelisted, passes through as `undefined`, then gets deleted), but the doc is mutated after construction.
2. **`scripts/import/import-idp-batch.mjs:280-281`** — `book.thumbnail` / `book.pages_count` are assigned after `makeBookDoc` returns and before the insert; both keys are whitelisted, but the constructor never sees those two writes.
3. **`scripts/import/import-sefaria.mjs:243-244`** — same shape: `doc.ocr` / `doc.translation` set after `makePageDoc`.
4. **`scripts/import/import-umn-health-texts.mjs:158`** — the only genuinely dynamic spread: `makePageDoc({ _id, id, book_id, ...p })`, where `p` comes from the IIIF canvas mapper at :108-113 (`page_number`, `photo`, `thumbnail`, `photo_original` — all whitelisted today, but a change to that mapper would reach the doc unseen). **This is also the one place with a real behaviour delta**: those page docs set no `created_at`/`updated_at`, so the constructor now stamps both. Every other converted literal sets them explicitly.
5. **Nested builders** — `import-cleveland-artworks.mjs` (`buildDoc()`'s `return`), `shwep-curator-pass-import.mjs` (`bookDoc:` inside the object `buildBookDoc()` returns), and both migration scripts (`makeBookRecord()` / `makePageRecord()` returns) are wrapped at the `return`, which is where the whole doc is built.
6. **Conditional spreads** are validated only on the runs where the condition holds — `import-artwork.mjs` has 14, `import-commons-artworks.mjs`/`import-nga-artworks.mjs`/`import-rijks-artworks.mjs` one apiece for `attribution_note`/`wikidata_id`/`enrichment`. All their keys are now whitelisted, so the intermittent-throw hazard is closed, but the constructor's guarantee here is per-run, not static.

## Out of scope

- `scripts/import/pdf-direct.ts` and `scripts/import/direct-import-cosmogony.ts` — the two
  remaining `.ts` importers that insert into `books`/`pages`. Only `import-thirukkural.ts`
  was in the brief; converting the other two is a small follow-up.
- `scripts/maintenance/bph-nas-ingest.mjs`, `scripts/workers/erara-import-queue.mjs`,
  `scripts/split-book.mjs`, `scripts/repair-foreign-folder-pair.mjs`,
  `scripts/tmp-import-berlin-papyri.mjs` and other non-import writers — same treatment is owed,
  separate PR.
- Duplicate-family consolidation (`pageCount`/`pages_count`, `place_published`/`place_of_publication`,
  `image_full`/`commons_full_url`, top-level `provider` vs `image_source.provider`) stays with
  #3969 Track B. This PR deliberately records the sprawl rather than fixing it.
- No import script's actual behaviour was changed, and no "bug I noticed" was fixed.

## Docs

`.claude/docs/import-workflow.md` gains a short **Building the insert docs** section: new
import scripts build docs through the constructors, because unknown keys throw and field
#478 therefore cannot ship silently. It points at `.claude/docs/invariants/field-sprawl.md`
(merged in #3998) for the rule itself and for where per-book *actions* go instead
(rows via `scripts/lib/sweep-log.mjs`), rather than restating either.

Rebased onto `f2817c12` — no conflict with #3998, which does not touch `book-docs.mjs`.

### DIFF
--- a/.claude/docs/import-workflow.md
+++ b/.claude/docs/import-workflow.md
@@ -68,8 +68,17 @@ Applies to all contributors — human and AI.
 - `scripts/import/al-badri-direct.mjs` — original direct-insert pattern (bundled IA item).
 - `scripts/import/daoist-alchemy-ia-batch{,-2}.mjs` — batch-via-API examples.
 - `src/lib/dedup.ts` — fingerprint + normalize + `checkDuplicate` + `scanForDuplicates`.
+- `scripts/lib/book-docs.mjs` — `makeBookDoc`/`makePageDoc`, the whitelisted doc constructors.
 - `.claude/docs/chinese-iiif-sources.md` — per-source IIIF manifest patterns + gotchas.
 - `.claude/docs/daoist-alchemy-acquisition-list.md` — a worked want→dedupe→get example.
+
+## Building the insert docs
+Build every `books`/`pages` insert document with `makeBookDoc()` / `makePageDoc()` from
+`scripts/lib/book-docs.mjs` — never a bare object literal. Unknown keys throw, so a 478th
+field cannot ship silently the way the first 477 did; adding one is then a reviewed edit
+to the whitelist. All 50 direct-import scripts already do this — copy the nearest one.
+The rule, the incident behind it, and where per-book *actions* go instead (rows via
+`scripts/lib/sweep-log.mjs`) are in `.claude/docs/invariants/field-sprawl.md`.
 
 ## Invariants (don't violate)
 - Imports land HIDDEN; flip visible only after QA. Always set `visible`/`hidden` as a pair.
--- a/scripts/iiif-discovery/import-from-cache.mjs
+++ b/scripts/iiif-discovery/import-from-cache.mjs
@@ -39,6 +39,7 @@
 
 import { ObjectId } from 'mongodb';
 import { getScriptClient } from '../lib/mongo.mjs';
+import { makeBookDoc, makePageDoc } from '../lib/book-docs.mjs';
 
 const args = Object.fromEntries(
   process.argv.slice(2).filter(a => a.startsWith('--')).map(a => { const [k, v] = a.slice(2).split('='); return [k, v ?? true]; })
@@ -105,7 +106,7 @@ for (const c of cands) {
   const slug = await uniqueSlug(slugify(title));
   const bookId = new ObjectId();
   const facsimile = isArtwork || FACSIMILE;
-  const doc = {
+  const doc = makeBookDoc({
     _id: bookId, id: bookId.toHexString(), slug,
     title, display_title: title, author,
     language: c.language && c.language !== 'Unknown' ? c.language : 'Unknown',
@@ -131,7 +132,7 @@ for (const c of cands) {
     ...(c.metadata?.subject_geographic ? { subject_geographic: c.metadata.subject_geographic } : {}),
     source_fingerprint: fp, normalized_title: normTitle(title), normalized_author: normAuthor(author),
     created_at: now, updated_at: now,
-  };
+  });
   if (doc.resource_type === undefined) delete doc.resource_type;
 
   try {
@@ -139,7 +140,7 @@ for (const c of cands) {
     if (!isArtwork) {
       const CHUNK = 500;
       for (let s = 0; s < pages.length; s += CHUNK) {
-        const pdocs = pages.slice(s, s + CHUNK).map((p, k) => { const pid = new ObjectId(); return { _id: pid, id: pid.toHexString(), book_id: doc.id, page_number: s + k + 1, photo: p.photo, photo_original: p.photo, thumbnail: p.thumbnail, created_at: now, updated_at: now }; });
+        const pdocs = pages.slice(s, s + CHUNK).map((p, k) => { const pid = new ObjectId(); return makePageDoc({ _id: pid, id: pid.toHexString(), book_id: doc.id, page_number: s + k + 1, photo: p.photo, photo_original: p.photo, thumbnail: p.thumbnail, created_at: now, updated_at: now }); });
         await db.collection('pages').insertMany(pdocs, { ordered: false });
       }
     }
--- a/scripts/iiif-discovery/import-leiden-artworks.mjs
+++ b/scripts/iiif-discovery/import-leiden-artworks.mjs
@@ -36,6 +36,7 @@
 
 import { ObjectId } from 'mongodb';
 import { getScriptClient } from '../lib/mongo.mjs';
+import { makeBookDoc } from '../lib/book-docs.mjs';
 
 const args = Object.fromEntries(
   process.argv.slice(2).filter(a => a.startsWith('--')).map(a => { const [k, v] = a.slice(2).split('='); return [k, v ?? true]; })
@@ -111,7 +112,7 @@ for (const c of cands) {
   const bookId = new ObjectId();
   const now = new Date();
   const normTitle = normalizeTitle(title), normAuthor = normalizeAuthor(author);
-  const doc = {
+  const doc = makeBookDoc({
     _id: bookId, id: bookId.toHexString(), slug,
     title, display_title: title, author,
     language: c.language && c.language !== 'Unknown' ? c.language : 'Unknown',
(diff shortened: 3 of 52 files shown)


JDerekLomas:
## Post-merge verification (independent of the PR's own checks)

Spot-checked because the PR's verification was static (AST byte-identity) plus a synthetic smoke test — **no import script was ever executed**, and the failure mode inverted: these scripts used to silently write a stray field, and now they *throw and kill the import*. The PR flagged the gap itself: conditional spreads 'validate only on runs where the condition holds — per-run, not static.'

**1. Conditional spreads — the untested path.** Wrote an independent checker over all 50 files. It must count only keys landing at the constructor's TOP level, since `makeBookDoc({ dublin_core: { ...(x ? {dc_publisher: v} : {}) } })` contributes `dublin_core`, not `dc_publisher` — and it has to skip template literals or the brace-depth count is garbage (these files are full of `${…}`). Two earlier, sloppier versions of my own checker reported **39** and then **27** false alarms before the depth tracking was right; every one was a `dc_*`/`iconclass`/`archive_metadata` key nested inside an already-whitelisted parent.

```
constructor call sites           : 89
TOP-LEVEL conditional-spread keys: 56
NOT WHITELISTED (would throw)    : 0
RETIRED (would throw)            : 0
unresolvable dynamic spreads     : 1
```

**2. The one dynamic spread** (`import-umn-health-texts.mjs:158`, `...p`) is provably safe by reading: `p` comes from the mapper ten lines above with a fixed shape — `page_number`, `photo`, `thumbnail`, `photo_original` — all four whitelisted, and the `.filter()` adds no keys.

**3. Real-data cross-check.** Source analysis can only prove things about source, so: pulled the 60 most recent `status:'draft', hidden:true` books from production (fresh imports) and diffed their actual top-level keys against the whitelist. **32 distinct non-whitelisted keys** — then attributed each to a writer. **None is written by any converted script**; they all arrive post-import from the pipeline (`translation_pct`, `pages_blank`), the edition layer (`edition_key`, `work_slug`), work identity (`work_id_source`), and enrichment (`author_id`, `scan_quality`). So the constructors accept everything these importers actually produce.

**4.** `vitest` 17/17 green on current main.

Conclusion: no latent mid-import throw found on any statically reachable path, and the one dynamic path is verified by hand. Still genuinely untested: an end-to-end import run against a live upstream — the residual risk is an upstream source starting to emit a field shape none of these paths cover, which is exactly the case the constructor is supposed to catch loudly.

PR #4008, opened 2026-08-17
audit: DB-side entities-sweep interlock for when ssh cannot answer

JDerekLomas:
## The gap

The `entities` interlock in `CLAUDE.md` is correct as written — the `[r]` bracket added on 2026-08-13 stops `pgrep` matching its own command line. But **both halves need the box to answer**, and on 2026-08-17 ssh to Hetzner connected and was then reset on every attempt. The interlock could not be evaluated at all, right before a merge that triggers a production build.

**'I could not check' is not 'clear.'** That is the whole failure mode: an unanswerable check quietly becomes an all-clear because nothing forces you to notice.

## The addition

`scripts/audit/entities-sweep-active.mjs` asks Atlas instead of the box: has anything bulk-written `entities` in the last N minutes? Same condition the build actually cares about (`/explore` prerenders its ~1M-doc counts at `maxTimeMS: 25000`), but it works when the box is unreachable, catches a sweep launched from *any* machine, cannot self-match, and **exits 2 rather than 0 when it cannot reach the DB.**

This does **not** replace `pgrep` — that one names the process, which is more useful when it works. This is for when ssh cannot answer, or as a second opinion before a risky merge. The doc entry says exactly that.

## Verified in both directions before trusting it

- **The field exists.** `entities.updated_at` is on 100% of a 5,000-doc probe (1,021,482 docs total). Checked first, because a filter on an absent field matches nothing and reports a permanent false all-clear — the corollary in `tenant-lockdown.md` that already cost us an EPUB index.
- **Negative:** 10-minute window → `0`, exit 0, 'quiet'. (Genuinely true right now: the newest `entities` write is 2026-08-13.)
- **Positive control:** a window wide enough to cover that 2026-08-13 sweep → `11+`, exit 1, ACTIVE. A probe that has only ever returned 'nothing' is worthless.

## One diagnostic worth recording

I first read the box's 906ms ping and ssh resets as a loaded host — plausible, and it would have meant a sweep was likely running. Then ping to **8.8.8.8 came back at 826ms against a normal 28ms**: the local link was degraded, not the server. The doc note now says to judge the link before blaming the host, since the wrong reading here argues for exactly the wrong action.

CLAUDE.md goes 227 → 228 lines, inside budget.


JDerekLomas:
## Found a real weakness on this check's first live use — do not merge as-is

Ran it before merging (dogfooding) and it **exited 2 / UNKNOWN**:

```
Check could not run: Executor error ... bookstore.entities :: operation exceeded time limit
Treat this as UNKNOWN, not as safe.
```

The loud-failure semantics worked exactly as intended — it refused to report a false all-clear. But the underlying reason matters: **`entities` has no index on `updated_at`.**

```
_id_ · books.book_id_1 · name_1_type_1 · type_1 · book_count_-1
name_1_book_count_-1_total_mentions_-1 · entities_aliases_idx
wikidata_{birth_date,death_date,coordinates,id}_partial
```

So the query is a collection scan over 1,021,482 documents. Worse, it is asymmetric in the wrong direction: the ACTIVE case can short-circuit once it passes the threshold, but proving **zero** matches — the normal, quiet case — must scan everything. **The common path is the expensive one**, which is why it answered twice and then timed out under load.

Two other instruments ruled out while looking for something cheaper:
- **`currentOp`** would be ideal (instant, no scan, directly answers 'is something writing now') — but it returns **Unauthorized** for the application credential.
- **A sweep run-log** doesn't exist: `repair-entity-page-attribution.mjs` writes only `entities.bulkWrite`, nothing else.

### Options

1. **Index `entities.{updated_at: -1}`** — turns this into an instant, reliable primary check that catches a sweep from any machine. Cost: a background build on ~1M docs plus a small ongoing write overhead on a collection that gets bulk-written.
2. **Heartbeat from the sweep** — the writer registers itself in a tiny collection; the check becomes trivial. Cleanest design and it fits the ROW-not-COLUMN rule, but it only covers sweeps that opt in.
3. **Ship as best-effort** — honest but weak: a second opinion that degrades to UNKNOWN exactly when the cluster is busy, i.e. when a sweep is most likely running.

Holding this PR pending that decision rather than merging a check whose CLAUDE.md entry would promise more than it delivers.

Issue #4037, opened 2026-08-18
Artworks and the books they came from are not connected — 5 edges across 24,912 artworks


JDerekLomas:
## Why

Source Library holds two libraries that don't know about each other.

- **24,912 artworks** (15,842 visible) — plates, folios, emblems, engravings
- **21,474 readable visible books** — OCR'd, mostly translated, citable, DOI'd

They live in the same Mongo collection, distinguished only by `content_type`. There are **5 edges between them**: `books.source_book` is set on 5 of 24,912 artworks (0.02%).

The consequence is concrete. Two examples, both found by a reader asking a simple question:

**`/artwork/e-codices-zbz-ms-rh-0172-044-max-copy`** — an alchemist beside a glass vessel holding an ouroboros. It is folio 44 of Zürich, Zentralbibliothek, Ms. Rh. 172 — the *Aurora consurgens*. **We hold that entire manuscript**: https://sourcelibrary.org/book/zurich-zentralbibliothek-ms-rh-172-aquinas — 208 pages, 208 OCR'd, 204 translated, from e-codices, with a Zenodo DOI. The folio is our page 45, whose own OCR image description reads "a blue-scaled dragon or serpent curled in a circle, biting its own tail (an Ouroboros)". A reader looking at that image is one click from the Latin and its English translation, and gets no click.

**`/artwork/art-theosophie-und-alchemie-ouroboros-emblem-1598`** — the Commons wikitext states the provenance outright: woodcut, Basel 1598, printer Sebastian Henricpetri, *"aus: Reusner, Hieronymus — Pandora"*, illustration on printed page 257, Dresden SLUB Chem.969,misc.1. **We hold two copies of that exact 1598 Basel edition** (BSB), both fully OCR'd and translated, and the plate is already extracted in one of them at scan page 277.

Nobody else has this pairing — the image *and* the translated text of the book it came out of. It is arguably the most distinctive thing the library can do, and it is currently unbuilt.

## What already exists (and is being wasted)

**19,530 artworks already carry `enrichment.cross_references`** — 23,363 assertions across 4,780 distinct targets, each with a `text_or_author`, a `relationship`, and a `confidence`. They are already rendered, as the "Connected Texts" panel in `ArtworkInfo`. **They are plain text. Nothing is a link.**

They are not noise. Sampled against the Ficino cluster (2,047 refs, the largest):

[...]

PR #4054, opened 2026-08-18
docs(claude): a summary number is not the records it summarises

JDerekLomas:
The `gnite` ratchet, upward. One session (2026-08-18) produced **three retractions of one shape**, and none of the existing rules covered it.

The existing "absence of a marker is not the absence of the mechanism" bullet is this rule's sibling — that one is about *inferring from something missing*, this one is about *a present, correct-looking aggregate read as evidence about the underlying records*. It lands immediately adjacent so the two can't drift into separate aphorisms for the same family of mistake.

## The three

**`books.pages_ocr: 0` reported as "this book has no OCR".** It counts reader-*visible* pages. Two of the books named that way held **257** and **195** OCR'd *and translated* pages on soft-hidden pre-split spreads. Derek pushed back with "we definitely have OCR for some of those" and was right. The remedy inverted completely: not "commission a re-OCR" but "remap text already paid for."

**A day of bot totals said traffic was *below* median** (72K vs 141K the day before) while a self-amplifying error loop was making **389,193 requests in 35 minutes**. A daily aggregate cannot see a 30-minute burst.

**"~54 hand-rolled slug call sites"** was a `grep | wc -l` that counted normalisation keys as slug builders. The real figure was **33** — and the useful fact was not the count but *where they live* (`scripts/import/`, where a bad slug becomes a permanent public URL).

Plus the corroborating evidence from the same session: all four defects in the #4037 artwork-link resolver were found by **reading rows**. Each was a plausible-looking wrong link. The aggregate counts looked healthy the entire time and never flagged one.

## Tier and budget

`CLAUDE.md` goes **230 → 231 lines**, inside the ~250 budget, so nothing is demoted.

It belongs in `CLAUDE.md` rather than `invariants/` by the file's own test: the rule fired on a `books` field, on analytics aggregates, and on a shell pipeline. No subsystem triggers it, so it can't be routed conditionally.

Docs only.


JDerekLomas:

JDerekLomas:
Closing as superseded (backlog sweep 2026-09-12). The lesson this one-line bullet carries — a summary counter is not the records it summarises, `pages_ocr` counts reader-visible pages — landed in `.claude/docs/invariants/paired-artifacts.md` (lines 45–59) and the counter/aggregate entries of `measurement-instruments.md`, and CLAUDE.md has since gone under a word budget (7efa0f4b) that demotes exactly this kind of bullet to the invariants tier. Branch also conflicts with the restructured CLAUDE.md.

Issue #4117, opened 2026-08-20
Detector: derive books.languages[] from the per-page OCR <language> tag (the evidence lane #4089 needs)


JDerekLomas:
Split out of #4089, which asked for a `languages[]` field and found one already there —
degenerate (245 of 17,857 live arrays hold more than one language), unread by any code in
`src/`, and populated by a sweep that can only *parse* the existing string, never *detect*
anything. This issue is the missing half: the detector.

## The evidence source

The OCR model writes a `<language>` tag into `pages.ocr.data` on every page. It is real
per-page detection and it is already paid for. **The `ocr.language` field is not** — it
holds the request parameter (`null`, `"auto-detect"`) and must not be used for this.

On the Lascaris pair from #4089, the tag reconstructs the facing-page structure exactly,
and does so almost identically on two independently OCR'd copies of the same edition:

| | copy `69b220cc…` (340pp) | copy `6a085748…` (336pp) |
|---|---|---|
| Latin | 178 | 175 |
| Greek (incl. "Ancient Greek") | 153 | 151 |
| other / none | 9 | 10 |
| catalogued as | **Greek** | **Latin** |

Two copies, two catalogue answers, one measured reality. That is the whole argument for
the detector.

## What it should do

For every book with OCR'd pages: aggregate the `<language>` tag, normalize it, compute
each language's page share, and write `languages[]` ordered by share with `languages[0]`
equal to `language`, plus `language_multi`. Dry-run first, bucketed by confidence; write
only the high-confidence tier; ambiguous rows go to a review file.

Threshold to start from: a second language on **≥10% of OCR'd pages** makes it a listed
language. That number is a starting point, not a finding — it should be tuned against a
hand-labelled sample before any bulk write, because it is exactly the knob that separates
"bilingual edition" from "Greek quotations in a Latin book."

## It pays for itself three times over

On a 40-book sample drawn only to evaluate `ai_metadata.secondary_languages`, comparing
the catalogue against the page tags surfaced, incidentally:

- **4 outright language mislabels** (#2184's class, ~10% of the sample): a book catalogued
  French whose pages are 99% Arabic; one catalogued Korean whose pages are 99% Chinese;
  one catalogued English whose pages are 92% German; one catalogued Latin whose pages are
  100% English.
- **4 of 33 books with a real, unrecorded second language** at ≥10% (*Histoire des
[...]

Issue #4089, opened 2026-08-20
Bilingual editions: languages[] already exists and is degenerate — the gap is a detector and a read path, not a schema change


JDerekLomas:
## Correction (2026-08-21): the field this issue asks for already exists

The original report below is right about the defect and wrong about the fix. Before
anyone builds `languages?: string[]`, three measurements against production
(`bookstore`, 2026-08-21, live = `visible: true && pages_count > 0` = **22,068**):

**1. `books.languages` already exists — on 45,675 books, 17,857 of them live.** So does
`language_multi` (boolean). Both are written by
`scripts/maintenance/normalize-language-tags.mjs` (#2258, first run 2026-05-31), whose
header describes exactly what this issue proposed: canonicalize the scalar, parse
compound strings into `languages[]`, set a multi flag. It is registered in
`scripts/lib/books-known-fields.json`.

**2. It is degenerate, unread, and structurally cannot answer this question.**

| | live books |
|---|---|
| `languages` present | 17,857 |
| …of which singleton arrays | 17,489 |
| …**more than one language** | **245** |
| …empty array | 123 |
| `languages[0] !== language` | 892 |
| missing `languages` entirely | 4,211 |

- **Nothing in `src/` reads it.** It is not in `src/lib/types/book.ts`. Search filters on
  the scalar (`src/app/api/search/route.ts:172`) — *including its `languages=` query
  parameter*, which is a param named `languages` that queries the field `language`.
  `?languages=Latin` looks like it hits the array and does not. That trap should be
  fixed or the param renamed regardless of what happens to the rest of this issue.
- **The sweep parses, it cannot detect.** It splits whatever string is already in
  `language`. It can turn `"Greek-Latin"` into `["Greek","Latin"]`; it can never learn
  that a book tagged `"Greek"` is also half Latin. On the Lascaris pair it ran and
  confirmed the wrong answer: `69b220ccf79d8af0eab7fd3a` → `languages: ["Greek"]`,
  `6a08574849638a50931c42e9` → `languages: ["Latin"]` (`language_raw: "lat"` — it
  arrived as a MARC code, cf. #3893), both `language_multi: false`.

**3. The workaround is already in the data.** 96 of the 229 distinct live `language`
values are list-shaped strings, on **262 live books** — `"Greek-Latin"`,
`"Hebrew and Aramaic"`, `"Sanskrit-English"`, `"Arabic, Ottoman Turkish, Persian, Latin,
[...]

Issue #3261, opened 2026-07-19
Period-translation re-tag: Renaissance Latin translations mislabeled as originals (19 vs 47,139)


JDerekLomas:
Workstream E of #3258; named as an open defect in `.claude/docs/work-identity-casebook.md`.

## Problem
`text_role:period-translation` is used 19× against 47,139 `original` — the great majority of Renaissance Latin translations (Ficino, Bruni, Argyropoulos, Trebatius, "…interprete X" editions) are stored as originals with `original_language:Latin`. This (a) inflates "we hold the original" counts, (b) corrupts the source-language guard in FT matching (a Latin→En translation cannot defeat a Greek→En first), (c) violates ad fontes — the real source of Ficino's De mysteriis is Greek.

## Plan (measurable, reversible)
1. Candidate query: books with a named pre-1800 translator signal — title patterns ("interprete", "translatione", "…latine reddit"), known period-translator list, or work_id whose work's language ≠ book language.
2. Dry-run report bucketed by confidence; backup; $set `text_role:period-translation` + correct `original_language` on HIGH only; ambiguous → review file.
3. Re-measure the 19/47,139 ratio + holdings-count deltas; update casebook.
4. Guard test: FT source-language rule behavior before/after on the De mysteriis cluster.

Flagship acceptance case: Ficino's Latin De mysteriis editions become period-translations with original_language Greek (casebook worked example).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JDerekLomas:
**Instance list, measured 2026-08-21** (#4117 detector, PR #4129). This issue has argued since July that Renaissance Latin translations are stored as originals; here are the books, from page evidence rather than title heuristics.

Of 1,014 live books whose per-page OCR `<language>` tags contradict their catalogue entry, **64 are catalogued Greek with pages 99–100% Latin** — exactly the shape this issue describes. Sample: *Heronis Alexandrini Spiritalium liber* (Commandino 1575), *Ocellus Lucanus De universi natura* (Vizzani 1646), *Astronomicon Libri VIII* (Firmicus/Ptolemy), *Hierocles on the Pythagorean Golden Verses*, *Politicorum Libri Octo*. A further 31 are catalogued Latin with predominantly Greek pages (the mirror case).

Corpus-wide, **989 of the 1,014 carry `text_role: original`** — the same signal #3957 reports from the classifier side.

[...]

PR #4145, opened 2026-08-21
blog: Two Copies, Two Languages

JDerekLomas:
A research note growing out of #4089 / #4117.

## The story

We hold the Lascaris *Erotemata* (Venice 1495, Aldus) twice — a Greek grammar with Crastonus' Latin translation on facing pages. One copy is catalogued Greek, the other Latin, and **both entries are correct**. Filter the library to Latin and you get one copy; filter to Greek and you get the other; no filter shows you the book.

## Why it's worth publishing

The methodological half is stronger than the result. Twice in one afternoon the headline finding turned out to be the measuring instrument:

1. Compound catalogue strings (`"Latin/English"`, 96 of 229 distinct live values) read as contradictions — Comenius's *Orbis Sensualium Pictus*, catalogued correctly, reported as a mislabel.
2. "Chinese + Classical Chinese" — one text tagged two ways by the OCR model — was **2,423 books, 38%** of an apparent 6,230-book bilingual finding.

Both were catchable by one cheap check: sort findings by frequency and look hard at the biggest group. An artifact is systematic so it clumps; real defects scatter. That generalises well beyond languages, which is what makes it a post rather than a changelog entry.

It also covers the *hanmun* case — 196 books catalogued Korean whose pages are Classical Chinese, which is a fact about how Joseon scholarly writing was done, not an error — and the 64 Renaissance Latin translations shelved under Greek that #3261 has wanted instances for since July.

## Checks

- Every number is from the merged detector's full run over 21,481 live books (2026-08-21), not estimated.
- Both book links use canonical slugs (the id form 301s).
- Header image verified 200.
- `npx tsc --noEmit` clean.
- Two claims were softened before commit: the opening no longer asserts a specific failed query I couldn't reproduce (it states the filter behaviour, which is exactly true), and a flourish about "different centuries of ownership" was cut as unsupportable.

Adds one entry to the `posts` array, which the RSS feed reads from the same source.

https://claude.ai/code/session_01WqVg3Ya2b5Q5WQgtVeEeBT

### DIFF
--- a/src/app/blog/page.tsx
+++ b/src/app/blog/page.tsx
@@ -26,6 +26,19 @@ export interface BlogPost {
 // Exported so the RSS feed (/api/feed/blog) reads the same list — single source of truth.
 export const posts: BlogPost[] = [
   {
+    slug: 'two-copies-two-languages',
+    title: 'Two Copies, Two Languages',
+    subtitle:
+      'We hold the same 1495 Aldine grammar twice. One copy is catalogued Greek, the other Latin, and both are right. Measuring 21,481 books to settle it produced two headline findings that turned out to be the measuring instrument — and one cheap check that caught them both.',
+    date: '21 August 2026',
+    readTime: '8 min read',
+    tag: 'Research',
+    tagColor: 'bg-blue-50 text-blue-700',
+    image: 'https://images.sourcelibrary.org/pages/69b220ccf79d8af0eab7fd3a/0043.jpg',
+    imageAlt:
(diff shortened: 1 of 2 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek** — a new public blog post. Its engineering finding (96 of 229 distinct `books.language` values are compound) already landed in `language-fields.md`; the post itself is unpublished and is your call. Branch conflicts on the blog index; rebase when you decide.

Issue #4125, opened 2026-08-20
BNCF Aldine gap: the '104 missing' counted copies, not editions — and most of the real gap is not fetchable from IA


JDerekLomas:
## Summary

Two findings from auditing the BNCF Aldine gap (PR #4123). The first corrects a number we have been carrying; the second says the remaining gap mostly cannot be closed from Internet Archive.

## 1. The "104 missing Aldines" figure counted copies, not editions

BNCF's Aldine collection on IA (`ita-bnc-ald`) has **one item per physical copy**. The 1501 Martial appears three times — shelfmarks `Ald.1.1.1`, `Ald.3.2.19`, `Ald.3.2.20` — sharing EDIT16 CNC 36108 and an identical ISBD fingerprint.

Our dedup is manifestation-level (`ia_identifier` / `source_fingerprint`), and structurally cannot catch this: a second copy is a different scan of a different physical object. This is #2318 (work-level dedup not assigned at import) biting in a corpus where duplicate copies are common.

Measured at edition level (`scripts/audit/bncf-aldine-edition-gap.mjs`, added in #4123):

| | |
|---|---|
| IA items | 739 |
| Distinct editions | 606 (+65 unclassifiable) |
| Editions we hold | 534 |
| **Edition-level gap** | **~72** |

Of the 94 items we still do not hold, **15 are second copies of editions already in the library** and should never be imported.

Ten items were imported before this was checked; six were confirmed same-edition duplicates and marked `same_edition_duplicate` in #4123 (nothing deleted).

## 2. Most of the remaining gap is not fetchable from IA

The residue appears to be missing precisely *because* IA cannot serve it:

- **93 of the 104** have no JP2 derivatives, no `imagecount`, and no loose JPEGs — only a PDF.
- Their **IIIF manifests return HTTP 500** (sampled; the same probe returns 350 canvases for a book we already hold, so the probe is sound).
- The **fallback PDF is truncated at the source**: a 37 MB BNCF PDF contains zero `%%EOF` and zero `startxref` — the cross-reference table was never written, and poppler cannot open it. Content-length matched exactly, so this is not a transfer artifact.
- **Positive control:** of the 638 books already held, all 635 with a known source came in via `page_count_source: iiif_v3`. **None** has ever come in through the importer's PDF-rasterization branch. That branch has never successfully imported a BNCF book.

## Recommendation

[...]

PR #4153, opened 2026-08-21
docs(invariants): a gap computed at the COPY layer is not a gap; a search endpoint can ignore your query

JDerekLomas:
Two lessons from the BNCF Aldine work (#4123, #4125), filed into invariant docs that CLAUDE.md already routes to — so CLAUDE.md stays at 233 lines, under its ~250 budget.

### `edition-identity.md` — the copy layer, applied to a *source* catalogue

This file already models Person / Work / Edition / **Copy**. The new section applies that table to an acquisition source rather than to our own shelves.

A digitised-library collection has one item per **physical copy**. BNCF's 739 IA Aldines are **606 editions** — the 1501 Martial appears three times (`Ald.1.1.1`, `Ald.3.2.19`, `Ald.3.2.20`) under one EDIT16 number and one fingerprint. Manifestation dedup (`ia_identifier` / `source_fingerprint`) *structurally* cannot see this, because a second copy really is a different scan of a different object — so every duplicate copy classifies as NEW. That is how "104 missing" produced 6 duplicate imports.

Records: the bibliographic test (EDIT16 CNC, else ISBD fingerprint); match survivors against the **whole catalogue**, not the source's own items (~50 of 75 "unheld" editions were already here from BSB Munich, Ghent, Zürich, Johns Hopkins, Tufts, Google Books); quote coverage as **editions**, never items; attribution evidence lives in the *source item's* imprint, not our mostly-empty fields; and the short-title false negative that makes any "no match" bucket an upper bound.

### `measurement-instruments.md` — a 200 that isn't a search

IA's `services/search/v1/scrape` silently **ignored `q`**, returning an identical unfiltered page for `mediatype:texts`, `petrarca` and a collection query alike — HTTP 200, well-formed JSON, every time. A hunt across 17 editions reported 0/17 found and was pure artifact.

The failure points the dangerous way: a confident negative ("not held anywhere") produced by an instrument that was not searching. The tell was a magnitude floor, not an error — "petrarca returns 1 item on all of Internet Archive" is impossible.

Prescribes: a positive control every run; a **per-item control** (a hunt for copies of X must return X itself — this is the only reason the corrected result is trustworthy); `advancedsearch.php` over `scrape` for searching; and logging the endpoint's own total beside your filtered count so "returned nothing" is distinguishable from "did not run".

https://claude.ai/code/session_017PZMThrz2Gt9LBKj2gejGC


JDerekLomas:

Issue #4146, opened 2026-08-21
Native Spanish books are invisible to Spanish search — 67 books on /es with zero page_texts rows


JDerekLomas:
## The gap

#4120 widened the /es READ rule: a book *written* in Spanish is a Spanish edition, not just one translated into it. 67 live books joined /es on that basis.

`page_texts` — the Spanish search store built by #4095 — still selects on the old rule. Measured:

| | |
|---|---|
| `page_texts` rows | 38,627, all `lang=es` |
| books represented | 105 |
| Kaua (pivot translation) | 39 rows ✅ |
| **Scherzer 1857 (native Spanish)** | **0 rows** ❌ |

So the 67 native Spanish books are **visible on /es and unfindable by Spanish search**. That is exactly the failure `es-translate-worker.mjs` documents in its own header: *"an unembedded book and a book with no match return the same empty list."*

## Why this is NOT a one-line filter change

`scripts/workers/embed-page-texts.mjs` selects `{ [COUNTER]: { $gt: 0 } }` where `COUNTER = pages_translated_es`. Widening that selector alone would enqueue 67 books and embed **nothing from them**, because the composer reads `pages.translations.es` — which is empty for a native edition and always will be.

For a book written in Spanish, the Spanish text is `pages.ocr.data`. The composer has to learn to source from the OCR when the book **is** the language. That is a design change in `page-embedding-text.mjs` / `embed-book-page-texts.mjs`, not a filter tweak.

Open questions for whoever picks this up:
- Does a native page get `lang: 'es'` rows sourced from OCR, or a separate provenance marker so the two origins stay distinguishable?
- Staleness: the pivot path keys on `translations.es.updated_at`; an OCR-sourced row needs a different staleness signal.
- The same question exists for every future locale, so the answer should be language-keyed rather than Spanish-specific.

## Read first

**#4142** ("store the whole page in page_texts, embed only …") is open and touches the same store. Read it before starting — the two changes will collide otherwise.

## Third instance of one drift

#4120 changed the definition of "Spanish edition" on the read surfaces and left three writers on the old rule:

1. `sync-es-collection.mjs` — worse than a miss: it **un-tagged** Spanish originals on every run (fixed, #4141)
2. `/es/collections/[id]/page.tsx` — filed native books under "these read in their original language and in English" (fixed, #4138)
3. **this one** — still open, because it needs a real fix rather than a filter

[...]

PR #4154, opened 2026-08-21
feat(librarian): browse_catalog — answer 'what do you have in Spanish?' from the shelf, not the top passages

JDerekLomas:
## The bug

A reader asked the Spanish Librarian *"can you find me all the book published in spanish?"* and got **5 books** — assembled from the 8 passages the hybrid search happened to rank highest, presented as the library's Spanish holdings.

Measured against production this morning (live = `visible: true && pages_count > 0`, 22,073 books):

| | |
|---|---|
| `language: "Spanish"` (printed in Spanish) | **68** |
| any Spanish shelf incl. Old Spanish / Spanish / Latin / Nahuatl-Spanish | 74 |
| readable in Spanish (native + `pages_translated_es > 0`) | **176** |
| what the Librarian answered | 5 |

This is structural, not a bad turn. The Librarian's only view of the catalogue was `search`, a passage ranker — so every *how many / what do you have in X / list them all* question was answered from a top-k list. "How many books before 1600?", "everything in the astrology collection", "how many first translations?" all fail the same way.

## The fix

A `browse_catalog` tool: filter the catalogue, return the **exact** total, a representative page of rows with links, and the browse URL that shows the rest.

Filters: `language` (as printed on the leaves), `readable_in` (native + translated — what "libros en español" actually means), `collection` (a hard filter here, unlike the lean in `search`), `author`, `year_from`/`year_to`, `first_translation`, plus `sort` and `limit`.

Care taken where this class of tool usually goes wrong:

- **The count and the link must agree.** Language resolution prefers an exact catalogued value because that is what `/languages/<slug>` counts — reporting 74 while linking a page showing 68 is the worse answer. Compound shelves come back as *named variants with their own counts* ("Related shelves NOT in this total: Old Spanish (4), Spanish / Latin (1)…") instead of being silently folded in.
- **`language` is the edition's language**, per `.claude/docs/invariants/language-fields.md`; `readable_in` is the separate question and reuses `localizedEditionFilter()`, the same predicate every `/es` surface gates on, so the total matches what a reader can actually open.
- **No browse URL that means something else.** One is emitted only when a single-axis filter has a page showing exactly that set; anything narrower gets `null`. `/languages/<slug>` has no `/es` twin (`LOCALIZED_PATTERNS`), so that link is always unprefixed — an `/es/languages/…` link would 404.
- **A year-bounded count says what it drops** (~13% of live books carry no parsed `year`).
- Rows carry the localized title and the percent readable in the conversation's language; Spanish natives read "original Spanish edition", not "0% translated".
- Main-site tenant filter applied, so partner-room books cannot leak into a count.

Prompt guidance added routing catalogue-shaped questions to the new tool, with the failure it came from named. UI step label added in both locales ("Counting the shelves" / "Contando los estantes").

## Out of scope

- No new browse *page*. `/es` still has no Spanish-language search surface (#4146) — this makes the Librarian able to answer the question, it does not build the shelf UI.
- No change to `search`, its ranking, or the collection lean.
- `readable_in` only accepts `es` because `es` is the only non-English edition we publish; it is driven by `PREFIXED_LOCALES`, so a new locale gets it for free.

## Verification

`npx tsc --noEmit` clean. Counts above measured directly against production `bookstore`. End-to-end turn checked on the preview deploy before merge.


JDerekLomas:
## Verified end-to-end on the preview

Both turns run against `sourcelibrary-v2-3f9zv29m0` (anonymous, `visibility: private`).

**The original question, in Spanish** — *"can you find me all the book published in spanish?"* on `/es/librarian`. The Librarian called `browse_catalog` twice and answered with **both** numbers and the distinction between them: 146 readable in Spanish (originals + our translated editions), 67 printed in Spanish, and a link to `/es/collections/en-espanol`. Every book link carried the `/es` prefix; every author link resolved. Before this PR the same question answered "5".

**A catalogue question in English** — *"how many books printed before 1600, and how many of those are first translations?"* → 6,162 and 1,326, from two `browse_catalog` calls, no invented browse link.

Two real defects surfaced from those live turns and are fixed in the branch — neither was visible from the code or from measuring the queries:

1. **Fabricated author links.** Handed only author *names*, the model wrote `https://sourcelibrary.org/es/author/alfonso-x-el-sabio` — 404 twice over (invented slug, and `/author` has no `/es` twin). Only `/book` links are verified after the fact by `citation-fixes.ts`, so an author link has to be correct when the model *receives* it. Rows now carry the same resolvable link `search` supplies (`author_id`, else the slug of the stored name), always unprefixed. All six author links in the re-run return 200.
2. **A fabricated browse URL.** With no browse line in the result (year filters have no page of their own), the model composed `sourcelibrary.org/books?year_to=1599` — also a 404. Absence read as "supply one yourself", so the tool now says *there is no such page* out loud, and the prompt repeats the rule. The re-run wrote no browse link.

Query timings against production, all inside the 20s caps: whole library 0.45s, Latin 0.47s, before-1600 0.57s, author regex 2.7s, readable-in-Spanish 3.0s.

**Known, not fixed here:** the model still sometimes relabels a linked author in prose (it wrote "Arnaldus de Villanova" over a link to `/author/king-khalid`, the name that book is actually catalogued under). That is the pre-existing prose-vs-catalogue naming gap (#3114), not something this tool introduces — the URL it emitted was the right one for the row.

Issue #1727, opened 2026-05-12
Refactor: centralize page-image URL resolution (6 duplicated functions, subtle precedence bugs)


JDerekLomas:
## Problem

Six different functions across the codebase compute "the right image URL for this page", each with a slightly different fallback chain. Bugs from precedence mismatches keep surfacing:

- Reader thumbnail grid showed duplicate spread images for split pages (fixed in #1713 saga by setting `thumbnail_blob = cropped_photo` per-page in the DB — but the underlying \`getThumbUrl\` in BookOverview is still missing the \`cropped_photo\` check).
- CoverImagePicker showed full spreads for split pages because its \`getPageImageUrl\` returned \`page.photo\` (= full uncropped spread URL on these books) instead of \`page.cropped_photo\`. Just patched in this session.

## Map of the six functions

| Location | Function | Precedence (top → bottom) |
|---|---|---|
| \`src/lib/utils.ts:156\` | \`getPageImageUrl\` | split_from_spread → cropped_photo → archived_photo → photo_original → photo |
| \`src/lib/utils.ts:180\` | \`getPageDisplayUrl\` | split → display_photo → cropped_photo → proxy |
| \`src/lib/utils.ts:209\` | \`getPageThumbUrl\` | split → thumbnail_blob → derived path → proxy |
| \`src/components/reader/BookOverview.tsx:28\` | \`getThumbUrl\` | thumbnail_blob → thumbnail → archived_photo → photo (**no cropped_photo**) |
| \`src/components/reader/BookOverview.tsx:36\` | \`getHiresUrl\` | split_from_spread → archived_photo → display_photo → cropped_photo → photo_original → photo |
| \`src/components/book/CoverImagePicker.tsx:48\` | \`getPageImageUrl\` | crop → cropped_photo → photo → thumbnail_blob → proxy |

The intent of each is "get the right URL for X usage" but each was written independently. Subtle differences cause split-page bugs.

## Image-URL fields on a page doc (reference)

| Field | What it is | Set by | Best for |
|---|---|---|---|
| \`photo\` | original upload URL | upload route | source of truth for non-split |
| \`photo_original\` | original (legacy s3) | early imports | fallback only |
| \`archived_photo\` | R2 archive copy of \`photo\` | archive cron | proxy base URL |
| \`cropped_photo\` | the cropped half-page (if split) | spread-splitter | **canonical for split pages** |
[...]

PR #4163, opened 2026-08-21
fix(images): the Florentine Codex was unreadable — our own CSP blocked Getty (2,506 pages)

JDerekLomas:
## What's wrong

Reported from `/es/book/florentine-codex-vol-1-books-1-5`: every page thumbnail is a broken-image icon.

It is not an `/es` bug and not a data bug. All 2,506 pages of the three Florentine Codex volumes store `image_thumb` on `media.getty.edu`, and that host was never added to `CSP_IMG_HOSTS`. Every one of those URLs returns a clean `200 image/jpeg` to curl and is refused by every browser, so the pages grid, the cover picker **and the reader itself** render nothing — on `/book/...` and `/es/book/...` alike. The three volumes are the only books in the corpus on that host (measured against `books` on 2026-08-21), so the blast radius is exactly those 2,506 pages — but it makes one of the library's most significant Mesoamerican sources unreadable.

```
$ curl -sI 'https://sourcelibrary.org/book/florentine-codex-vol-1-books-1-5/page/20' | grep img-src
img-src 'self' data: blob: https://images.sourcelibrary.org … https://images.sub.uni-goettingen.de
                                            # no media.getty.edu

rendered:  <img alt="Page 20" src="https://media.getty.edu/iiif/image/…/full/150,/0/default.jpg">
```

## What this changes

**1. Allowlist the host.** `media.getty.edu` goes into `CSP_IMG_HOSTS` (what the browser may load) and `ALLOWED_IMAGE_HOSTS` (what `/api/image` may fetch). Both were missing, so neither the direct URL nor the proxy fallback could work.

**2. Screen page-image candidates against the CSP list.** `page-image-url.ts` now runs every provider-host candidate through `isBrowserRenderableImageUrl` — the same list the CSP directive is built from — and falls through to an R2 variant or the same-origin `/api/image` proxy when a host is not allowlisted.

That second half is the point. `getBookThumbnailUrl` has screened *covers* this way since 2026-08-04 (~1,550 books were rendering blank cards); the *page* resolver never did. So the identical failure could reappear on a different surface — and did — while a perfectly good R2 thumbnail sat one field away in `thumbnail_blob` for all 2,506 pages. After this change an un-allowlisted host degrades to something that loads instead of to nothing.

## Tests

`tests/unit/page-image-url.test.ts` gains three:

- no `thumb`/`display`/`hires` URL may leave the resolver on a host the CSP blocks
- a blocked host falls through to the R2 variant, or to `/api/image` when there is none
- an **allowlisted** IIIF host still gets its free origin-side resize and is *not* pushed onto our own compute (the cost-tier invariant from #1727 must survive the new screen)

`npx tsc --noEmit` clean; `page-image-url` (22), `csp-image-hosts` + `image-proxy-hosts` (80) all pass.

## Out of scope

- Backfilling `pages.image_thumb` for the three volumes to point at the R2 `-thumb.jpg` variants that already exist. The resolver now prefers whatever renders, so this is a tidiness question, not a correctness one.
- A standing detector that sweeps stored image hosts corpus-wide against the CSP list. Worth doing — this is the second time the list has drifted behind the data — but it belongs in `scripts/audit/`, not here.

### DIFF
--- a/src/lib/csp-img-hosts.ts
+++ b/src/lib/csp-img-hosts.ts
@@ -57,6 +57,10 @@ export const CSP_IMG_HOSTS = [
   'https://iiif-images.library.upenn.edu',
   'https://content.staatsbibliothek-berlin.de',
   'https://images.sub.uni-goettingen.de',
+  // Getty (Florentine Codex vols 1-3, 2,506 pages). Missing here meant every
+  // page thumbnail AND the reader's own page image were CSP-blocked — the
+  // books were unreadable in a browser while every URL returned 200 to curl.
+  'https://media.getty.edu',
 ] as const;
 
 /** The full img-src directive value, consumed by next.config.ts. */
--- a/src/lib/image-proxy-hosts.ts
+++ b/src/lib/image-proxy-hosts.ts
@@ -45,6 +45,7 @@ export const ALLOWED_IMAGE_HOSTS = [
   'rijksmuseum.nl',                  // Rijksmuseum
   'dl.ndl.go.jp',                    // National Diet Library Japan
   'e-rara.ch',                       // Swiss libraries
+  'media.getty.edu',                 // Getty (Florentine Codex)
 ] as const;
 
 /** IPv4 literal, or null if the string is not one. */
--- a/src/lib/page-image-url.ts
+++ b/src/lib/page-image-url.ts
@@ -25,6 +25,7 @@
  * archive pipeline runs opj_decompress → JPEG on R2. 100% of stored URLs are .jpg.
  */
 import { isUsableImageUrl, isArchiveFailed } from '@/lib/utils';
+import { isBrowserRenderableImageUrl } from '@/lib/csp-img-hosts';
 
 export type ImageSize = 'thumb' | 'display' | 'original' | 'hires';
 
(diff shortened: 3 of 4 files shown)


JDerekLomas:
Corpus-wide measurement, run after opening this PR — every distinct host in `pages.image_thumb` across all visible books with pages, checked against `CSP_IMG_HOSTS`:

```
OK       images.sourcelibrary.org   pages=3,803,279   books=14,760
BLOCKED  media.getty.edu            pages=    2,506   books=      3
OK       dl.ndl.go.jp               pages=    2,386   books=     30
OK       gallica.bnf.fr             pages=      827   books=      6
OK       images.uba.uva.nl          pages=      140   books=      1
OK       digi.vatlib.it             pages=       77   books=      1
```

Getty is the only blocked host, and it is exactly the three Florentine Codex volumes — so the allowlist entry closes the whole present-tense gap, and the resolver screening is what keeps the next one from being invisible.

JDerekLomas:
**Correction to this PR's description — the severity was overstated.**

The body says the page grid, the cover picker "and the reader itself" rendered nothing. **The reader worked.**

These pages carry `display_photo` on R2 and only `image_thumb` on Getty, so the damage tracked the **size tier**, not the book:

| surface | tier | outcome |
|---|---|---|
| page grid (`getPageGridUrl`) | thumb | **broken** — no upgrade path |
| cover picker | thumb | **broken** |
| book cover at `thumb` size | thumb | placeholder instead of a cover |
| adjacent-page prefetch | thumb | silently failed (perf only) |
| **the reader's main image** | display → R2 | **worked** |

`ImageWithMagnifier` paints `thumbnail` first and background-decodes the full-size `src`, swapping when ready — so a reader on these books saw a broken placeholder on first paint and then the correct page once the R2 display image decoded. Degraded, not dead.

How the error was made, which is the part worth keeping: the server-rendered HTML carries only the **thumb** ``; the display image is swapped in by script. Curling the reader page and reading its single `` tag therefore reports the thumb and looks like the whole story.

It surfaced from a contradiction: an R2-coverage census said *zero* pages are held-on-R2 yet served by a provider, which appeared to contradict this PR. The census measured the **display** tier; the breakage was at **thumb**. Both were right.

**The fix itself is unaffected and remains correct** — the thumb tier was genuinely broken for 2,506 pages, both allowlists were genuinely missing Getty, and the resolver genuinely wasn't screening. Only the blast-radius sentence was wrong.

Corrected in `.claude/docs/invariants/image-host-allowlists.md` (#4183), which now also carries the general rule: ask which *tier* a surface uses before declaring it broken, and check whether anything upgrades away from it.

PR #4176, opened 2026-08-21
fix(es): the catalog mirror carries the Spanish page counter (#4166)

### LINKED ISSUE #4166: books_catalog has no pages_translated_es, so /es rails miss books translated into Spanish
Split out of #4164, which fixed the link-side of this and hit the wall on the data side.

## What happens

On `/es`, the "Traducidas recientemente" rail links a book to its **English** page even when the book has a Spanish edition — so a Spanish reader clicking a card lands on English and loses the `/es` prefix for the rest of the visit.

`CollectionBookCard` now asks the shared `hasLocalizedEdition()` helper whether the book exists in the surface language. That helper answers from either of two signals:

- `language` matches Spanish → a **native** Spanish edition (#4120), or
- `pages_translated_es > 0` → **translated** into Spanish.

The homepage rails are fed from Supabase `books_catalog` via `CatalogBook`, which carries `language` but **not** `pages_translated_es` — the column does not exist on the table at all:

```
err: column books_catalog.pages_translated_es does not exist
```

So natives resolve correctly and translated-into-Spanish books do not. The guard is right; the data behind it is missing. (`.claude/docs/invariants/` — this is the "a correct fallback with nothing behind it" shape again, #4159.)

## Measured

Of the 15 books in the rail as it stood on 2026-08-21, **2 are readable in Spanish and still link to English**:

| slug | `language` | `pages_translated_es` |
|---|---|---|
| `libro-de-chilam-balam-de-kaua-scribes` | Yucatec Maya | 39 |
| `the-books-of-chilan-balam-the-prophetic-and-historic-brinton` | English | 19 |

One (`las-historias-del-origen-de-los-indios-de-esta-provincia-de-scherzer`, `language: Spanish`) is now correctly localized by the native path. The other twelve are genuinely English-only and correctly stay English.

The rail rotates, so the count is a snapshot, not a fixed 2.

## Why it wasn't fixed in #4164

Adding `pages_translated_es` to `books_catalog` means a column, a sync-writer change, and a backfill. **The writer is the part that bites** — widening a read rule without moving the writer is exactly what #4120/#4141 cost us, where a sync writer went on un-tagging Spanish originals on every run. That deserves its own PR and its own verification, not a rider on a nav fix.

## Suggested shape

1. Add `pages_translated_es` (int, default 0) to `books_catalog`.
2. Update the Mongo→Supabase catalog sync to write it — **and confirm it does not zero the column for books it does not touch**.
3. Add it to `BOOK_SELECT` and to the `CatalogBook` interface in `src/lib/books-catalog.ts`.
4. Backfill from Mongo, then re-crawl `/es` and confirm the rail's Spanish-readable books carry the `/es` prefix.

Generalizes beyond this one rail: any catalog-fed surface rendering `CollectionBookCard` without an explicit `href` has the same blind spot.

JDerekLomas:
Closes #4166.

## The defect

`hasLocalizedEdition()` answers "does this book exist in Spanish?" from two signals — a native `language` match (#4120) or `pages_translated_es > 0`. Supabase `books_catalog` carried the first and **not** the second: the column did not exist on the table at all (`err: column books_catalog.pages_translated_es does not exist`).

So on `/es`, every catalog-fed `CollectionBookCard` rendered without an explicit `href` resolved Spanish *originals* correctly and sent books *translated into* Spanish to their **English** page — dropping the reader off the `/es` prefix for the rest of the visit. The guard was right; the data behind it was missing. Same shape as #4159.

## What changed

**Data.** `scripts/migration/add-books-catalog-pages-translated-es.mjs` adds `pages_translated_es` (`int NOT NULL DEFAULT 0`) plus a partial index, and backfills from Mongo in both directions — sets nonzero counters, resets rows Mongo no longer counts — so a re-run repairs drift rather than only growing. **Applied to production already:** 107 books, 39,643 Spanish pages. `DEFAULT 0` is the fail-toward direction: a missing counter must never mint an `/es` URL we cannot serve.

**Writers.** Both full-row writers (`scripts/workers/sync-books-catalog.mjs`, `scripts/migration/rebuild-books-catalog.mjs`) gain the field in the row builder **and** the Mongo projection. That pairing is the whole risk and the reason this wasn't a rider on #4164: the upsert is `onConflict: 'id'`, so an omitted column is left untouched — but a field in the builder and *not* in the projection reads `undefined` and writes `0` for **every** book, which looks like a clean sync while un-localizing the entire corpus. Exactly #4120/#4141.

`scripts/maintenance/sync-pages-translated-es.mjs` now bumps `updated_at` when it moves a counter. The catalog sync is incremental on `updated_at`; without the bump, a counter changed by that sweep would never reach the mirror and the card would keep linking to English (#3445).

**Read side.** `BOOK_SELECT`, the `CatalogBook` interface, `MiniBook`, and the `/api/books/browse` select list. The last one has no `/es` twin today so it is inert — it is in scope because it feeds `CollectionBookCard` with no `href`, and would fail silently the day an `/es` browse ships. That is the generalization the issue names.

## Verification

Read back from `books_catalog` after the backfill, pinned to the slugs the issue measured (the rail rotates, so a count would prove nothing):

| slug | `language` | `pages_translated_es` | |
|---|---|---|---|
| `libro-de-chilam-balam-de-kaua-scribes` | Yucatec Maya | **39** | now localizes |
| `the-books-of-chilan-balam-…-brinton` | English | **19** | now localizes |
| `las-historias-…-scherzer` | Spanish | 0 | positive control — still resolves via the native path |

`tests/unit/catalog-localized-counter.test.ts` pins the read half by value and the writer pairing by source (comments stripped, because both files explain the pairing in prose that would otherwise satisfy the match on its own). **Negative control run on all three assertions** — each goes red with its guarded token removed, per `tests-that-are-not-guards.md`. Full suite: 2,858 passing; `tsc --noEmit` clean.

## Out of scope

- **Rendering confirmation on a live `/es`.** The data and the helper are verified; the end-to-end crawl belongs to the preview build and is noted below rather than claimed here.
- **`RecentlyRead`** builds `MiniBook`s from the reading-history API, a different payload lane with its own select. Untouched.
- **Languages beyond `es`.** `TRANSLATED_COUNTER` is a one-entry map; a second language needs a second column and the same writer pair, which is a decision, not a rider.

https://claude.ai/code/session_01MojJdMhnuk3PkZfpTPyA7k

### DIFF
--- a/.claude/docs/i18n.md
+++ b/.claude/docs/i18n.md
@@ -231,11 +231,21 @@ it is enforced at the ROUTE, not at link sites:
   which a crawler will index. Measured: the page-level version returned 200
   with `NEXT_REDIRECT` in the body. Same reason `notFound()` lives up there.
 - `hasLocalizedEdition` returns **`boolean | null`**. `null` means the payload
-  could not answer — the counter is a Mongo field and the Supabase catalog
-  fast-path lacks it. Never read that as false: it would 307 a genuinely
-  Spanish book away. Where the counter WAS explicitly projected, resolve with
-  `?? false` — absence is then an answer. Both halves of this were gotten wrong
-  once each and caught by measurement, not by review.
+  could not answer — a hand-built card object, a narrowed API select, anything
+  that carries neither `language` nor the counter. Never read that as false: it
+  would 307 a genuinely Spanish book away. Where the counter WAS explicitly
+  projected, resolve with `?? false` — absence is then an answer. Both halves of
+  this were gotten wrong once each and caught by measurement, not by review.
+- **The store that feeds a card must carry BOTH signals, or the guard answers
+  with half the truth.** Supabase `books_catalog` carried `language` and not
+  `pages_translated_es` — the column did not exist — so `/es` rails resolved
+  Spanish *originals* and sent books *translated into* Spanish to their English
+  page (#4166). Nothing errored; the helper was correct and the data behind it
+  was missing (#4159's shape). When you add a localization signal, walk every
+  mirror, select list and card interface that feeds a `CollectionBookCard`, and
+  remember the **writer**: a mirror column needs the field in the row builder
+  *and* the Mongo projection, or the sync writes 0 for every book and
+  un-localizes the corpus while reporting success (#4120/#4141).
 
 What this rule buys: the `/es` index contains only pages that are actually in
 Spanish, and a Spanish reader who clicks a book we haven't translated lands on
--- a/scripts/maintenance/sync-pages-translated-es.mjs
+++ b/scripts/maintenance/sync-pages-translated-es.mjs
@@ -98,7 +98,11 @@ for (const b of existing) {
   changed++;
   console.log(`${dryRun ? '[dry] ' : ''}${b.id}  ${prev} → ${next} / ${b.pages_count ?? '?'}  ${(b.title || '').slice(0, 60)}`);
   if (dryRun) continue;
-  const r = await books.updateOne({ id: b.id }, { $set: { pages_translated_es: next } });
+  // `updated_at` is load-bearing, not cosmetic: the Supabase catalog sync is
+  // incremental on `updated_at`, and books_catalog now mirrors this counter
+  // (#4166). A counter changed here without the bump would never reach the
+  // mirror, so a /es card would keep linking to the English page (#3445).
+  const r = await books.updateOne({ id: b.id }, { $set: { pages_translated_es: next, updated_at: new Date() } });
   if (r.modifiedCount !== 1) console.warn(`  ! updateOne for ${b.id} modified ${r.modifiedCount}`);
   await recordSweepAction(db, {
     sweep: 'sync-pages-translated-es',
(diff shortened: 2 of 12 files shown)


JDerekLomas:
## Verified on the preview build, against production as the control

Preview `sourcelibrary-v2-jf6g8eo1b` (commit `89ee19a`, `● Ready`), both crawled with a browser UA:

| slug | prod `/es` (before) | preview `/es` (after) |
|---|---|---|
| `libro-de-chilam-balam-de-kaua-scribes` | `/book/…` | **`/es/book/…`** |
| `the-books-of-chilan-balam-…-brinton` | `/book/…` | **`/es/book/…`** |
| `las-historias-…-scherzer` (native, positive control) | `/es/book/…` | `/es/book/…` |
| `chilam-balam-de-kaua-maler-photographs-1887` (no Spanish pages) | `/book/…` | `/book/…` |

Pinned to slugs rather than a count, since the rail rotates. The last row is the negative control that fell out of the same crawl — a neighbouring Chilam Balam item with no Spanish pages, which correctly stays English, so the fix is not simply prefixing everything.

`test` green (2m46s), `field-write-lint` green, `tsc --noEmit` clean, 2,858 unit tests passing.

Production is unaffected until merge: the column and its backfill are already live, but nothing deployed selects it, and the sync's `onConflict: 'id'` upsert leaves a column absent from the row object untouched rather than zeroing it.

Issue #3826, opened 2026-08-09
Spend dial was blind and bypassed: $2.3K billed on relight day vs $15 budget (GCP anomaly 2026-08-08)


JDerekLomas:
## What happened

Google flagged a cost anomaly: **$2.3K billed on 2026-08-08** on GCP project `booksplit` (expected $0), almost all Gemini API — $1.97K of it flash **output** tokens. Aug 9 was on the same pace (~$900 computed by mid-day) until the line was paused at 13:38Z.

Three independent defects stacked:

### 1. The spend guard reads the store the logger doesn't write to
`scripts/lib/spend-guard.mjs` → `getTodaySpendUsd()` aggregates **Mongo** `gemini_usage`. But `supabase-usage-logger.mjs` writes to **Supabase** `gemini_usage`, falling back to Mongo only on error. On Aug 8 Mongo recorded **$9.00** while Supabase recorded **~$830 computed** (billed ~$2.3K — the known ~3x computed-vs-billed factor). The dial compared $9 < $15 and kept dispatching all day. This is the documented "two stores, mutually exclusive — sum them" trap (infrastructure-costs doc), embodied in the safety control itself.

**Fix:** `getTodaySpendUsd` must sum BOTH stores (Supabase primary + Mongo fallback), or the logger must dual-write a daily counter the guard can read cheaply.

### 2. translate-worker self-dispatch never consults the dial
`budgetAllowsDispatch()` gates only orchestrator Phases 2/4. `translate-worker.mjs` self-dispatches books directly (`Self-dispatched: <book> — N pages`, incl. "fresh books mid-run") with no guard call. Even a correct meter would not have stopped the biggest spender.

**Fix:** call `budgetAllowsDispatch` before each self-dispatch (and before mid-run top-ups).

### 3. Duplicate concurrent dispatch, ~4x
The translate log shows the same book self-dispatched 3–4 times within seconds with different job ids (e.g. "Krestos Gospels (Illuminated) — 394 pages" x4). Same-cycle duplication multiplies cost on the exact books that are already the most expensive.

**Fix:** dedupe by book id on dispatch (in-memory set + job-collection uniqueness).

## Amplifier: runaway output on the non-Latin cohort
[...]

PR #4189, opened 2026-08-21
fix(images): 5,490 provider-hosted page thumbs now served from our own R2 (APPLIED)

JDerekLomas:
**Dry-run only — this commits the sweep, it has not been applied.**

## The problem

`image_thumb` is the **first** candidate the thumb tier of `getPageImageUrl` considers, so a provider URL there beats an R2 variant sitting in `thumbnail_blob` — even for pages we archived long ago. That is exactly how 2,506 Florentine Codex pages rendered nothing in #4163: Getty was missing from `CSP_IMG_HOSTS`, every browser refused the URL, and the working R2 thumbnail one field away was never reached.

The Getty entry fixed the instance. This removes the class.

## Scope, measured across live books

**5,490 pages, 39 books — every one already has its R2 thumbnail.**

| host | pages | books |
|---|---:|---:|
| `media.getty.edu` | 2,506 | 3 |
| `dl.ndl.go.jp` | 1,940 | 28 |
| `gallica.bnf.fr` | 827 | 6 |
| `images.uba.uva.nl` | 140 | 1 |
| `digi.vatlib.it` | 77 | 1 |

Four further pages (one book) hold a malformed thumb URL with no R2 alternative, and are correctly left alone.

Nothing is fetched, stored or generated — this copies a verified value one field over. Afterwards, 5,490 pages stop depending on another institution's uptime, rate limits and URL scheme, and five hosts stop needing to be kept in step across **two** allowlists where a miss is silent and total.

## Safety, in the order it runs

1. Candidates require `thumbnail_blob` to be on R2 already — no key is derived or invented. Per #3362 (a page key must contain its own `book_id`), it **asserts** book scope on every value and refuses the whole run if any fails.
2. **HEAD-checks the R2 targets before writing** and refuses on any non-200 — repointing at a 404 would turn a blocked image into a missing one, which is worse. `--verify-all` checked all 5,490: **every one returned 200.**
3. Preserves the old value in `field_provenance.image_thumb.previous`, so it is reversible.
4. Records one `sweep_log` **row** per book, never a new column (field-sprawl rule).
5. No `--apply` flag passed → no writes.

The HEAD check is 16-way concurrent with a progress line every 250. Serially and silently it was indistinguishable from a hang for several minutes, which is how the first run got killed before printing a single line — worth fixing for whoever runs it next.

## Out of scope

The **24,488 pages with no R2 object at all** (concentrated in IIIF-sourced books). Those need the archiver, not a metadata fix, and the line is deliberately paused (#3826). Note this is the same set as the "served by provider at display tier" pages — the two gaps are one seen from two angles. Where we hold the bytes, the resolver already serves them from R2 with zero exceptions.

## Also found, and filed separately

`pages_archived` is unreliable in **both** directions and should not be used to answer "how much is archived". Over the 8 largest books with a counter gap it claims **8,262** pages missing when the real figure is **388** — one book reads `9 / 1625` while all 1,625 pages hold an R2 URL, and another claims 1,186 pages archived while having **zero page documents**. `archive-bulk.mjs` selects its work by this counter.

### DIFF
--- a/scripts/maintenance/repoint-provider-thumbs-to-r2.mjs
+++ b/scripts/maintenance/repoint-provider-thumbs-to-r2.mjs
@@ -1,0 +1,176 @@
+#!/usr/bin/env node
+/**
+ * Repoint `pages.image_thumb` from a provider's server to the R2 thumbnail we
+ * already hold.
+ *
+ * WHY (#4163)
+ * -----------
+ * `image_thumb` is the FIRST candidate the thumb tier of `getPageImageUrl`
+ * considers, so a provider URL there wins over an R2 variant sitting in
+ * `thumbnail_blob` — even though we archived the page and hold the bytes.
+ * That is precisely how 2,506 Florentine Codex pages rendered nothing: Getty
+ * was missing from `CSP_IMG_HOSTS`, every browser refused the URL, and the
+ * working R2 thumbnail one field away was never reached.
+ *
+ * Serving our own bytes removes the whole class:
+ *   - no dependency on a provider's uptime, rate limits or URL scheme
+ *   - no allowlist entry to maintain in TWO places (CSP + the image proxy),
+ *     where a miss is silent and total
+ *   - R2 egress is free; we are already paying to store these objects
+ *
+ * SAFETY
+ * ------
+ * - Only touches pages whose `thumbnail_blob` is already an R2 URL. Never
+ *   invents a key, never derives one — see the #3362 rule that a page key must
+ *   contain its own book_id; this script copies an existing verified value and
+ *   asserts book scope before writing.
+ * - HEAD-checks a sample of the R2 targets before any write, so we cannot
+ *   repoint at a 404. `--verify-all` checks every one.
+ * - Preserves the old value in `field_provenance.image_thumb.previous` so the
+ *   change is reversible.
+ * - Records one `sweep_log` ROW per book, not a new column (field-sprawl rule).
+ *
+ * Usage:
+ *   node --env-file=.env.production.local scripts/maintenance/repoint-provider-thumbs-to-r2.mjs
+ *   node --env-file=.env.production.local scripts/maintenance/repoint-provider-thumbs-to-r2.mjs --verify-all
+ *   node --env-file=.env.production.local scripts/maintenance/repoint-provider-thumbs-to-r2.mjs --apply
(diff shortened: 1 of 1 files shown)


JDerekLomas:
**Status change: this is now APPLIED, and the script was rewritten after the first run died mid-flight.** The description above still says "dry-run only" — read this instead.

## What happened

The version reviewed above issued **one** query: every page of every live book, a 22,073-element `$in` plus an `$expr` regex, over 18.9M documents. It ran ~10 minutes and died on `MongoNetworkError: read ECONNRESET` before printing a line.

**Nothing was written** — verified rather than assumed: 0 `sweep_log` rows, and probe pages on all three Florentine Codex volumes still read `media.getty.edu` with no provenance stamp. So the cost was time. But the design was one flaky connection away from a partial write with no record of where it stopped, which is the corpus-walk rule this repo already has: *a walk needs a checkpoint first.* The dry run passing hid it, because a dry run that dies costs nothing while a write that dies costs correctness.

## The rewrite

- **Phase A** discovers candidate *book* ids once and caches them. Drops the giant `$in` (intersecting with the live set in memory is free) and uses a plain `$not` instead of `$expr`.
- **Phase B** runs book by book — each query indexed on `book_id`, each book checkpointed after its write. An interruption costs one book; re-running resumes.
- `socketTimeoutMS: 0` for the legitimately-long discovery scan, `retryReads`/`retryWrites`, and a per-book skip that deliberately **does not** checkpoint when a HEAD check fails, so a suspect book gets looked at rather than silently passed over.

## Result — verified independently of the script's own log

**5,490 / 5,490 modified, 0 skipped**, host totals matching the dry run exactly (Getty 2,506 · NDL 1,940 · Gallica 827 · UvA 140 · Vatican 77).

- **39 `sweep_log` rows**, one per book.
- The three pages probed *before* the run now read `images.sourcelibrary.org/pages/<book>/NNNN-thumb.jpg`, each carrying its `previous` value — **reversible**.
- **4 provider thumbs remain**: exactly the malformed-URL pages with no R2 alternative, correctly untouched.
- Live page after revalidation: **33 rendered `` tags, 0 using Getty.**

The 108 `media.getty.edu` strings still in that page's HTML are the `photo` field in the serialised payload — the **source** URL, which is provenance and should keep pointing at Getty.

Issue #4190, opened 2026-08-21
books.pages_archived disagrees with the pages in both directions — and the archiver selects work by it


JDerekLomas:
`books.pages_archived` does not agree with what the pages actually hold, in **both** directions — and `archive-bulk.mjs` selects its work by it.

Found while answering "is everything archived and served from R2?". Two measurements of the same question disagreed by 4.7×, and the counter is the one that is wrong.

## The divergence

Page-level truth (does the page doc carry an `images.sourcelibrary.org` URL in `archived_photo` / `photo` / `display_photo`), across all live books:

```
pages on live books   5,539,263
holding an R2 object  5,514,775   99.56%
no R2 object             24,488    0.44%
```

Book-level counter (`pages_count - pages_archived`) over the same corpus: **115,316 pages missing across 899 books**. That is 4.7× the real figure.

Spot-checked against the pages themselves — the 8 live books with the largest counter gap:

| book | counter | reality |
|---|---|---|
| Ss. theologiae problemata | 2347 / 2368 | partial: 2347/2368 on R2 — **correct** |
| Cosmographey (`6a26bea5…`) | **9 / 1625** | **all 1,625 pages on R2** |
| Cosmographey (`6a26bd05…`) | **9 / 1558** | **all 1,558 pages on R2** |
| Hoffmann, Syntagma (`6a26b9d8…`) | **42 / 1390** | **all 1,390 on R2** |
| Bibliotheca Classica (`6a26b940…`) | **42 / 1368** | **all 1,368 on R2** |
| Elementa Matheseos II (`6a26ba50…`) | **20 / 1294** | **all 1,294 on R2** |
| Les espistres familieres (`69db957a…`) | 1186 / 1194 | **zero page documents exist** |
| Joh. Ravisii Textoris (`6a26c47b…`) | 0 / 1120 | partial: 753/1120 on R2 |

**Over those 8 books the counter claims 8,262 pages missing; the real number is 388.**

Two distinct defects:
- **Under-counting** — books fully on R2 reading `9 / 1625`. All are `image_source.provider: iiif`, `ia_identifier` absent, so an archiver that isn't `archive-bulk` wrote the pages without syncing the counter.
- **Over-counting** — `Les espistres familieres` claims 1,186 pages archived and `pages_count: 1194` while having **no page documents at all**. A visible book with no pages is its own problem.

## Why it matters

[...]

PR #4191, opened 2026-08-21
docs(ratchet): ask "could this be a CHECK?" before "which doc tier?"

JDerekLomas:
Yes — and the evidence for it is a single session's findings.

## The problem with the current reflection step

`/gnite` step 4 asks two questions, and **both are about prose**: "does a doc need something new?" and "is anything in the doc no longer earning its place?" That institutionalises the weakest layer. `knowledge-layer.md` already says docs are the weakest layer; the ratchet didn't act on its own conclusion.

Measured 2026-08-21, one session, four findings. **Three were classes where the doc already existed and the thing recurred anyway:**

| the doc said | what happened |
|---|---|
| `csp-img-hosts.ts`: *"one edit, both layers"* | the second image resolver still never screened — 2,506 pages browser-blocked for months (#4163) |
| #3293: *"validate a denormalised counter against the READ path"* | `pages_archived` drifted 4.7×, and the archiver selects work by it (#4190) |
| edition + language invariants | a Florentine art treatise still imported as an Aztec codex on a title match (#4172) |

The one doc that demonstrably *worked* that day is the instructive case, and it is **procedural, not descriptive**: `language-fields.md` says "hand-check the largest cluster before quoting a rate". Following it caught a 211-book hanmun artifact that would otherwise have shipped as a headline number.

## What changes

The reflection now asks **three** questions, guardrail first, naming the repo's own existing pattern: **21 of 214 unit tests already sweep a directory and assert a property** (`csp-image-hosts.test.ts` over `next.config.ts`, `locale-prefix-not-tenant.test.ts` over `src/app/`). That shape would have caught #4163 at commit time. Other forms named: a detector that files what it finds, a constructor that throws (`makeBookDoc()`), a script that refuses to run without its guard.

Applied to `/gnite`, `/lesson` (which runs the same loop mid-session), and the CLAUDE.md bullet that describes the ratchet.

## Two brakes, deliberately

This could easily become the tests-shaped version of the 827-line CLAUDE.md, so:

1. **Point at `invariants/tests-that-are-not-guards.md` and require the negative control** — delete the guarded line, watch it go red, restore it — before anything counts as a guard. That doc exists because this repo has already shipped source-shape tests that passed with the guarded code deleted.
2. **State when prose is right.** A guard is the wrong tool for lessons about *judgment*: "ask which size tier a surface uses before calling it broken" cannot be asserted, and it's the lesson that corrected a severity claim in this same session. The discriminator: **nameable file or symbol → check; human about to draw a conclusion → doc.**

## Budget

CLAUDE.md stays at **236** lines — the existing ratchet bullet was rewritten, not extended. No new routing entry, no new invariant file.


JDerekLomas:
A fourth case for the table, from a different session the same day — and it lands on the argument rather than beside it, because the recurrence happened **inside the instrument**.

| the doc said | what happened |
|---|---|
| #3293: *"validate a denormalised counter against the READ path"* | `page-texts-coverage.mjs` — the detector written to catch *"an unembedded book and a book with no matches return the same empty list"* — selected books by `pages_translated_es > 0`, which is `0` for a book **written** in Spanish. 68 books / 19,489 pages outside its own denominator; it reported **exit 0, clean** over the whole failure (#4146/#4186). |

What makes it evidence for your thesis specifically: the relevant doc rule was not merely present, it was *procedural* — and it still did not fire, because the person running the detector had no reason to doubt a green result. The tell was never the exit code. It was the scope line: **107 books before the fix, 175 after.**

Applying your discriminator to my own fix, it split exactly as you predict, and the check half was the durable half:

- **Nameable symbol → check.** The worker and its audit now import one `embeddableEditionFilter`; a test pins the clause and that it is *not* the read-side filter. Negative control run — removing the `visible` / `pages_ocr` guards turns two tests red. Worth noting for your brake #1: the obvious shared filter was the **wrong** one, and importing it produced 21 phantom gaps — a check that can never go green. So "write a check" is not automatically safe either; the check has to encode the right predicate, and only running it against production told me which.
- **Human drawing a conclusion → doc.** "A green detector may be measuring a corpus that excludes the failure" cannot be asserted. That went to `measurement-instruments.md` as #4193, alongside the existing *"verify a detector by making it fail"* — which this case defeats, since breaking a book *inside* the scope goes red and proves nothing about the books outside.

One suggestion, take it or leave it: the reflection step might ask specifically **"did a DETECTOR's scope need to move?"** A detector is a writer-shaped thing that silently changes meaning when a predicate widens, and it fails in the most expensive direction available — quietly, while displaying green.

Issue #4149, opened 2026-08-21
OCR fabricates a full invented page for BLANK leaves — ~678 pages est.; root cause is the prompt; write-guard shipped


JDerekLomas:
> **Status 2026-08-21.** This issue opened describing one book and ~1,100 characters. It is
> now measured, the root cause is understood, and the write path is guarded. Original report
> preserved at the bottom; the four comments carry the working. Read this header first.

## What this is

Shown a blank or unreadable leaf, the OCR model does not decline. It emits fluent
period-appropriate prose with a complete invented apparatus — `<page-type>text</page-type>`,
a running header, a signature mark, a decorative initial and **a `<page-num>`**. A fabricated
page number is a fabricated citation, on published pages that readers, exports, the MCP tools
and the embeddings lane all serve.

## Root cause — the prompt, not the model's prior

Two defects, both in `prompts/ocr/standard-ocr-v5.md` and `v10.md` (the versions every
confirmed fabrication ran under) and still in the live v14/v15:

1. **`<page-type>` is defined with nothing.** Its enum — `blank` included — was appended to
   the `<columns>` bullet, a tag about counting text columns. It is the only metadata tag in
   that list with no definition of its own, which is why all 50 fabricated pages carried
   `<page-type>text</page-type>` and none carried `blank`.
2. **The prompt supplies a complete specimen page.** Rule 2's example is
   `"DISCURSUS IV." → <header>DISCURSUS IV.</header>` — and "DISCURSUS IV." is the running
   header in **every** fabrication across eleven unrelated books. With line 38's `A2, B1`
   signature example and rule 3's drop-caps, a model with nothing to read has a template to
   fill. It is not reaching into training data; it is echoing its own instructions.

`getOcrPromptFromDb()` passes `prompt.content` verbatim (only `{language_instruction}` and
`{language}` are substituted), so the model receives these lines as written.

## Measured

| | |
|---|---|
| books screened | 21,481 live |
| method | text screen, then ink coverage on the page image; seeded random subsample |
| screen precision | 1.14% (95% CI 0.36–1.93%) |
| **estimated population** | **~678 pages (CI 211–1,145), ~0.4% of books** |
| models | `gemini-3.1-flash-lite-preview` **and** `gemini-3-flash-preview` |
| write paths | `ai`, `pipeline_preview`, `batch_api` — all three |
| languages fabricated | Latin, German, Sanskrit, English+Latin |

[...]

PR #4192, opened 2026-08-21
docs(invariants): secret-lover reports an unreadable secret as a missing one

JDerekLomas:
## Why

Two credential-injection traps cost most of a session on 2026-08-21, and **neither fails loudly** — both hand you an empty result that reads like a finding.

**1. `secret-lover` reports an UNREADABLE secret as a MISSING one.** `secret-lover run` prints `Warning: N secret(s) not in Keychain (skipping)` and then **runs the command anyway** with those variables unset, exit 0. The worker fails later, far from the cause. `secret-lover get` says "not found" for a name `list --all` shows under that exact project — **`list` and `get` disagreeing is the tell**. `secret-lover check` reported "All checks passed / Keychain read/write working" throughout a total outage. One-call diagnostic: read a *global* secret; globals working while project-scoped ones fail isolates it to the auth session.

`clear-auth` from a non-TTY agent shell destroys a working session and needs a human at a real terminal to restore. Documented so nobody else runs it to "reset" something.

**2. A worktree gets ZERO secrets.** The project is derived from the directory holding `.secrets.json`, so from `.claude/worktrees/<name>/` secret-lover looks for a project literally named `<name>`, finds nothing, and skips **every** secret — `SUPABASE_DB_URL` included, which exists nowhere else.

This one earns a doc because it collides head-on with the repo's own rule that **all feature work happens in worktrees**: a pipeline worker launched from a worktree runs uncredentialed and does not fail. The fix is to keep cwd in the main checkout and pass the worktree's script by absolute path — relative imports resolve from the script file, so the worktree's code runs under the main namespace.

**3. A parent-set variable shadows `--env-file`.** `secret-lover run` sets the variable in the parent environment and `node --env-file` will not override it. In August a stale Keychain key shadowed the working env-file one on **both** arms of the usual `TIER3 || KEY` fallback, so the fallback could not save it and the fault arrived as 175 identical Gemini 400s — one per book — reading like every book breaking at once. Rotating a key means rotating it in every store that injects it, and verifying by use with a positive control in the same sweep.

## Also in here

**Corrects the Hetzner auto-pull cadence in `CLAUDE.md`.** It claimed "every ~5 min"; the box's crontab says `17 * * * * … auto-pull.sh` — **hourly**. Verified 2026-08-21 by reading the crontab. This matters because a `scripts/**`-only merge can sit up to 59 minutes before going live, and that lag reads as a failed deploy.

## Scope

Docs only — no code paths touched. `CLAUDE.md` goes 236 → 237 lines, inside the ~250 budget: the new material is a subsystem-triggered invariant, so it lives in `.claude/docs/invariants/credential-injection.md` with a one-line routing entry, per the tiering rule.

Out of scope, deliberately: the ~678-page fabricated-OCR backfill (#4149, guard already shipped and wired into both writers), and the 22,408 books holding page images with zero OCR (the paused line, #3826).

https://claude.ai/code/session_01H4WkdDTCyUfnzoh2DwBarf


JDerekLomas:
Independent corroboration from another session today, plus one measurement that rules out the hypothesis a reader is most likely to reach for.

I hit this running `scripts/audit/page-texts-coverage.mjs`: it exited **2** with `Missing MONGODB_URI or SUPABASE_DB_URL`, twice in a row, having succeeded with the identical command minutes earlier.

The obvious diagnosis is a flag interaction — `secret-lover run` vs `node --env-file` — and it is wrong, but it looks right, because the bare probe works while the `--env-file` invocation fails:

```
secret-lover run -- node --env-file=.env.production.local -e '…'
  MONGODB_URI: set(93)
  SUPABASE_DB_URL: EMPTY          ← from secret-lover, not the file
secret-lover run -- node -e '…'
  OK len=87                       ← same secret, moments later
```

`.env.production.local` has **no** `SUPABASE_DB_URL` line at all, so `--env-file` cannot be shadowing it. The variable was simply unreadable on those runs.

**The discriminator is the banner, and it is worth naming in the invariant:** every failing run carried `Warning: 22 secret(s) not in Keychain (skipping)`; the successful run carried none. So the check is `grep -c "not in Keychain"` on the output, not the exit code and not the flags. A caller that only reads the exit code sees "missing config" and starts editing env files.

Third retry succeeded with no banner and exit 0 — consistent with transient Keychain state rather than anything about the invocation.

Two things this adds to the writeup, if useful: the failure is **intermittent within a single session**, so "it worked ten minutes ago" is not evidence against it; and the affected script did the right thing by exiting `2` rather than reporting clean, which is the `measurement-instruments.md` three-value contract earning its keep — worth citing here as the shape that contained the damage.

Issue #4071, opened 2026-08-19
Three scheduled detectors have never had a database — no MONGODB_URI repo secret


JDerekLomas:
## What

`gh secret list` on this repo returns exactly four secrets:

```
CF_ZONE_ID, CLOUDFLARE_API_TOKEN, CRON_SECRET, HETZNER_SSH_KEY
```

There is **no `MONGODB_URI`**. Three scheduled jobs reference `secrets.MONGODB_URI` and therefore receive an empty string:

| workflow / job | cadence | state |
|---|---|---|
| `corpus-integrity-watch.yml` → `alignment-sample` | weekly, Mon 07:00 UTC | never sampled a book |
| `corpus-integrity-watch.yml` → `feedback-clusters` | daily, 08:00 UTC | never clustered a report |
| `field-sprawl-watch.yml` → `field-sprawl` | weekly, Tue 07:30 UTC | never counted a field |

(The PR-time `field-write-lint` job in the same workflow needs no database and is genuinely working — that half of the #3969 ratchet is fine.)

## How it hid

Both failure directions were available and both fired.

**Loudly wrong.** `bulk-archive-alignment.mjs` exited `1` for "no MONGODB_URI" *and* for "found misalignment", and the workflow files a finding issue on `1`. So it filed **#3572 (Aug 3), #3862 (Aug 10), #4009 (Aug 17)** — three consecutive Mondays of "Image/text misalignment found in weekly sample" whose bodies are the script's own `MONGODB_URI not set` message. `field-sprawl-watch` filed on `fired != '0'`, sweeping its script's exit `2` into the same branch → **#4023**. Four issues, zero measurements.

**Silently wrong.** `feedback-symptom-clusters.mjs` exits `2` and its job files only on `1`, so it has been green every day since July while never reaching the database. No issue, no alarm — the shape #3740 is about.

Because `set +e` swallows the exit code, every run reports `success` in the Actions UI. The four filed issues were the *only* visible signal, and they pointed at the corpus rather than at the harness.

## What is already fixed

PR (this issue's companion) makes the exit code a contract — `0` clean / `1` finding / `2` could-not-run — across all three scripts, files findings only on `1`, and fails the job red on anything else. The four false-positive issues are being closed with a pointer here.

That stops the misreporting. **It does not restore the detectors** — this issue stays open until the secret exists.

## What is needed (Derek — Atlas + GitHub console)

[...]

PR #4194, opened 2026-08-21
feat(archive): the MDZ master gap — 100,188 pages that serve from R2 but whose originals we don't hold

JDerekLomas:
**Dry-run only. Nothing fetched, nothing written.** Opened as a draft because the scope question below is a decision, not a detail.

## The finding, which matters more than the script

"Is everything archived?" has **three** answers, and the middle one is invisible to the question as usually asked. Here is a real page from *Cosmographey*:

```
photo           api.digitale-sammlungen.de/.../full/full/...   (MDZ, full-res)
archived_photo  (absent)                                        ← no master
display_photo   images.sourcelibrary.org/.../0040.jpg           (1200px, ours)
image_thumb     images.sourcelibrary.org/.../0040-thumb.jpg     (150px, ours)
```

That page serves **100% from R2** — so it counts as "served from us", correctly — while the only full-resolution copy of the scan sits on Munich's server. If MDZ changes a URL scheme or withdraws an item, we keep serving 1200px forever and can never regenerate anything larger.

| state | meaning |
|---|---|
| master on R2 (`archived_photo`) | fully preserved |
| **derivatives only** | serves fine, **we do not hold the original** |
| nothing on R2 | 24,488 pages — the gap previously reported |

**So the real job is 100,188 pages across 428 books**, not the 16,434 across 64 that a "no R2 object" count reports. The middle category is ~75K pages that every serving-side measurement calls healthy.

## Why this is a decision and not a run

Measured MDZ `full/full` sizes: **185 KB, 428 KB, 582 KB, 2.56 MB** — call it ~900 KB average. 100,188 pages ≈ **90 GB**, about **+0.4%** of R2's 21.18 TB, **~$1.35/month** ongoing, plus a long crawl against a partner institution's servers. Cheap in absolute terms, but it is paid, ongoing, and outward-facing.

Three options:

1. **Masters for everything** — 100,188 pages, ~90 GB. Full preservation.
2. **Only the 24,488 with nothing on R2** — closes the serving gap; ~22 GB. Leaves ~75K pages served-but-unpreserved.
3. **The 10 largest books first** (~12,500 pages, ~11 GB) to turn the size estimate into a measurement before committing.

## The driver

- **Checkpointed per book** — corpus-walk rule. An interruption costs one book. (The thumb repointer in #4189 learned this the hard way: its single-query version died on `ECONNRESET` after ten minutes.)
- **Reads progress from Mongo after every call, never from the HTTP response.** `/api/books/<id>/archive-images` returns Cloudflare 524s while the function keeps working to its 300s `maxDuration` — its own source comments say so — so the response is not the measurement.
- Stops a book after two calls with no progress rather than looping.
- `--max-books=N` for a cautious first run.

## Follow-up worth more than this script

**A detector for "pages with derivatives but no master."** This entire class is invisible to every check we have, because every serving-side measurement reports it healthy. That is the guardrail; this is the one-off — and by the ratchet rule now in #4191, the guardrail is the thing that should outlive the session.

### DIFF
--- a/.github/workflows/master-copy-watch.yml
+++ b/.github/workflows/master-copy-watch.yml
@@ -1,0 +1,74 @@
+name: Master copy watch
+
+# The detector for a class every other check calls healthy.
+#
+# A page can serve entirely from our own CDN — our 1200px display variant, our
+# 150px thumbnail — while the only FULL-RESOLUTION copy of the scan sits on the
+# source institution's server. It renders, it counts toward "served from R2",
+# and no broken-image sweep touches it. If that source changes a URL scheme or
+# withdraws the item, we keep serving 1200px forever and can never regenerate
+# anything larger.
+#
+# Measured 2026-08-21: a corpus census reported 99.56% of live pages "held on R2
+# and served from R2" — true, and it missed ~100,000 pages in exactly this state.
+# Nothing we ran would have noticed, which is why this exists.
+#
+# READ-ONLY, and findings go to an issue rather than failing a build. Repair
+# fetches from the source and costs bandwidth and R2 storage, so it is a
+# decision a person makes, never something a cron does on its own.
+#
+# Weekly, not daily: it is a full scan of ~19M page documents, and the number it
+# watches moves in weeks, not hours.
+
+on:
+  schedule:
+    - cron: '0 6 * * 2'   # Tuesdays 06:00 UTC
+  workflow_dispatch:
+    inputs:
+      threshold:
+        description: 'Pages above which this is a finding'
+        required: false
+        default: '0'
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  master-copy:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+      - run: npm ci --ignore-scripts
+      - name: Look for pages served without a master
+        id: run
+        env:
+          MONGODB_URI: ${{ secrets.MONGODB_URI }}
+        run: |
+          set +e
+          node scripts/audit/pages-served-without-a-master.mjs \
+            --max=${{ github.event.inputs.threshold || '0' }} > report.txt 2>&1
+          echo "fired=$?" >> "$GITHUB_OUTPUT"
+          cat report.txt
+      - name: File findings
+        if: steps.run.outputs.fired == '1'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          TITLE="Pages served from R2 with no master held — $(date -u +'%Y-%m-%d')"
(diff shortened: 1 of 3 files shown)


JDerekLomas:
**Now running, and the detector is built.** Two updates.

## 1. The archiving is running — all 100,188 pages

Confirmed on the way in that the gap was real and not a field-naming artifact. New-era R2 stores masters as `pages/<book>/NNNN-full.jpg`, so "no `archived_photo`" could have meant "old field unused". Curled it first:

```
before:  0040-full.jpg → 404
after:   0040-full.jpg → 200, 2,555,758 bytes
```

404 on every book sampled, then a real 2.5 MB master after the first call. The gap was genuine.

The first call also vindicated the one design decision that mattered:

```
Cosmographey  http=524  1616 -> 1502  (+114)
```

Cloudflare **524**, and 114 pages archived anyway. A driver trusting the HTTP response would have logged a failure. Reading progress from Mongo is what makes it correct. Steady since — ~110 pages/call, now at `222 -> 117` on that book.

## 2. New: a detector for the class, which is the part that outlives this

`scripts/audit/pages-served-without-a-master.mjs` + weekly `master-copy-watch.yml`.

**First production run: 89,475 pages across 486 live books.**

| provider | books | pages |
|---|---:|---:|
| iiif (overwhelmingly MDZ) | 446 | 89,231 |
| leiden | 18 | 116 |
| mdz | 10 | 63 |
| gallica | 2 | 42 |
| ndl_japan | 2 | 11 |
| e-rara | 7 | 8 |
| loc | 1 | 4 |

Two things make it a guard rather than a decoration:

- **It checks both storage eras** — `archived_photo` (old), `photo → …-full.jpg` (new), `cropped_photo` (split halves). Checking one era reports this whole class as a naming artifact.
- **Exit 2 fails the workflow loudly.** Per the `corpus-integrity-watch` contract: 0 clean, 1 finding → files an issue, 2 → error. #4071 is why — three detectors sat blind behind a missing `MONGODB_URI`, and a silent green run is indistinguishable from "checked, found nothing".

Weekly, read-only, and the issue body states that repair crawls a partner's infrastructure and is a decision rather than a cleanup.

**Known limit, named rather than hidden:** not every source is repairable by the current route. `digital.slub-dresden.de` shows up in the findings and is absent from the route's `ARCHIVABLE_SOURCES_REGEX`, so those books will stall. The driver handles that by stopping a book after two calls with no progress and deliberately *not* checkpointing it, so it surfaces instead of being silently skipped.

## Why this matters beyond the number

I reported "99.56% held and served from R2" this morning. Held-and-served was true; **held-at-full-resolution was not**, for ~90,000 pages. Same corpus, two different questions, and only one of them had a check. That is the case for the detector: the archiver fixes 100K pages once, the detector notices the next 100K.

Ready to un-draft once the run completes and I can report actual bytes from `archive_metadata.bytes` rather than my 4-image estimate.

PR #4201, opened 2026-08-22
User lists: save books, pages, and images to your own collections

laubaumau:
## What

Signed-in readers can save items to named lists, playlist-style:

- A bookmark button next to the existing like button on the book page (both action rows of the v2 layout) and the gallery image page opens a **Save to a list** modal: check/uncheck your lists, or create a new one inline. Anonymous visitors who click it get the sign-in prompt (same AuthPrompt likes use for their nudge).
- **/lists** shows your lists as cards with 4-thumbnail cover collages (sign-in invitation when logged out); **/lists/[id]** lets the owner rename, describe, flip visibility, remove items, and delete the list. Items link out via the same URL shapes /favorites uses.
- **Lists are signed-in only** — unlike likes, which keep their anonymous dual identity. A list is durable curation, so it lives on an account, not a device-local visitor id. All list routes 401 without a session.

## Data model

New `user_lists` + `user_list_items` collections — deliberately not fields on `users`/`books` (field-sprawl.md). Index specs added to `scripts/maintenance/ensure-indexes.mjs` (run with `--apply` after merge): unique `user_lists.id`, owner+updated, unique `list_id+target_type+target_id`, list+added_at. Items are validated to exist at add time (same lookup as the likes toggle) and capped (100 lists/owner, 500 items/list). `items_count` is recounted, not $inc'd, so it self-heals.

## Safe defaults & tenant lockdown

- **Visibility defaults to private**; the toggle's labels say what public means ("anyone with the link"). A public list page shows title/description/items only — **never the owner's name** (safe-defaults.md, #3505 shape). Public lists are readable by anyone with the link, signed in or not.
- Private lists **404** (not 403) to non-owners, matching the embassy-threads pattern.
- **No `/api/[tenant]/lists` twin, on purpose**: list pages link to global `/book` and `/gallery` URLs, which leak off partner subdomains (tenant-lockdown invariant 6). Instead of half-building a twin that drifts, the SaveToListButton hides itself on embed routes, iframes, and tenant subdomains via `useIsEmbedded()` (host-based, not `usePathname` — per the #3383 lesson), and it's only mounted in the non-embed return of `book/[id]/page.tsx` (the two-returns trap, #3916).

## Out of scope (deliberate)

- The bookmark icon doesn't pre-fetch "is this saved anywhere" on page load (that'd be a per-user query on every book view); it fills after you open the modal. A batched status endpoint can come later if wanted.
- No public discovery/browse of readers' public lists — public means link-shareable only for now.
- Reordering items within a list.


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek** — a new user-visible feature (user lists) with `auth()`-gated routes and POST-based deletes on two new collections, which is on the hold list twice over. Nothing on main supersedes it (likes/favorites is a different feature). Mergeable clean as of today.

PR #4212, opened 2026-08-23
Add a Slime Moulds collection page

laubaumau:
Adds `/collections/slime-moulds`, a sub-collection of Fungi & Mycology covering the Mycetozoa.

### The books

Three works of the group itself:

| Work | Pages | Language | State |
|---|---|---|---|
| Panckow, *Herbarium Portatile*, 1656 | 622 | German | already in the library, was hidden and filed under `astrology` |
| de Bary, *Die Mycetozoen*, 1864 | 168 | German | newly imported |
| Rostafiński, *Śluzowce (Mycetozoa). Monografia*, 1875 | 466 | Polish | newly imported, first-translation candidate |

Panckow 1654 carries the first published notice of a slime mould (the 1656 Leipzig edition is what is digitised). de Bary takes the group out of the fungi and names it Mycetozoa. Rostafiński writes the first monograph, in Polish, and it reaches the field mainly through Lister's 1894 English rearrangement.

Plus the four general mycological works already in the library that carry substantial myxomycete matter: Micheli 1729, Persoon 1801, Fries *Systema* vol 3, and de Bary's 1866 *Morphologie*. Those are already translated, so the page has material from day one.

Membership is pinned by `scripts/import/create-slime-moulds-collection.mjs` (already run against production) because the phase 7.6 classifier is paused. The acquisition batch is recorded in `scripts/import/slime-moulds-import.json`.

### The page

The mycology page's skeleton pointed at the new slug, per `.claude/docs/collection-page-redesign-spec.md`. No new design primitives. Differences:

- **No quote band.** It needs real passages from these books' own OCR and translations, and the batch jobs are still running. Left out with a note rather than filled with anything else.
- **Introduction figure is data-driven** rather than a hand-picked asset, so it appears once image extraction produces a plate.
- **Anchors are computed** from the sections that actually render, so the page degrades cleanly while the pipeline fills it in.
- **Featured work is pinned to Rostafiński by slug** rather than picked by translated-page count, so the authored copy can never sit under a different title.

### Shared components

`MycoAnchorBar` and `LibrarianSearch` move out of the mycology route folder into `src/components` (as `CollectionAnchorBar` and `LibrarianSearch`) so both pages share one copy. `MycoSlider`/`MycoMasonry` were already thin re-exports, so the new page imports `BookSlider`/`GalleryMasonry` directly.

The librarian search input also goes to 16px on mobile (`text-base sm:text-sm`), which stops iOS zooming the page when it gets focus. Desktop is unchanged.

### Still to come

OCR batches are running for all three books. The two large ones were submitted in 250-page chunks, so each needs one more chunk once the first collects. Then translation, image extraction, the first-translation flag on Rostafiński, and publishing.


JDerekLomas:
**Triage finding (read-only backlog sweep, 2026-08-28): real regression, not a flake.**

`test` fails on this PR's head commit (run 33031397885). The failures are in `tests/unit/book-cover-loader.test.ts` (3 failed — mobile-width thumb-swap assertions) and this PR directly modifies `src/lib/book-cover-loader.ts` (part of the gallery/collection-image-curation work added here), so this reads as a real regression introduced by the change, not pre-existing flake. Labelling `blocked` so it isn't merged on the strength of the Vercel/DCO checks alone — the `test` line is the one that matters and it's red for cause.

PR #4257, opened 2026-08-27
fix: records that are complete from the writer's side and incomplete where the site looks

JDerekLomas:
Follow-on from #4227, from one question: *"is all the metadata captured, such as holding library?"* No — and the way it was missing generalised well past the four books that prompted it.

## The shape of every bug here

A record can be complete from the writer's side and incomplete where the site looks. None of these throws, none shows up in a count of books, and each is only findable by asking the READ path what it needs.

| what was recorded | where the site looks | consequence |
|---|---|---|
| `provider: 'byu'` | the `ImageSourceProvider` union → `LIBRARY_PARTNERS` | no `/libraries` page, no credit at all |
| call number in `dc_identifier` | `image_source.shelfmark` | classmark never shown |
| `license: 'Public domain'` | `IMAGE_LICENSES`, keyed by **id** | rendered as a raw string |
| `book_visible` absent | `{ book_visible: true }` — an **exact** match | **27,115 illustrations invisible** |

## 27,115 extracted illustrations that existed and could not be seen

`gallery_images` denormalises `book_visible` from `books.visible`, and ~15 read paths filter on it exactly — the gallery, collections, artist pages, hero mosaics, and **a book's own illustration strip**. A row where the field is *absent* fails that filter. Extraction had run; the reader was told the book had no illustrations.

- **The tell** it's a bug and not a design: exactly one path, `/api/search/unified`, uses `{ $ne: false }`, which *does* match absence. The same row is findable in search and missing from the gallery.
- **The trap:** `dedup-clean-gallery.mjs` sets `book_visible: false` to suppress images it judged DUPLICATES — the same field. A blanket "set true where the book is live" would have un-suppressed every de-duplicated image. 436 rows carrying `is_duplicate`/`dedup_hidden_at` deliberately untouched.
- **Root cause:** the reconciler still exists (`/api/admin/sync-gallery-images`); its cron was archived under "replaced by Hetzner workers" and no worker took it over. A denormalized field whose refresher was retired drifts silently, because nothing about the read path looks broken.
- 26,787 restored. It also answers the 2026-08-25 feedback *"How to see all illustrations?"* — the whole illustrations section is gated on `galleryPlates.length > 0`, so it never rendered. There genuinely was no way.

## `iiif` is a protocol, not a library

The audit reported 1,380 live books with no library page, mostly `iiif`. Adding an `iiif` entry would credit a standards body for the Bayerische Staatsbibliothek's scans. Grouping by the **host** of the manifest gives **fifteen** institutions and 16,994 records (the audit said 997 because it only counts live books). Twelve already had a key *and* a page and had simply never been pointed at.

**16,993 reassigned; "no library page" 1,380 → 10.** New pages: SLUB Dresden, SUB Göttingen, ETCSL, the Morgan, Tartu, B-NICE, Wikimedia Commons.

## The audits, which are the durable part

- **`scripts/audit/record-completeness.mjs`** — read-path completeness by provider (importer-shaped batches, so one bad importer is one row), plus duplicate clusters whose copies **differ**. 898 editions have >1 copy; redundancy is harmless, so only the **219 that materially differ** are reported, ranked by what a reader loses landing on the thinner one. Worst: two *live* copies of *Historia Critica Philosophiae*, one with 1,416 pages OCR'd and translated, the other with 24.
- **`scripts/audit/gallery-denorm-drift.mjs`** — separates ABSENT from STALE and never touches a dedup suppression.

Both parse their vocabularies out of the TypeScript and **check the parse**, after two bugs of the same shape in one function: a `;` inside a comment truncated the provider union (18 of 60 — would have called three quarters of the corpus invalid), and stripping `//` to end of line shredded the licence alias table (every `'http://…'` key became `'http:`). A parser that silently under-reads manufactures findings rather than missing them.

## Two judgement calls

**The licence fix is in the renderer, not the data.** 3,740 books store something outside our seven ids and almost none of it is wrong — sources record their own rights statement and the importers kept it faithfully. Rewriting would be lossy and in one case false: 1,260 books store `rightsstatements.org/vocab/NoC-NC` = *"No Copyright – Non-Commercial Use Only"*, which is **not** public domain. A test pins that. The audit was then corrected to count what a reader sees (585) rather than the data shape (3,740) — a check that reports a fixed problem as broken is one people learn to ignore.

**Absence has more than one cause.** 121 live books had no cover; 105 are ETCSL, a text corpus with no page images, where coverless is correct. Ten were real.

## Also in here

Calkiní imported (26 pages, split by hand from four-up photostats, page order read off the leaves and independently confirmed by the OCR's `<page-num>` — 25 agree, 0 disagree), Teabo, Nah and Ixil imported and OCR'd, and the BYU/Gates provenance completed.

## Out of scope

7,899 books with no categories (needs classification, costs money) · the 219 diverging duplicate clusters (curatorial; `--json` makes a worklist) · Tizimín (sparse foliation, needs the OCR route) · **re-arming a scheduler for `sync-gallery-images`, or the drift returns.**

## Verification

`npx tsc --noEmit` clean · `npx vitest run` — 217 files, **3,013 tests**, all passing. Every data change measured before and after; the Fludd images Derek originally couldn't find are now reachable.


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek**, leaning cherry-pick. Git says CLEAN but merging would break the build: all seven provider keys this adds (`etcsl`, `goettingen`, `slub_dresden`, `b-nice`, `wikimedia_commons`, `tartu_dspace`, `morgan`) already exist on main via #4615/#4585/#4558 — duplicate object keys in `image-source.ts` / `library-partners.ts`. Still unique here: `record-completeness.mjs`, `gallery-denorm-drift.mjs` (the ABSENT direction of `book_visible` drift), `backfill-missing-covers.mjs`, the licence-alias table. Those want a fresh PR of their own.

JDerekLomas:
Superseded by #4980, which merged. That PR's seven ImageSourceProvider keys (etcsl, goettingen, slub_dresden, b-nice, wikimedia_commons, tartu_dspace, morgan) already landed on main via #4615/#4585/#4558, so merging this branch as-is would have produced duplicate object keys and broken the build. #4980 cherry-picks the still-unlanded, still-wanted parts: scripts/audit/record-completeness.mjs, scripts/audit/gallery-denorm-drift.mjs, scripts/maintenance/backfill-missing-covers.mjs, and the LICENSE_ALIASES/licenseDisplay() licence-display fix (with its test).

Issue #4286, opened 2026-08-27
Gallery image pages are near-invisible to crawlers: no SSR image, template og:image, no crawlable pagination (from MCP agent audit)


JDerekLomas:
From an MCP agent session (research on a Fludd plate that turned into a crawlability audit). Claims below were spot-verified before filing; verification status noted per item.

## Verified (curl with browser UA, 2026-08-27)

1. **Gallery image detail pages ship no plate `<img>` server-side.** `/gallery/image/6a19bfb49b13f9ea2b27284b-0` server HTML has 4 `<img>` tags, none pointing at `images.sourcelibrary.org` — no image, no caption body, no page link in the served HTML. Meta description is good; body is empty for a crawler. This is the canonical citable unit for a plate and it is the least indexable page class on the site (~222k pages).
2. **`og:image` is a generated template card everywhere** — `/gallery/image/<id>/opengraph-image`, never the actual plate JPEG. Same on `/book/<slug>`. A 222k-image collection presents a template card in every link preview.
3. **Correction to the report:** JSON-LD is NOT absent — gallery image pages carry a schema.org `WebPage`/`CollectionPage` graph. What's missing is `ImageObject` with `contentUrl`/`caption`/`isPartOf`, which is the part that would matter for image surfacing.

## Reported, not yet verified

4. `/gallery` listing ends at a "Load more" control with no paginated URLs — a crawler sees ~45 of 222k images. Suggested: image sitemap with `<image:loc>` + `<image:caption>` (descriptions already exist), or `/gallery/page/N`.
5. Thumbnail grids (gallery, book Illustrations section) expose no resolvable `src` in server HTML (likely lazy-load/background-image) — check view-source.
6. `images.sourcelibrary.org` as a separate host: robots.txt + Search Console verification gate all of the above.

## Bot policy (policy call for Derek, see `.claude/docs/invariants/crawler-access-gate.md`)

7. `tdm-reservation: 1` + blanket posture doesn't distinguish **training** crawlers (GPTBot, Google-Extended) from **retrieval** crawlers (OAI-SearchBot, ChatGPT-User, PerplexityBot). Allowing retrieval while disallowing training keeps the reservation honest without making the library invisible to assistants that send scholars here. (Report also suggested "an llms.txt" — we already have one; the audit missed it, which is itself a discoverability data point.)

## Related
[...]

PR #4295, opened 2026-08-27
fix(gallery): serve the actual plate to crawlers — og:image + noscript figure on image pages

JDerekLomas:
Implements the two verified, low-risk halves of #4286 (MCP agent crawlability audit):

1. **og:image / twitter.images = the actual scan** instead of the generated template card (which was identical-in-kind across ~200k pages). Falls back to the file-convention card when no image URL resolves. `twitter.images` set explicitly per the shallow-metadata-merge invariant.
2. **noscript figure in the layout** — img + caption + museum description + book/page link. The layout already fetched all of this server-side for JSON-LD; now the raw HTML (what non-JS crawlers and LLM retrieval bots actually read) carries the content instead of an empty body.

Out of scope (rest of #4286): gallery pagination / image sitemap, thumbnail-grid SSR, bot-policy split, images.sourcelibrary.org Search Console — those need their own decisions.

Verification: `npx tsc --noEmit` clean. Post-merge: `curl -A <browser UA> https://sourcelibrary.org/gallery/image/6a19bfb49b13f9ea2b27284b-0?cb=1 | grep -o 'og:image[^>]*'` should show the archived JPEG, and the noscript block should contain an images.sourcelibrary.org img.

Refs #4286

https://claude.ai/code/session_01EcC7qVRrjjrt34nMu2vF92

### DIFF
--- a/src/app/gallery/image/[id]/layout.tsx
+++ b/src/app/gallery/image/[id]/layout.tsx
@@ -211,6 +211,7 @@ export async function generateMetadata({
   const author = page.book?.author;
   const year = page.book?.published;
   const description = detection.description || 'Historical illustration';
+  const plateUrl = (page as { enhanced_photo?: string }).enhanced_photo || page.cropped_photo || page.archived_photo || page.photo;
 
   // Short title: first sentence (up to 70 chars) for social card headline
   const firstSentence = description.split(/\.\s/)[0];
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Post-merge accuracy note from preview verification: `twitter:image` and the noscript figure serve the raw plate URL as intended, but **`og:image` still serves the file-convention `opengraph-image` card** — Next's file-based metadata outranks `generateMetadata`'s `openGraph.images`, so that half of the change is inert. Mitigating fact found while verifying: the generated card is not a generic template — it embeds the page scan itself (letterboxed 1200×630 PNG), so og consumers do see the artwork; what they don't get is the raw stable JPEG URL. If we want that, the follow-up is to make `opengraph-image.tsx` serve/redirect to the plate JPEG (or delete it and let metadata win). Left open on #4286.

Issue #4285, opened 2026-08-27
Text-grounded dedup verification: page-text comparator — pilot finds 27% of same-edition-key pairs fail the content check


JDerekLomas:
## Idea (Derek's, 2026-08-27): test whether deduping works by reading the text

Every dedup test we have judges metadata against metadata, so the tests share the blind spots of the system under test. But we own OCR for 6.3M pages — the content itself can be ground truth. Pilot built and run 2026-08-27; results below. Raw method: for a pair of books, sample interior pages at 25/40/55/70/85% of each book's length, normalize Unicode-aware (`\p{L}\p{N}`, NFC — never `\w`), compare character-4-gram sets, search an alignment window of ±(|ΔpagesCount|+8) on the second book, take the median of per-sample best matches. **A sample only counts when BOTH sides produced ≥200 chars of text** — absence is recorded as no-evidence, never as disagreement (a copy of Monas with empty OCR at the sampled pages initially read as "not the same book").

## Calibration (controls first)

- **Positives** (known copies via `duplicate_of`, both OCR'd): all true copies scored ≥0.55 (typically 0.6–0.9; OCR noise keeps it off 1.0). One "positive" scored 0.11 — and inspection shows its sides are dated **1532 vs 1585-86**: a mislabeled `duplicate_of` link (different editions recorded as copies, books `69d3bdb1…Pymander` pair — see control file). The metric caught a wrong link in its own control set.
- **Negatives** (same work, different edition): 0.04–0.38. One "negative" scored **0.73** (De arte gymnastica, different year strings) — a candidate *missed* duplicate wearing divergent metadata.
- Working thresholds: **≥0.5 same printing, <0.3 different content, between = gray.**

## Pilot result: all 225 both-visible full-quality same-edition clusters (259 keeper↔copy pairs)

| Verdict | Pairs | Share |
|---|---|---|
| Confirmed copy (text matches) | 128 | 49% |
| **Suspect — text does NOT match, with evidence** | **71** | **27%** |
| Gray (0.3–0.5) | 31 | 12% |
| Insufficient text (skip, recorded) | 29 | 11% |

The 71 suspects with both-sides evidence break into recognizable classes:

1. **Fragments/excerpts under a full edition's key** — *Summa de Arithmetica* 626pp vs 26pp; *Didone* 448/96; Ficino *Epistolae* 794/254. The key has no length dimension.
[...]

Issue #4271, opened 2026-08-27
Identity-review batch lane: 1,940 LLM-screened work merges + 311 keeper choices stuck behind one-at-a-time clicks


JDerekLomas:
## Problem

The detection and screening machinery finished its job weeks ago; the drain is the bottleneck. Measured 2026-08-27:

- **`work_merge_queue`: 1,940 pending** (all enqueued 2026-08-08 by the containment lane of `merge-work-clusters.mjs`, #3759). Exactly **1** row adjudicated since.
- The LLM screening pass (`scripts/analysis/stamp-work-merge-queue-llm.mjs`, 08-10) already stamped 1,874 of them: **980 `same` / 745 `different` / 149 `unsure` / 66 unstamped**.
- **`edition_keeper_queue`: 311 pending**, zero worked (seeded by `ingest-keeper-choice-queue.mjs`, #3730 §3 — the triage report already classes 89 as mechanical keeps).

The review UI (`/curation/identity-review`) handles one row per click. At that rate the queue never drains, and the stamps expire into stale evidence.

## What to keep

Approve-actuates-inline (`/api/admin/work-merges` POST performs the merge, no downstream job reads the rows) is the right shape per the ingest-is-actuation rule — the human clicks at the point the change happens. Batch mode must preserve that: the click still directly causes the merges, just N at a time.

## Proposal

**Work-merges tab:**
1. Filter by `llm.verdict` (already in the GET route) + bulk-select.
2. **Approve-all-`same`**: render the full list first (list-first-approve, Data Protection rule), require a spot-check step (random sample of ~20 pairs shown with titles/authors/evidence for eyeballing), then run the existing inline merge per row in batches with progress + abort. All merges use the existing lib (aliases, provenance log, backup, catalog sync) — nothing new on the write path.
3. **Reject-all-`different`**: writes `status: rejected` only (no actuation) — same list-first flow, lower ceremony.
4. `unsure` + unstamped stay manual; run the stamper over the 66 unstamped rows first (pennies, flash-lite — estimate before running per standing rule).

**Keeper tab:** actuation here is a visibility flip, which is a different risk class. Start with the 89 "mechanical keep" clusters from `keeper-choice-triage.mjs` only, same list-first + spot-check flow; tossups/FT-flagged stay per-cluster manual.

[...]

Issue #4270, opened 2026-08-27
Dedup tier 2 recall gap: bilingual/romanized titles miss the edition-key prefix (blocks shadow-tier retirement)


JDerekLomas:
## Context

Since the 2026-08-08 flip (#3730 §2, PR #3802), `checkDuplicate()` tier 2 is `editionKeyTierMatches()` — a stored-`edition_key` prefix lookup. The retired ASCII title+author tier still runs as the post-flip shadow (`dedup_shadow_decisions`, `regime: 'edition_live'`), and the standing rule (`.claude/docs/invariants/edition-identity.md`) is: **a clean post-flip week deletes the shadow block and `titleAuthorTierMatches()`.** That clean week has not happened — and now we know why it can't yet.

## Measured 2026-08-27

`scripts/audit/dedup-shadow-agreement.mjs`: post-flip agreement **99.31% over 434 decisions, 3 shadow-only rows** (the regression class) on 08-26/27:

| Incoming | Old tier matched | Verdict on inspection |
|---|---|---|
| "Muqaddimah", Ibn Khaldun, no year | `مقدمة ابن خلدون (Muqaddimah)` 1858 (`69e8b29f2ff2a8dc09e78458`) | **False alarm** — incoming is a Portuguese Prolegómenos vol. 1 (`6a8f620de408ab73f484c990`), a different edition |
| "Arthaśāstra", Kautilya, no year | `अर्थशास्त्र (Arthaśāstra)` 1919 (`69ea3e63300d991d0f1d3186`) | **False alarm** — incoming is a 1991 combined-commentaries edition, 1758pp (`6a8f621eddd2f57fe0f809e3`) |
| "Summa, aurea armilla…", Fumo, no year | e-rara 1587 copy in `books_warehouse` (`69b661fdb3f4fc044154031c`) | **No import landed** — check-only; also a truncated-title-tail prefix mismatch |

All three are the old tier over-matching on a missing year ("assume duplicate"). **No duplicate was imported; no cleanup needed.**

## But the mechanism is a real recall gap

[...]

Issue #3895, opened 2026-08-11
search_concept / search_translations: no author- or work-level diversity control, so edition-heavy authors swamp cross-author surveys


JDerekLomas:
Reported via MCP `submit_feedback` (feedback `6a7ac2d930665fc3449c1071`, 2026-08-11), from a research session compiling definitional passages on "magic" across the corpus for an alt.chi paper on Renaissance natural magic.

## The task that exposed it

"25 quotes on *magia* from 25 different authors." A survey of how a period defined a term — arguably the single most natural scholarly use of a passage-search tool over a corpus like ours.

## The problem

`max_per_book` dedupes on `book_id`. But the corpus holds **many editions of the same work**: 7+ printings of *De occulta philosophia*, 6+ of *Magia naturalis*. Every one is a distinct `book_id`, so `max_per_book` happily returns all of them, and the result list comes back dominated by Agrippa and della Porta. The caller wanted author diversity and the only lever available operates one level too low.

The library *has* the layer that would fix this — `work_id` clusters editions of a work, and `author_id` canonicalises the person — but neither is exposed as a constraint on the passage-search tools.

## Asks, in the reporter's order

1. **`max_per_author`** on `search_concept` / `search_translations` — the direct fix.
2. **`exclude_authors`** filter — lets a caller iterate: run, see Agrippa dominate, exclude, run again. Cheap to implement, and useful even if nothing else lands.
3. **Work-level rather than edition-level dedup** — i.e. a `max_per_work` keyed on `work_id`. Arguably the more principled version of (1): seven printings of *De occulta philosophia* are one witness to Agrippa's definition, not seven.
4. **An author facet/filter on the passage-search tools.** `author` is already a constraint on `list_books` but not on `search_concept` / `search_translations`, which is an asymmetry a caller has no way to predict from the tool descriptions.

## Notes

- (3) depends on `work_id` coverage being good enough on the searched slice; worth measuring before choosing between (1) and (3). (1) works today off `author_id` / `books.author`.
- Related but distinct: #3700 (search_translations never got the ranking fix search_within_book did), #3699 (score normalised within the book), #3083 (MCP tool ergonomics checklist). This is a *result-set composition* problem rather than a ranking one — a perfectly ranked list still returns seven Agrippas.
- The same session's other two items are #3894 (printer-as-author) and a comment on #2916 (work-within-volume attribution).

[...]

PR #4324, opened 2026-08-28
feat(search): collapse copies and group results by work (#4300)

### LINKED ISSUE #4300: Search results should collapse copies and group by work — one result per work, editions fanned out beneath
## Who this is for

A reader searching "Monas hieroglyphica" today gets several near-identical rows — multiple scans of the same printing plus multiple editions — competing with each other and pushing genuinely different material (commentaries, related works, the English translation) off the first screen. The reader's mental model is a WORK; our result list serves them a flat pile of scans. Search is the surface where the identity stack (#3258/#3730) either pays off for readers or doesn't.

## The change

Result lists group at the **work** grain, with edition/copy structure folded underneath:

- **One primary row per work** — best edition chosen by the existing keeper signals (translation availability, scan quality, pages processed).
- **"+N other editions" fan-out** under the row (expand or link to `/work/[id]`), replacing N sibling rows.
- **Copies never appear as separate results at all**: books linked by `duplicate_of` or sharing a text-CONFIRMED `edition_key` collapse to one line with an "other scans" affordance (the rail from PR #3819 already renders this on book pages).

## Guardrails (all measured, 2026-08-27)

1. **Collapse only on verified links.** #4285's text comparator found **27% of same-edition-key pairs fail the content check** (fragments, generic-title collections, pecha volumes, Spinoza's 1670 variant printings). Collapsing on a raw key match would hide *different texts* behind one result — the search-side version of the wrongly-hidden-book failure. Confirmed pairs collapse; suspect/gray pairs stay separate rows.
2. **Work grouping inherits work_id fragmentation** (14,946 unjudged local-mints; 1,940 pending merges, #4271). Acceptable: a fragmented work shows as two rows — the failure mode is mild and self-healing as the queue drains.
3. **A count shown to the reader is a claim** (`visibility-and-stats.md`): "+N editions" must count what the target surface renders — same rule as collection cards.
4. Tenant lockdown + embedPolicy: the fan-out must not leak books a tenant subdomain doesn't serve (`tenant-lockdown.md`).

## Where

- Site search (`/search` lane over books) — the reader-facing win.
- `/api/search` semantic page lane: apply copy-collapse to result composition so one passage in a much-copied edition doesn't occupy N slots. (Agent-side per-work caps are #3895 — same principle, different surface.)
- `/es/search` inherits whatever the shared lane does (#4187: a localized lane is only offered if its API takes `lang` — don't fork logic).

## Side benefits

- **Embedding spend**: text-confirmed copies don't need their pages embedded separately (3.9M page embeddings and growing; copies are pure duplicate spend).
- **Recall**: the `edition_key_latin` romanized residue being added for #4270 is equally a lexical-search asset — "Muqaddimah" should find "مقدمة ابن خلدون (Muqaddimah)" without depending on a cataloguer's parenthetical. Wire the same field into the search index when it lands.

## Sequencing

Depends on: #4285 (text verification of the 225 clusters — the collapse allowlist), #4270 (`edition_key_latin`). Complements: #3895 (MCP diversity caps), #4271 (queue drain reduces work fragmentation), #3736 (`/work/[id]` as the fan-out destination).

Suggested order: (1) copy-collapse on `duplicate_of` + confirmed pairs only — small, safe, immediate dedup of the ugliest rows; (2) work grouping with fan-out once the merge queue has drained enough that fragmentation is rare; (3) semantic-lane composition caps.

JDerekLomas:
Implements #4300 step 1 + most of step 2: search results collapse copies and group at the **work** grain, with an honest fan-out to `/work/[id]` instead of a silent drop.

## The bug, reproduced

An external agent auditing the catalogue found four separate records of Kircher's *Musurgia Universalis* Vol. I presented as four unrelated results. Reproduced on production, and the cause was not where the issue assumed:

| lane | `q=Musurgia Universalis` before |
|---|---|
| `/api/search/unified` → `books` | **1** row (already collapsed by `work_id`) |
| `/api/search/semantic` | **6** rows — all `work_id: Q18942469` |
| `/api/search` | 1 book + 1 page (already collapsed) |

`/search` renders the `/api/search/semantic` lane as **"conceptual matches", above** the keyword results. So the keyword lanes each had their own inline `seenWorkIds` filter and the semantic lanes had none — the copies came back in through the lane nobody had patched. That is the shape this PR is really fixing: **three lanes, three private notions of "same thing"**.

## In scope

- **`src/lib/search/work-grouping.ts`** — one definition, shared by every lane. Groups only on links a human or a HIGH-confidence resolver already blessed: `duplicate_of` (a copy collapses onto its keeper) and `work_id`, folded through `work_id_aliases` so a partially-applied merge (#3759) doesn't split a work in two. Never re-ranks: the caller's sort *is* the keeper choice.
- **`edition_key` is deliberately NOT a grouping key.** #4285's text comparator read the OCR of 259 keeper↔copy pairs sharing a full-quality key and found **27% are not the same content** (fragments under a full edition's key, generic-title collections, pecha volumes, the 1670 Spinoza variant printings). Collapsing on a raw key match would hide *different texts* behind one row — the search-side version of the wrongly-hidden-book failure. Pinned by a test.
- **Applied in all three lanes** — `/api/search` (shared helper replaces its inline dedup, now also honouring `duplicate_of` and aliases), `/api/search/unified` (keyword lane, plus work-grain dedup of the semantic lane both within itself and against the keyword lane), `/api/search/semantic` (had no work dedup at all).
- **`work_group` on a row that stands in for siblings** — `{ work_id, href, editions, collapsed_in_results }` — rendered under the card as "N editions & copies of this work →" (EN + ES strings), linking to `/work/[id]`.

## The count is a claim, and it is coupled to its target

`visibility-and-stats.md`: *a card must count what its TARGET page renders.* So `editions` is **not** the number of rows we hid. `fetchWorkFanouts` counts `workEditionsCountFilter()`, added directly beside `workEditionsFilter()` — the filter `/work/[id]` builds its edition list from — and a unit test pins the two to each other so the number cannot drift from the page the link opens. `collapsed_in_results` (the local number) is reported separately and never shown as the headline.

Why a second function rather than calling the page's filter directly: measured 2026-08-28, `workEditionsFilter`'s `$or` reaches `work_slug`, which carries **no index**, so Mongo scans the whole `books` collection — **7,990 ms** for one work against **38 ms** for the `work_id` half. Fine on `/work/[id]` (ISR, 6h window); not a request-path query (`request-path-queries.md`). Dropping the `work_slug` branch is safe because `work_slug` holds human slugs (`ficino-on-the-mysteries`), never work_ids: of the 81,135 books carrying a `work_slug`, exactly **4** have `work_slug == work_id` and the `work_id` branch matches those same documents. Verified by comparing both filters over 25 sampled multi-edition works — **0 mismatches**.

## Guardrails honoured

1. **Collapse only on verified links** — `duplicate_of` + `work_id` only; `edition_key` excluded pending #4285's allowlist.
2. **Work-id fragmentation** (#4271) means a fragmented work still shows as two rows. Accepted, self-healing.
3. **The count describes the target page** — see above.
4. **Tenant lockdown** — the fan-out is suppressed entirely under any tenant context (`id`, `slug`, or `isEmbedded`). An edition census across the global library is not a partner reading room's claim to make; same gate `embedPolicy.showRelatedEditions` puts on the book page's editions rail. The *collapse* still runs on tenant hosts (it only ever removes rows).

- **Text-confirmed `edition_key` collapse** — blocked on #4285 producing the allowlist. The module has the seam for it and says so.
- **Work-id fragmentation** (#4271) — grouping inherits it; draining the merge queue is the fix.
- **Semantic-lane per-work composition caps** (#3895) — different surface, same principle.
- **`edition_key_latin` recall** (#4270) — not landed yet.
- **Volume grain** (#3708) — `work_id` is work-grain, so Vol. I and Vol. II of one set share a key and collapse to one row. This is why the fan-out matters: the collapsed volumes are one click away and the count describes exactly that page. Fixing it properly needs volume represented in the identity layer.
- **`/api/search`'s pre-existing one-passage-per-work cap** — the work dedup has always applied to page results, so `pages_only=true` already returns at most one page per work. Left exactly as it was (and the fan-out is skipped there, so the MCP passage contract and its latency are unchanged).
- **`/api/search/semantic` has no tenant scoping at all** — pre-existing; this PR only reads the tenant context to gate its own fan-out, and does not attempt the wider fix.

## Verification

- `npx tsc --noEmit` clean.
- Full unit suite: **222 files, 3,077 tests passing**, including `tests/unit/search-match-quality.test.ts` (the #4304 weak-match abstentions, untouched).
- New `tests/unit/search-work-grouping.test.ts` (16 cases) — the abstention cases are the load-bearing ones: no identity ⇒ no collapse; different `work_id`s ⇒ no collapse; shared `edition_key` alone ⇒ no collapse; a copy-keyed group can never claim an edition count.
- Before/after against the preview deploy in a comment below.

Closes #4300 (steps 1–2).

https://claude.ai/code/session_01EcC7qVRrjjrt34nMu2vF92

### DIFF
--- a/src/app/api/search/route.ts
+++ b/src/app/api/search/route.ts
@@ -14,10 +14,36 @@ import { expandLanguages } from '@/lib/language-utils';
 import { logSearchQuery } from '@/lib/search-log';
 import { stripEditorialWrappers } from '@/lib/strip-editorial-wrappers';
 import { logSearchEvent } from '@/lib/search-event-log';
+import { collapseByWork, type WorkGroupable } from '@/lib/search/work-grouping';
+import { fetchWorkFanouts } from '@/lib/search/work-fanout';
 
 export const preferredRegion = 'fra1';
 
 const MAX_PAGE_RESULTS = 25;
+
+/**
+ * Carry a book's identity fields onto a result as transients, so the
+ * work-grain collapse below can read them. Stripped before the response —
+ * `_work_id` was already doing this; aliases and `duplicate_of` join it so
+ * every lane in the app collapses on the same rule (src/lib/search/work-grouping.ts).
+ */
+function attachIdentity(result: SearchResult, book: Record<string, unknown>): void {
+  const r = result as unknown as Record<string, unknown>;
+  if (book.work_id) r._work_id = book.work_id;
+  if (Array.isArray(book.work_id_aliases)) r._work_id_aliases = book.work_id_aliases;
+  if (book.duplicate_of) r._duplicate_of = book.duplicate_of;
+}
+
+/** Read the transients back as the shape `collapseByWork` expects. */
+function identityOf(result: SearchResult): WorkGroupable {
+  const r = result as unknown as Record<string, unknown>;
+  return {
+    book_id: result.book_id,
+    work_id: r._work_id as string | undefined,
+    work_id_aliases: r._work_id_aliases as string[] | undefined,
+    duplicate_of: r._duplicate_of as string | undefined,
+  };
+}
 
 /** Strip XML/HTML tags and clean up OCR artifacts for display */
 function cleanText(text: string): string {
@@ -224,8 +250,8 @@ export const GET = withApiAuth(async (request: NextRequest, _ctx, identity) => {
         thumbnail_blob: (typedBook as any).thumbnail_blob,
         quality_score: (typedBook as any).quality_score,
       };
-      // Transient field for work-level dedup (stripped before response)
-      if ((typedBook as any).work_id) (result as any)._work_id = (typedBook as any).work_id;
+      // Transient fields for work-level dedup (stripped before response)
+      attachIdentity(result, typedBook as unknown as Record<string, unknown>);
       if ((typedBook as any).text_role) (result as any)._text_role = (typedBook as any).text_role;
       return result;
     }
@@ -260,7 +286,7 @@ export const GET = withApiAuth(async (request: NextRequest, _ctx, identity) => {
         try {
           const matchingIds = await searchBookIds(query, { limit: limit * 2 });
           const bookFilters = buildBookFilters();
-          const bookProjection = { id: 1, slug: 1, title: 1, display_title: 1, author: 1, editor: 1, thumbnail: 1, thumbnail_blob: 1, image_display: 1, image_thumb: 1, language: 1, published: 1, pages_count: 1, pages_translated: 1, doi: 1, categories: 1, is_first_translation: 1, quality_score: 1, summary: 1, reading_summary: 1, work_id: 1, text_role: 1 };
+          const bookProjection = { id: 1, slug: 1, title: 1, display_title: 1, author: 1, editor: 1, thumbnail: 1, thumbnail_blob: 1, image_display: 1, image_thumb: 1, language: 1, published: 1, pages_count: 1, pages_translated: 1, doi: 1, categories: 1, is_first_translation: 1, quality_score: 1, summary: 1, reading_summary: 1, work_id: 1, work_id_aliases: 1, duplicate_of: 1, text_role: 1 };
 
           let books: any[] = [];
           if (matchingIds.length > 0) {
@@ -533,7 +559,7 @@ export const GET = withApiAuth(async (request: NextRequest, _ctx, identity) => {
         const pageBooks = await db.collection('books')
           .find(
             { id: { $in: pageBookIds }, ...(tenantId ? { tenantId } : {}) },
-            { projection: { id: 1, slug: 1, title: 1, display_title: 1, author: 1, editor: 1, thumbnail: 1, thumbnail_blob: 1, image_display: 1, image_thumb: 1, language: 1, published: 1, pages_count: 1, pages_translated: 1, doi: 1, categories: 1, hidden: 1, quality_score: 1, work_id: 1, text_role: 1 } }
+            { projection: { id: 1, slug: 1, title: 1, display_title: 1, author: 1, editor: 1, thumbnail: 1, thumbnail_blob: 1, image_display: 1, image_thumb: 1, language: 1, published: 1, pages_count: 1, pages_translated: 1, doi: 1, categories: 1, hidden: 1, quality_score: 1, work_id: 1, work_id_aliases: 1, duplicate_of: 1, text_role: 1 } }
           )
           .toArray();
         for (const b of pageBooks) {
@@ -600,7 +626,7 @@ export const GET = withApiAuth(async (request: NextRequest, _ctx, identity) => {
           thumbnail: (book as any).thumbnail,
           thumbnail_blob: (book as any).thumbnail_blob,
         };
-        if ((book as any).work_id) (pageResult as any)._work_id = (book as any).work_id;
+        attachIdentity(pageResult, book as unknown as Record<string, unknown>);
         if ((book as any).text_role) (pageResult as any)._text_role = (book as any).text_role;
         results.push(pageResult);
       }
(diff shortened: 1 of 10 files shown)


JDerekLomas:

Preview: `https://sourcelibrary-v2-guoi3n1we-dereklomas-projects.vercel.app` (commit `ff0b6cf4`)

### The Kircher case — what a reader actually sees on `/search?q=Musurgia+Universalis`

Counted with a headless browser against both hosts, same query, same viewport:

| | production (before) | preview (after) |
|---|---|---|
| Musurgia rows on the first screen | **7** | **2** |
| distinct Musurgia *records* | **6** | **1** |
| fan-out affordance | none | `8 editions & copies of this work →` |

Production:
```
Universal Music-making (Musurgia Universalis), Vol. II
Musurgia Universalis (Universal Music)
Universal Work on Music (Musurgia Universalis), Vol. I
Musurgia Universalis, Vol. II
Athanasii Kircheri ... Musurgia universalis, sive ars magna consoni et dissoni
Universal Music-making (Musurgia Universalis), Volume I
Universal Music-making (Musurgia Universalis), Volume I     <- see note
```
Preview:
```
Universal Music-making (Musurgia Universalis), Volume I
Universal Music-making (Musurgia Universalis), Volume I     <- see note
  └ 8 editions & copies of this work →   (/work/Q18942469)
```
*Note:* the repeated "Volume I" line is a pre-existing double-render in `/search`'s adaptive layout — it appears identically on production, above, and is not touched by this PR.

### Per-lane, by API

| lane | before | after |
|---|---|---|
| `/api/search/semantic` | 6 rows, all `work_id: Q18942469` | **1** row + `work_group.editions: 8` |
| `/api/search/unified` → `books` | 1 (already collapsed, silently) | 1 + `work_group.editions: 8` |
| `/api/search/unified` → `semantic` | 4 of the 6 rows were Musurgia copies | 0 Musurgia — the slots now hold **different** material (Fludd's *Utriusque cosmi*, *Harmonik der Musik*, *Phonurgia nova*) |
| `/api/search` | 1 book + 1 page | same, now with `work_group.editions: 8` |

That last row is the point of the issue: collapsing the copies didn't just tidy the list, it gave the space back to genuinely different works.

### The count is the target page's count

`/work/Q18942469` returns 200, its own `<title>` reads **"Musurgia Universalis — 8 Editions"**, and it renders **8 distinct `/book/` links**. The fan-out says 8.

### Not-collapsed control (`q=Kircher`)

Distinct works stay distinct; only works with siblings get a fan-out:
```
Kircher Ars Magna Lucis et Umbrae (1646)              → 3 editions
Kircher: Ars Magna Sciendi (1669)                     → (none)
Magnes sive De Arte Magnetica                         → (none)
Universal Music-making (Musurgia Universalis), Vol. I → 8 editions
Obeliscus Pamphilius                                  → (none)
Egyptian Oedipus, Volume 1                            → 8 editions
The Ecstatic Journey                                  → 5 editions
Obeliscus Aegyptiacus                                 → (none)
```

### Regression checks

- **#4304 weak-match abstentions intact** — `q=Rainer Maria Rilke` still returns `match_quality: "weak"` on the preview; `tests/unit/search-match-quality.test.ts` passes unchanged.
- `npx tsc --noEmit` clean.
- Full unit suite: **222 files / 3,079 tests passing**.
- Tenant gate pinned by test: `fetchWorkFanouts` is handed a `db` stub that throws, and returns an empty map — proving the suppression happens *before* the query, not after it.

### Data measurements behind the count filter

- `workEditionsFilter` (the work page's own filter): **7,990 ms** for one work — its `work_slug` branch has no index, so Mongo scans `books`. The `work_id` half: **38 ms**.
- `work_slug` holds human slugs, never work_ids: of **81,135** books carrying one, exactly **4** have `work_slug == work_id`, and the `work_id` branch matches those same documents.
- Both filters compared over **25** randomly sampled multi-edition works: **0 mismatches**.

Issue #4406, opened 2026-08-30
R2 growth explained: the provenance bake is silently upgrading display variants 1200px→2000px (~$40/mo); bake hung 9 days; forward path unmarked


JDerekLomas:
## What this resolves

The open item "explain R2's ~165 GB/day with a flat object count" in the infrastructure cost
reference. Cause established, measured two independent ways, and the earlier "provenance bake
ruled out" verdict is **withdrawn** — the bake is the cause; it was acquitted of the wrong charge.

## The finding

`scripts/maintenance/bake-provenance-mark.mjs --all --apply` regenerates every visible page's
`display_photo` **from the clean master at `min(2000px, native)`** (`DISPLAY_MAX_W = 2000`,
quality 85) and overwrites the same R2 key.

The forward-path generator, `scripts/workers/lib/display-image.mjs`, writes display variants at
`DISPLAY_WIDTH = 1200`.

So the bake is not marking images — it is **silently upgrading the whole display corpus from
1200px to 2000px**, at 2.8× the pixels. Overwriting the same key is why the object count stayed
flat while bytes climbed.

## Evidence

| check | result |
|---|---|
| Cloudflare R2 ops analytics, Aug 12–17 | ~500,000 `PutObject`/day while `objectCount` moved +100…+2,500/day → overwrites, not additions |
| 250 random books, R2 keys with an Aug 12–18 `LastModified` | all `pages/<book_id>/NNNN.jpg`; none under `archived/`, none `-thumb` |
| `HeadObject` on those keys | R2 metadata `provenance: sl-v5` — written only by this script |
| live process on the Hetzner box | PID 278703, `bake-provenance-mark.mjs --all --apply`, started Aug 10 by `/root/run-provbake.sh` |
| run log `/root/bake-provenance-sl-v5.log` | `mode: APPLY  scope: ALL visible (16335 books)  target: display_photo @<=2000px  mark: sl-v5` |
| measured object | `pages/695aa97dbe4023bd34bd6715/0015.jpg` — 1941×3194, 720 KB, `provenance=sl-v5` |

**Controlled test** (6 pages, same master, same quality 85, only the width differs — the test the
earlier analysis should have run):

```
2000px costs 1.98x the 1200px variant; delta 336 KB per page
```

Independently: 165 GB/day ÷ 500K puts/day = **330 KB per put**. Two methods, same number.

[...]

Issue #4373, opened 2026-08-29
Direct /archived/ R2 originals are publicly fetchable unmarked — decide and gate


JDerekLomas:
Follow-up to #4366 and the clean-image-tier PR.

## The situation
`images.sourcelibrary.org/archived/...` serves the archival page originals from the public R2 bucket. These are deliberately unmarked (`bake-provenance-mark.mjs` only marks display/thumb variants — the originals stay bit-faithful for OCR, exports, deep-zoom, preservation). That's correct archivally, but it means the "clean images are a paid capability" boundary (?clean=1 on /api/image) can be walked around by anyone who enumerates `archived_photo` URLs: full-resolution, mark-free, keyless, and invisible to every app-side log (Cloudflare alone sees this traffic).

## Options (not mutually exclusive)
1. **Cloudflare rule on `images.sourcelibrary.org/archived/*`**: challenge or block non-browser traffic (no Sec-Fetch-Dest/Referer), allowing browser loads, our own server IPs/UAs, and (if feasible) keyed requests. Risk: the reader's deep-zoom, the export pipeline, and the OCR workers all fetch originals server-side — enumerate the consumers FIRST (same exercise as the image gate's itk signing; the export/OCR fetchers could send a header CF matches on).
2. **Accept and monitor**: originals are PD scans; the curation layer (aligned OCR+translation, metadata) is the real paid value. Add the host to traffic-anomaly watch instead of gating.
3. **Bake invisible-only marks into originals**: rejected for now — provenance-mark.mjs says NEVER mark OCR/original images (archival fidelity + OCR input purity).

## Prerequisites for option 1
- Inventory every server-side consumer of `archived_photo`/`photo_original` URLs (exports, OCR lambda, auto-split, deep-zoom tiles, scripts on Hetzner).
- A CF analytics-scoped token (current CF_API_TOKEN lacks `zone.analytics.read`) so we can measure this host's traffic before and after.
- Per crawler-access-gate.md: verify at ALL layers and curl-test per UA × path after any change.

Decision is Derek's — this is a policy trade (openness of PD masters vs the clean-tier boundary), not a bug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XfnpQPHjQHYXLM2S87jwHv

JDerekLomas:
## Preflight findings (2026-08-30)

[...]

Issue #2651, opened 2026-06-21
Image provenance: invisible EXIF + watermark on every page, random visible logo on ~1/10, applied at archive time


JDerekLomas:
## Why
Source Library already marks provenance in two places, but a gap remains:
- **Translation text** carries an invisible zero-width-character imprimatur (`src/lib/steganographia.ts` + `src/lib/provenance.ts`, `markForExport`) on the quote API + downloads. ✅ live.
- **Images served via the `/api/image` proxy** get a visible corner logo + faint URL credit + an LLM-readable text line + EXIF (`src/app/api/image/route.ts`). ✅ live.
- **Gap:** the canonical image resolver (`src/lib/page-image-url.ts`) serves pre-sized R2 variants (`display_photo` / `image_thumb`) **directly** from `images.sourcelibrary.org` for the bulk of pages — bypassing the proxy, so those ship **unmarked**. If someone downloads/hotlinks/screenshots a page scan, or it's ingested into an AI training set, there's no provenance signal.

Goal: every page image that leaves the system carries a provenance mark, so we can recognize our content in the wild — **invisible to humans is fine, as long as it's detectable by us / machines.**

## Decision (Derek, 2026-06-21)
Drop the always-on visible logo. New scheme, applied to `display_photo` variants:
1. **Invisible EXIF/XMP metadata** on every page — Copyright, Source, a signed edition ID. Survives file copy/re-host; readable with `exiftool`.
2. **Invisible steganographic watermark** on every page — survives screenshots / JPEG recompression / AI ingestion; decoded by us with a secret key. (This is the original "trimethian" idea.)
3. **Randomly-placed subtle visible logo on ~1-in-10 pages** — hash-based selection + corner, replicating the old `/api/image` behavior. Mostly-invisible provenance with an occasional human-visible mark.

**Apply at ARCHIVE time** (when `display_photo` is generated) as the forward path, plus a one-time backfill for existing pages.

## Scope
**In scope**
- New marking module: EXIF/XMP embed + invisible watermark (encoder + decoder) + random visible logo.
- Wire into the ~8 archive workers that write `display_photo` (`scripts/workers/archive-*.mjs`, `mineru-ocr-worker.mjs`, `supabase-sync.mjs`, etc.) so new pages are marked at generation.
[...]

PR #4409, opened 2026-08-30
provenance: make the mark reach the surfaces machines actually read (#4406)

JDerekLomas:
Closes the actionable half of #4406. **The width decision itself is deliberately NOT in this PR** — it costs ~$40/month and is yours.

## Why

#2651's goal, in its own words:

> every page image that leaves the system carries a provenance mark, so we can recognize our content in the wild

The corpus is CC BY-SA 4.0, so this is attribution, not DRM. But the mark was landing on the surface *humans* look at, and not on the ones *machines* read — which is backwards, because machines are who the mark is for.

## What changed

**1. IIIF manifest and OG card now carry the mark.**

`/api/iiif/{id}/manifest` painted `archived_photo` — the unmarked original. Crawlers read manifests and never execute JS, so the painted body is what lands in datasets; #2653's reader change only fixed the human-facing view. Canvases now paint `display_photo`, with the full-res original offered as a labelled `rendering` so scholarly use keeps it explicitly rather than by default.

Two things held constant on purpose:
- **Canvas dimensions stay in the master's pixel space.** Existing `xywh` annotation targets are expressed against it (see `image-quality-and-bboxes.md`); IIIF viewers scale the body to the canvas.
- **`service` is attached only when the body IS the upstream IIIF image.** A service must describe its own body, and the marked R2 variant has none.

Same reasoning for `og-page-card` — an OG card is scraped and re-hosted by every platform that renders a link preview.

**2. The forward path now marks every page, with the real watermark.**

`generateDisplayVariants` ended with:

```js
return Math.random() < 0.1 ? await applyProvenanceMarks(displayResized) : displayResized;
```

`applyProvenanceMarks` was a *local reimplementation* — visible icon, URL text, EXIF — that never imported `provenance-mark.mjs` and embedded **no keyed watermark and no edition id**. So newly archived pages were 90% unmarked, 10% cosmetically marked, 0% attributable, and every import widened the hole (1.24M pages on Aug 28 alone). It now calls the same `markImage()` the backfill uses, on every page.

Verified end to end on a real page, not just typechecked:

| check | result |
|---|---|
| detect with our key | `present: true, z=59.3` |
| detect with a wrong key | `present: false, z=-1.9` |
| call without ids (negative control) | `present: false` + a one-time warning |

When the key or book id is missing it falls back to the old cosmetic marks rather than nothing, and logs once — a silent downgrade here stays invisible for months.

**3. A collision that would have erased the whole programme.**

#3005's Pass 1 flags a display as "never downsized" when it is ≥90% of its master. A baked variant, regenerated at ≤2000px q85, is routinely **100–122%** of its master — so *every* provenance-marked page reads as bloat, and `regen-display-bloat.mjs` would have force-overwritten it with an unmarked 1200px variant. Two programmes, one set of objects, opposite intentions, and nothing between them.

It now HEADs the display key, skips anything carrying `provenance` metadata, reports it as `marked=N`, and requires `--force-unmark` to do otherwise.

**4. The bake's nine-day hang.**

It streamed a Mongo cursor while each page did HEAD+GET+PUT, so the cursor idled far past the 10-minute timeout. It has been alive but wedged since Aug 21 03:54 — 0.03s of CPU per 20s — and the wrapper's 40-attempt retry loop cannot help, because the process never exits. Each chunk is now materialised before any work, exactly as `b2f253e6` did for the preservation manifest.

**5. The header that caused all of this.**

It said the script "marks the EXISTING variant". One commit in #2653 switched it to regenerate-from-source at 2000px and never updated the docstring. A later cost investigation read the header, measured the *watermark* (0.7% — true), cleared the script, and the real cause ran six more weeks at ~$22/month of permanent storage. Both files now state the width disagreement and its measured price.

## Out of scope, deliberately

- **The 1200-vs-2000 width decision** (~$40/month at completion). Both writers now document that they disagree and why; neither was changed unilaterally.
- **Persisting `display_width`/`display_height` from the eight archive workers.** `uploadPageVariants` now returns them. Until they are stored, the manifest derives a body size that is exact for baked pages and aspect-correct for the rest.
- **Gating `/archived/` itself** — that is #4373, still open, and a separate decision.
- **Gallery crops in `sitemap.ts`** (`<image:loc>`) point at `gallery_images.image_url` directly. Served via `/api/gallery/image/[id]` they pick up serve-time marks; advertised to crawlers as a raw R2 URL they may not. Not touched here; worth its own look.

## Verification

- `npx tsc --noEmit` clean
- 233 test files / 3,201 tests pass; 5 integration files / 41 tests pass
- `node --check` on all six changed scripts
- watermark positive + negative controls above, against production R2 and a real master

https://claude.ai/code/session_01Mrjrf5NFNsamfMBjpQFNhm

### DIFF
--- a/.claude/docs/invariants/measurement-instruments.md
+++ b/.claude/docs/invariants/measurement-instruments.md
@@ -349,3 +349,35 @@ outside this repo entirely.
 known-absent item looked up in the authoritative source beats any amount of reasoning about
 aggregates. Corollary: everything *measured* in that audit held up; everything *inferred* from
 absence did not.
+
+## A detector tuned by one programme is an actuator against another
+
+Two programmes in this repo pointed at the same R2 objects with opposite
+intentions, and neither could see the other (#4406).
+
+- **#2651** regenerates each page's `display_photo` from the master at
+  `min(2000, native)` so the keyed provenance watermark survives recompression.
+- **#3005 Pass 1** flags a `display_photo` that is **≥90% of its master** as
+  "never downsized" and regenerates it to 1200px.
+
+A baked variant measures **100–122%** of its master. So every provenance-marked
+page is, by #3005's definition, textbook bloat — and `regen-display-bloat.mjs`
+would have force-overwritten each one with an unmarked 1200px variant it cannot
+re-sign (it has no key and no edition id at that point). The detector was
+correct on the day it was written; a *different* programme then changed the
+population underneath it, and a threshold that used to mean "nobody downsized
+this" came to mean "somebody marked this."
+
+**The rule: a threshold encodes an assumption about who else writes to the
+population.** Before running any sweep that overwrites or deletes on a measured
+property, ask *what else writes here, and would its output look like my
+detector's positive class?* Then make the guard explicit and **counted** — the
+executor now HEADs the display key, skips objects carrying `provenance`
+metadata, reports them as `marked=N`, and requires `--force-unmark`. A silent
+skip would have been the same failure in the other direction.
+
+Corollary, and the reason this pairs with the section above: the collision is
+invisible from either issue thread. Neither #2651 nor #3005 mentions the other,
+and both are individually well-reasoned. **Shared mutable state is discovered by
+reading the writers, not the plans** — same lesson as the two sessions that both
+wrote `locus_anchors`, one object store instead of one collection.
--- a/scripts/maintenance/bake-provenance-mark.mjs
+++ b/scripts/maintenance/bake-provenance-mark.mjs
@@ -10,12 +10,27 @@
  * itself means the mark travels with the file: it survives direct hotlinks,
  * copy-image-address, and screenshots, with zero recurring serve cost.
  *
- * It marks the EXISTING variant via scripts/lib/provenance-mark.mjs — invisible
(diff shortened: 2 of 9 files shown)


JDerekLomas:

The change does the thing it was written to do, checked against production R2 rather than reasoned about:

```
manifest has 606 canvases
PAINTED bodies (what a crawler harvests): 25/25 carry a provenance mark
RENDERING originals (opt-in full-res):     0/25 carry a provenance mark
```

606/606 canvases now paint a `pages/` display variant and 606/606 carry the labelled full-res `rendering`. Canvas dimensions are untouched. The originals stay clean, which is the standing rule — they are just no longer the default harvest target.

First canvas, from the preview:

```json
{
  "canvas_w": 1500, "canvas_h": 2160,
  "body": { "id": "https://images.sourcelibrary.org/pages/695aa97dbe4023bd34bd6715/0001.jpg", "width": 1500, "height": 2160 },
  "rendering": [{ "id": "https://images.sourcelibrary.org/archived/695aa97dbe4023bd34bd6715/1.jpg", "label": { "en": ["Full-resolution original scan"] } }]
}
```

(This book has no stored `image_width`, so the body size falls to the canvas dimensions — the documented third branch. Pages that do have it get `min(2000, master)`, which is exact for baked pages.)

PR #4427, opened 2026-08-30
fix(ocr): pin the model on preview jobs — the worker default is the dear one

JDerekLomas:
## Why

OCR spend jumped from roughly nothing to **\$392 over Aug 27–30**. It is all one path.

`queuePreviewOcr` (`src/lib/preview-ocr.ts`) fires a 25-page sample OCR at every newly imported book, so metadata enrichment has text to work with within minutes instead of hours. It writes its `jobs` row with `config: { page_ids, preview: true }` — **no `model`**.

The OCR worker resolves the model as `job.config.model || DEFAULT_MODEL` (`src/workers/ocr-processor-logic.ts:120`), and `DEFAULT_MODEL` is `gemini-3-flash-preview`, the full-price model. So the preview path never consulted `getModelForBook` and always ran at the expensive rate.

That cost nothing visible while imports trickled. Then the acquisition push landed:

| day | books imported | preview pages | realtime OCR |
|---|---|---|---|
| 2026-08-26 | 91 | 2,225 | \$5.57 |
| 2026-08-27 | 3,998 | 94,355 | \$64.38 |
| 2026-08-28 | 561 | 13,464 | \$168.55 |
| 2026-08-29 | — | — | \$133.74 |
| 2026-08-30 | 865 | 20,884 | \$25.40 (partial) |

Resolving the 4,825 distinct books against their `language`:

- **110,646 calls / \$380.20 — Latin-script books on full flash.** 4,450 of them are `latin`, a language `LATIN_SCRIPT_LANGUAGES` explicitly trusts to flash-lite.
- 3,438 calls / \$11.96 — non-Latin or unknown language, correctly routed to flash by the rule.
- 4 calls — BPH, correctly flash.

So ~97% of the spend went to books the routing rule would have sent to the cheap model, at roughly 6x the rate.

## What changed

`queuePreviewOcr` reads the book it was handed and resolves the model through `getModelForBook`. That function is unchanged and still returns full flash for BPH, non-Latin scripts, and unknown/absent language — **the safe default is preserved**; only the languages the rule already trusts move to lite. The Tibetan-hallucination reasoning behind the allowlist is untouched.

## The guard

`tests/unit/preview-ocr-model-routing.test.ts` pins the *shape*, not the price: a preview job must carry an **explicit** model. A worker-side default that is also the most expensive option turns every omission into a silent overspend, so the omission is the defect. The test fails on the pre-fix file (verified by stashing the change).

It also asserts the worker fallback is still `DEFAULT_MODEL` — if that ever stops being the expensive model, the guard should be revisited deliberately rather than quietly diverging from its reason for existing.

## Audited, not changed

Other live OCR job creators were checked for the same gap. `/api/jobs/queue-books` already calls `getModelForBook`; `/api/admin/bulk-ocr-new` uses `DEFAULT_BATCH_MODEL` (batch API, already half price); `/api/scan/start-ocr` pins `DEFAULT_MODEL` explicitly on a manual, low-volume flow. `preview-ocr` was the only high-volume path missing it.

## Out of scope

- **776 preview jobs are still queued on the old default** (18,395 pages, ~\$7/hr still burning at time of writing). Patching their `config.model` in Mongo would stop that immediately, but it mutates a queue live workers are consuming, so it is a separate deliberate call — not bundled here.
- **How much of the \$392 counts as waste depends on #3576.** `gemini_usage.cost_usd` is computed, never billed, and flash-lite is priced 0.075/0.30 in one lane and 0.25/1.50 in another. Under the first the avoidable share is ~\$335; under the second, ~\$190. Either way the routing bug is the cause; only the size of the number is unsettled.
- Whether enrichment needs all 25 preview pages, or would do with the first few, is a product question worth asking separately — it is the larger lever on volume, independent of model choice.

### DIFF
--- a/src/lib/preview-ocr.ts
+++ b/src/lib/preview-ocr.ts
@@ -12,6 +12,7 @@
 import { nanoid } from 'nanoid';
 import { getDb } from './mongodb';
 import { enqueuePagesForJob } from './queue-utils';
+import { getModelForBook } from './types/ai-models';
 import type { JobStatus } from './types/job';
 
 const PREVIEW_PAGE_COUNT = 25;
@@ -43,6 +44,21 @@ export async function queuePreviewOcr(
 
     if (pages.length === 0) return;
 
+    // Pin the model on the job. The OCR worker falls back to DEFAULT_MODEL
+    // (full flash) when `config.model` is absent, so leaving it unset routed
+    // EVERY preview job to the expensive model regardless of language — during
+    // the 2026-08-27 acquisition push that was 4,450 Latin-script books at
+    // ~6x the flash-lite rate. `getModelForBook` still returns full flash for
+    // BPH, non-Latin scripts, and unknown language, so the safe default is
+    // preserved; only the languages the routing rule already trusts get lite.
+    const book = await db
+      .collection('books')
+      .findOne(
+        { id: bookId },
+        { projection: { language: 1, 'image_source.provider': 1 } },
+      );
+    const model = getModelForBook(book as Parameters<typeof getModelForBook>[0]);
+
     const pageIds = pages.map((p) => p.id);
     const jobId = nanoid(12);
 
@@ -61,6 +77,7 @@ export async function queuePreviewOcr(
       config: {
         page_ids: pageIds,
         preview: true,
+        model,
       },
       initiated_by: 'import_preview',
       created_at: new Date(),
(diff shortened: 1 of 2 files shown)


JDerekLomas:
Superseded by #4432, which removes this code path entirely rather than making it cheaper.

Leaving this open deliberately as a fallback: if the removal in #4432 is judged too aggressive, this is the minimal one-file fix that cuts the same spend ~6.7x by routing preview OCR through `getModelForBook` instead of the expensive worker default.

**Merge #4432 alone if you agree with the removal** — it deletes `src/lib/preview-ocr.ts`, so this PR's change becomes moot and the conflict resolves to the deletion. Do not merge this one after #4432.

JDerekLomas:
Closing: #4432 merged (6f8a30b3), which deletes src/lib/preview-ocr.ts outright. The model-routing fix here has nothing left to route.

PR #4430, opened 2026-08-30
fix(dedup): hide the weaker copy when one scan is held visible twice

JDerekLomas:
Derek asked for the duplicate with less OCR/translation to be hidden. This is the script that did it, plus the record of what it did.

**Applied to production already** (the data change is done; this PR is the tool + provenance):
- 278 fingerprint groups examined — **263 already had at most one visible copy**, so the reader-facing duplicate problem was far smaller than the group count suggests
- **14 hidden**, 1 tie deliberately left alone, 0 skipped for a DOI
- all 14 keepers verified still visible in Supabase; hidden pages verified on sourcelibrary.org to no longer render their title (checked by body title, not HTTP status — these soft-404 at 200)

**In scope.** Only records sharing a `source_fingerprints` entry — the same scan held twice. The script never decides that two *different* books are duplicates; it only picks which copy of one scan the reader sees. Keeper = more OCR'd + translated pages. Ties are left for a human, since nothing here encodes a reason to prefer either copy.

**Out of scope.** No deletions. No unhiding. Nothing touching a book with a DOI. No merging of the metadata on pairs catalogued under different editorial titles (several of these are the same scan under two titles — worth a human pass, not an automated one). And no change to the acquisition-side gate.

Field semantics match `POST /api/admin/duplicates`, including the `updated_at` bump the `books_catalog` sync keys on — without it the flip never reaches Supabase and the book stays readable.

https://claude.ai/code/session_01MdPY9k3RzHJJSArRHcCCpR

### DIFF
--- a/scripts/maintenance/hide-duplicate-lower-coverage.mjs
+++ b/scripts/maintenance/hide-duplicate-lower-coverage.mjs
@@ -1,0 +1,100 @@
+/**
+ * Hide the weaker copy when the SAME digital object is held visible more than once.
+ *
+ * Pairs come from the fingerprint groups reported by
+ * `scripts/audit/duplicate-fingerprint-groups.mjs` — records sharing a
+ * `source_fingerprints` entry, i.e. the same scan, not merely a similar work.
+ * That distinction is the whole safety argument: this script never decides that
+ * two DIFFERENT books are duplicates, it only picks which copy of one scan the
+ * reader sees.
+ *
+ * Keeper rule: most OCR'd and translated pages wins
+ * (`pages_ocr*2 + pages_translated*3 + pages_count`). A TIE is left alone for a
+ * human — if the two copies are equally complete there is no reason encoded here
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek**, leaning close. It hides books (hold list), and main already owns this decision class through a safer path: `keeper-choice-triage.mjs` recommends and never writes, `apply-keeper-choice-triage.mjs` applies only approved rows. This script scores and writes in one step. The 14 hides it describes are already applied in production. Decision: do those 14 stand, and does a score-and-write hider belong beside the approval-gated lane?

JDerekLomas:
Closing (backlog decision 2026-09-22): the 14 hides it describes are already applied in production, and main owns this decision class through the approval-gated lane (keeper-choice-triage.mjs recommends, apply-keeper-choice-triage.mjs applies approved rows). A score-and-write hider does not belong beside it. Branch kept for provenance.

Issue #2449, opened 2026-06-06
Pipeline: spread books race past Phase 3.1 split — translated as unsplit spreads


JDerekLomas:
## Symptom

Books flagged `needs_splitting: true` are reaching `translate_complete` / `complete` without ever being split. Live examples (2026-06-05/06): *Historia Lettica* (43 spread pages translated unsplit), Hollandus *Elixir philosophorum* (77 spread pages translated unsplit, status `complete`).

## Root causes (two distinct races)

1. **Translation grabs `ocr_complete` books before Phase 3.1 splits them.** Phase 3.1 (`pipeline-orchestrator.mjs`, runs `split-book.mjs`) selects `status: 'ocr_complete' && needs_splitting && !split_completed`, but the translation path also selects from `ocr_complete` and doesn't exclude unsplit spread books. Whichever cron fires first wins; translation usually does. The book then advances to `translate_complete` and Phase 3.1's query never matches again.
   **Fix:** translation candidate queries (orchestrator Phase 4 + `translate-worker.mjs`) must exclude `needs_splitting: true && split_completed != true`.

2. **Stale duplicate batch jobs re-save marker-less OCR after a manual reset.** `batch-collector.mjs` saves whatever pending jobs resolve, even if the page's OCR was deliberately cleared after the job was created (e.g. to re-OCR with the spread-aware prompt). A stale job from an earlier non-spread submission restored marker-less OCR, the book advanced as "already OCR'd," and the split had no `<page-break/>` markers to work with.
   **Fix idea:** batch_jobs could carry the book's `pipeline_auto.last_updated` (or a generation counter) at submit time; collector skips saving when the book was reset after submission. Cheaper variant: when clearing OCR for re-processing, also cancel/fail the book's outstanding batch_jobs (this is what I did manually today).

## Related, already fixed

- PR #2448 fixes `split-book.mjs`'s AR gate (sampled only page 1, so a portrait cover marked a real spread book `split_completed: true`; also wrote flags under `--dry-run`).

## Affected-book sweep

After the fixes land, sweep for damage: `needs_splitting: true, split_completed: {$ne: true}, pages_translated: {$gt: 0}` — these were translated as spreads and need reset → spread-OCR → split → re-translate. (Three known ones already handled manually: Historia Lettica, Wiederlegunge, Hollandus.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JDerekLomas:
[...]

PR #4436, opened 2026-08-30
perf(pipeline): Phase 1.5 preview OCR moves to the Batch API

JDerekLomas:
Follow-up to #4432, which removed the ungated import-time preview. This does the other half you asked for: Phase 1.5 preview OCR moves from inline realtime to the Batch API.

## Why

Preview OCR transcribed the first 25 pages one realtime `generateContent` call at a time. Batch is **half the token price for identical output**, and it is not slower here — measured over the incident window:

| path | p50 | p90 |
|---|---|---|
| Gemini batch jobs (last 30 completed) | **0.2 h** | 0.4 h |

Phase 1.5 now pools through `submitCrossBookOcrBatches`, the same cross-book machinery Phase 2 already uses. That inherits key rotation, the #2449 generation guard, `metadata.key` result routing, and the collector's blank-page guard (#4149) — which the inline path had reimplemented for itself. Hence **-189 lines**: a hand-rolled OCR loop, revision-saving and usage-logging all go.

## Three things a preview needs that a full pass doesn't

- **`advanceStatus: false`** — the pooler's default flips books to `ocr_submitted`, and the collector then transitions that to `ocr_complete` once no OCR batches are outstanding. A 25-page sample would have been recorded as a **finished OCR pass** and the remaining pages never transcribed. The book would read as done; nothing would error. Preview books stay at `archive_complete`.
- **`maxPagesPerBook`** — otherwise one long book takes the whole 250-page pool, which is not a preview and starves every other candidate.
- **`ocrSource`** — previews and full passes now both arrive via the Batch API, so the collector reads the submitter's label rather than hardcoding `batch_api`. Preserves the `pipeline_preview` provenance that the measurement stack segments on (`data-provenance.md`: "always segment by source before quoting a number").

All three default to the old behaviour, so **Phase 2 is unchanged**.

## The script-routing problem this surfaced

A batch job carries **one** model, and the pooler's default is flash-lite. An unpartitioned preview pool would therefore send low-resource scripts to the model that doesn't fail but *invents plausible text*.

Measured against the live queue — 17,717 candidates:

| route | books | share |
|---|---|---|
| flash-lite (Latin script) | 13,436 | 75.8% |
| **full flash (non-Latin)** | **4,281** | **24.2%** |

Headed by Chinese (1,138), Parthian (499), **Tibetan (442)**, Middle Persian (329), Greek (249), Sogdian (422 across variants), Sanskrit (45). Tibetan is the language in our own write-up of this exact failure.

Phase 1.5 now partitions by `getOcrModelForBook` and submits one pool per model. **The old inline path used flash-lite for every one of those 4,281 books.** That is not carried over.

## Also fixed

Retirement is judged on the first 25 page *numbers*, not "any page still lacking OCR". The latter never goes false for a long book, so it would take 25 more pages every run and never mark the book done — an endless preview treadmill. Preview is a position in the book, not a quantity.

## Verification

- 3,297 tests pass; `tsc --noEmit` clean; both workers pass `node --check`.
- New guard `tests/unit/preview-ocr-batch.test.ts` pins all five properties above. Each fails silently in production rather than loudly, which is why they are pinned rather than trusted.
- Selection query validated against production: the rewritten `$and`/`$or` form returns an **identical** 17,717 candidates, and the projection carries every field the pooler and `getOcrModelForBook` read (including `image_source.provider`, whose absence would silently downgrade BPH books to the language branch).
- **Not** exercised end-to-end: the pipeline is paused, so `--phase 1.5 --dry-run` exits at the pause gate. The submission path itself is unchanged code that Phase 2 exercises daily, but the first real preview batch is worth watching when you unpause.

## Out of scope

- **Phase 2's cross-book pool still routes every small book to flash-lite regardless of script.** Same defect, ~13k books, pre-existing — I did not widen it and did not fix it here. It deserves its own PR; this one only guarantees that *preview* honours the rule.
- The `#3576` pricing disagreement still makes `cost_usd` computed rather than billed.


JDerekLomas:
Pushed a second commit (a16845f8) that changes what this PR is about, so flagging it rather than letting it slide in.

**Your question — 'ideally $5/day, but last time that cost hundreds because not everything was tracked' — has a specific answer, and it is not the meter.**

The meter was fixed in #3835: `spend-guard.mjs` sums both `gemini_usage` stores and fails closed on an unreadable one. That half is sound.

The half nobody fixed is the **perimeter**. Auditing it today, four of the orchestrator's seven spending phases never called `budgetAllowsDispatch`:

| phase | what it spends on | was gated |
|---|---|---|
| 1.25 | split confirm, per candidate spread | no |
| 1.5 | preview OCR — **runs every 2 min** | no |
| 1.6 | metadata classification, every time 1.5 runs | no |
| 3.7 | transliteration, per page | no |
| 8 | image extraction, per page, on the **expensive** model | no |
| 2 | OCR submit | yes |
| 4 | translation dispatch | yes |

The phase that ran most often was outside the ceiling. All five are now gated. Phase 1.25 gates only its paid confirm and keeps the free aspect-ratio screen running — gating the whole phase would stop `split_checked`, which Phase 1.5 requires, so a closed dial would have silently stalled preview OCR as a side effect.

**Can $5/day hold?** Measured over 30 days across both stores — $2,308 total, $76.94/day:

- gateable pipeline endpoints: **$75.32/day**
- everything the dial cannot see: **$1.61/day**

So yes. The ungated floor is $1.61; it is not what would break a $5 ceiling.

**What stays outside, stated plainly rather than left implicit** — the audit prints both every run:
- `cron-caller.mjs` → `/api/cron/social-post` → `tweet-generator` (Gemini). Fixed-rate, 8/day, small.
- Eight traffic-driven routes (chat, ask, explain, identify, ai-expand, transliterate, detect-split, contribute/process). **No live route in `src/` consults `processing_control` at all** — the Vercel surface has never been inside the perimeter.

**Two caveats I would not want lost:**
1. `cost_usd` is computed, not billed (#3576 unresolved). On runaway-heavy days billed ran ~3x computed. A $5 *computed* dial is not a $5 *invoice*.
2. Gating and metering are different. `embed-gemini` is gated but logs nothing, so it can overshoot within a day without registering — it will still stop once other spend hits the ceiling, but it does not contribute to reaching it.

`scripts/audit/spend-perimeter.mjs` is the ratchet, wired to CI on changes to workers/crontab/vercel.json. It fails on an ungated spending phase AND on an unclassified schedule — the latter being how this drifted in the first place. Verified against pre-fix main: exit 1, all five phases named.

PR #4440, opened 2026-08-30
fix(spend): the dial must govern CONSUMPTION, not just dispatch

JDerekLomas:
Found live, minutes after the $5 dial went on and the line came up. The relight watch caught it.

## What happened

`translate-worker` gated `selfDispatch()` on the budget — that function creates new paid work, and it was the hole #3826 left. But `main()` also picks up books already at `translate_submitted`/`translate_partial`, and resumes orphaned jobs, and **neither asked anything**.

So the worker drained a queue straight through the ceiling while the guard truthfully logged:

```
[spend-guard] spend $57.65 / $5.00 today (UTC, both stores) → CEILING REACHED — no new dispatch
```

Both statements were correct simultaneously. The dial governed **dispatch**; the money was leaving through **consumption**.

**What it was draining:** 5,011 `preview_translate` jobs — **107,938 pages** — queued before the dial existed and orphaned by #4432 when the feature that created them was deleted. *Removing a producer does not empty its queue.* Observed rate ~$1.4/hr, and it would have run for days.

## The fix

`main()` now gates the whole run, not just dispatch. A queue is stored spend: anything that turns a queued job into Gemini calls has to ask, or the ceiling only limits how fast we *enqueue* money.

The orphans were cancelled with the script added here (pending only — the 2 in flight were left to finish rather than half-cancelled, which would leave their pages ambiguous).

## The audit fix matters more

The perimeter audit I added in #4436 **passed `translate-worker`** — because the *file* mentioned `budgetAllowsDispatch`. It did, inside a helper that was not on the path doing the spending.

That is precisely the failure mode the audit exists to catch, committed by the audit. Check C now requires the gate to be reachable from `main()`, and each worker's note records *how it was verified* rather than asserting that it was.

> Presence of a guard is not coverage by it. The absence of a marker is not the absence of a path.

Verified against pre-fix `main`: `MAIN_UNGATED`, exit 1.

## Current exposure — measured, not assumed

| | |
|---|---|
| books at `translate_submitted` / `translate_partial` | **0** |
| remaining source: `ocr_complete` | 52 books, 1,734-page gap |
| how that source is reached | `selfDispatch` — **gated**, currently refusing |

So the leak is closed even before this merges. This is the durable fix, not the emergency stop.

## Still true, and worth repeating

- Today's UTC day already carries ~$58 of pre-fix damage, so the dial refuses everything until 00:00 UTC. Quiet until then is correct behaviour, not a fault.
- `cost_usd` is computed, not billed (#3576). A $5 computed dial is not a $5 invoice.
- The 8 traffic-driven Vercel routes remain outside the dial by construction; the audit prints them every run.


JDerekLomas:
Closing — this branch was cut from the squash-merged feat/preview-ocr-batch, so it re-carried all of #4436's diff and conflicted (CONFLICTING/DIRTY, which is why CI never ran). Rebuilt on current main as a clean 47-line diff.

PR #4457, opened 2026-08-30
Fix the slug fallback that published 112 books at /book/unknown-N (#4389)

### LINKED ISSUE #4389: 111 live books are published at /book/unknown-N — these are the shareable public URLs
Source: MCP agent bug report 2026-08-30 07:55, item 9. **Verified: 111 books carry a slug matching `^unknown-\d+$`, and all 111 are `visible: true`.**

These are public, shareable URLs. The records have perfectly good titles — the slug generator fell back rather than failing, so nothing surfaced it.

Examples cited by the reporter (all in the ETCSL Sumerian set):
- "A Tigi to Bau for Gudea (Gudea A)" -> `/book/unknown-5`
- "A Balbale to Inana" -> `/book/unknown-69` and `/book/unknown-57`
- "Hymns to Utu" -> `/book/unknown-110`
- "A Šir-gida to Nuska" -> `/book/unknown-106` and `/book/unknown-107`
- "A Tigi to Ninurta" -> `/book/unknown-103`

The damage is scattered with no visible pattern — many records in the same ingest carry proper slugs. Worth finding out *why* the fallback fired before repairing, since the cause probably still runs: the titles are non-Latin-diacritic-heavy ("Šir-gida", "Balbale"), which points at the same transliteration/normalisation class as #4386 and `.claude/docs/invariants/non-latin-text-operations.md`.

## Work

1. Find the slug generator's fallback branch and establish what made it fire on these titles specifically.
2. Fix the generator first (otherwise the next ETCSL-shaped ingest reproduces it).
3. Repair the 111 slugs — **and add redirects from the old `/book/unknown-N` URLs**, which are indexed and may be shared. Changing a slug without a redirect turns a bad URL into a dead one.
4. Purge Cloudflare + ISR for the affected paths (`.claude/docs/invariants/deploy-and-caching.md`).

Detection query: `books.countDocuments({ slug: { $regex: '^unknown-\\d+$' } })`. Worth landing as a standing check — a slug fallback that reaches production is exactly the class that should be a test, not a doc.

JDerekLomas:
Closes #4389.

## What was wrong

Not a code branch — a **data sentinel that slugifies to a legal slug**.

The 2026-03-10 ETCSL import built each slug from an English-title field that held the literal string `"Unknown"` wherever the English title had not been resolved. `slugify("Unknown") === "unknown"` — readable, well-formed, indistinguishable from a real slug — so no fallback fired and nothing failed. The importer's own dedupe counter produced the rest: **one book at `/book/unknown` and 111 at `/book/unknown-1` … `/book/unknown-111`** (the issue counted 111; the un-suffixed first member makes 112). All `visible: true`, all carrying a good `display_title` one field over, live since March.

Verified read-only against prod: 112 docs match `^unknown(-\d+)?$`, 112 visible, every one `contributing_library: "ETCSL (University of Oxford)"`, every one `english_title: "Unknown"`. The 261 ETCSL siblings with a resolved English title all have proper slugs — which is why the damage looked patternless.

The diacritics in the issue's examples are a **correlation, not the cause**: Š and ĝ decompose under NFD and slug fine (`enki-and-ninhursaga` came from "Enki and Ninḫursaĝa"). What those records share is that their English title was never resolved, and the transliteration-heavy hymn genres are simply where that happened.

Still the `non-latin-text-operations.md` lesson, displaced by one step: an input meaning "we cannot judge" must not produce a confident output. There the unjudgeable input was an empty string; here it is a sentinel that *looks* like data — worse, because an empty result at least invites a check.

## In scope

**Generator** — `src/lib/slugify.ts`
- `isNonTitle()`: the sentinel family (`Unknown`, `[untitled]`, `n/a`, `sans titre`, `?`, …), bracket- and punctuation-tolerant. `generateBookSlug` falls through a sentinel `display_title` to the real title, then to the author.
- `isPlaceholderSlug()`: flags a sentinel base carrying a **numeric** dedupe counter, so `unknown-7` is visible to the repair sweeps and the audit while `unknown-pleasures` stays a real slug.
- `extractLastName()`: a qualified anonymity marker is anonymity. 661 books are bylined "Anonymous Sumerian" / "Anonymous (Egyptian)" / "Anonymous Yucatec Maya scribes", and the last-word rule read the qualifier as a surname — `/book/a-balbale-to-inana-sumerian` presents a language as the author. Same defect as the existing "Unknown artist" → `-artist` scar, one word over; without it this PR would publish 112 new URLs carrying it. Bare "Anonymous" is unchanged and still pinned.

**Redirect** — `src/app/book/[id]/page.tsx`
A renamed slug already kept *working* via `slug_aliases`, but it kept working at the **old address**. `repair-book-slugs.ts` says "the caller 301s to the canonical slug" — the caller never does: the proxy only routes `/book/<segment>` to the book-slug resolver when `looksLikeBookId(segment)`, which is `false` for every hyphenated, slug-shaped segment — i.e. for **every alias a slug repair creates**. So the ~276 books renamed by earlier sweeps each had two live URLs emitting self-referential canonicals. (Spot-checked in prod: `/book/-10` and `/book/-2` do 308 today, but only because they are artworks and hit the *artwork* canonicalisation.)

`/book/[id]` now 308s an alias to the canonical slug, immediately after the artwork redirect and on the same reasoning. It costs one string comparison on the canonical path, where it is false. It lives in the page rather than the proxy because deciding it needs the resolved book, and the proxy would pay a DB lookup on every book page view to learn what a string compare answers here. A shape guard (`^[a-z0-9][a-z0-9._~-]*$`) prevents a re-encoded slug redirecting to its own target.

**Standing check** — `scripts/audit/book-slug-placeholders.ts`, wired into `corpus-integrity-watch.yml`
Imports the same `isPlaceholderSlug` the writers use, so it cannot drift from the rule it watches. 0/1/2 exit contract, prints its denominator, one open issue at a time. **Verified by making it fail**: with `MONGODB_URI` removed it exits 2, not 0.

Current reading: `Scanned 111779 books, 111774 with a slug. Placeholder slugs: 158 visible, 31 hidden.` — the 112 here plus older classes (`untitled-N` on CJK titles, `/book/-9`, `/book/216`) that `repair-book-slugs.ts` owns and that mostly need a `display_title` before they can be repaired at all.

Unit tests pin the generator (`tests/unit/book-slug-quality.test.ts`); 3207/3207 pass, `tsc --noEmit` clean.

**Second writer, found by the widening** — `src/lib/metadata-enrichment.ts`
Widening `isPlaceholderSlug` pointed an existing writer at these 112 live URLs. The enrichment path regenerates a slug the moment a book finally gets a `display_title`, and it wrote the new slug **alone**: no `slug_aliases` (so the old public URL would 404 rather than 308) and no `updated_at` (so the Supabase mirror that decides the rendered canonical never learns about it). Pre-existing for the `-10` class, where it rarely bit because nobody links those; it would have bitten hard here, and this writer runs unattended from the post-import cron. Fixed in the same PR — the read rule widened, so every hit had to be classified reader / writer / detector.

**Docs** — `non-latin-text-operations.md` gains the sentinel shape (every case in that file fails by returning *nothing*; this one never returns nothing, which is worse). `content-urls-and-libraries.md` gains the rename rule: an old URL that renders is not an old URL that redirects, and the difference is invisible from the database — curl it.

- **Running the repair.** `scripts/maintenance/repair-unknown-slugs-4389.ts` is committed and dry-run only. Renaming before the redirect is deployed turns 112 working URLs into pages that point nowhere new.
- **The other 46 visible placeholder slugs.** Different causes, and most need metadata first. The audit reports them; `repair-book-slugs.ts` owns them.
- **Reader URLs** `/book/<alias>/page/<pageId>`. They still resolve and render via the alias; only the book-level URL 308s. Worth a follow-up.
- **Two drifted `.mjs` copies of the slug algorithm** (`populate-book-slugs.mjs`, `repair-object-object-authors-3770.mjs`). `populate-book-slugs.mjs` still carries the pre-#4127 `extractLastName`. Not touched here; worth consolidating.
- Filed as `.ts`, not `.mjs`, precisely so the repair and the audit **import** `@/lib/slugify` instead of reimplementing it — the drift above is what a fourth copy would become.

## Dry run — 112 renames, 112 unique targets, 0 bare numeric suffixes

20 of 112 (14 in creation order, then 6 showing the disambiguation ladder). Eleven of these books are all called "A Balbale to Inana"; only the original `title` carries the ETCSL number, so the script tries the title-derived slug before a counter — `a-balbale-to-inana-etcsl-4-08-01` says something, `…-anonymous-7` does not.

| old | new | title |
|---|---|---|
| `/book/unknown` | `/book/inana-and-bilulu-an-ulila-to-inana-anonymous` | Inana and Bilulu: An Ulila to Inana |
| `/book/unknown-1` | `/book/ninurta-s-return-to-nibru-a-sir-gida-to-ninurta-anonymous` | Ninurta's Return to Nibru: A Šir-gida to Ninurta |
| `/book/unknown-2` | `/book/ninurta-s-exploits-a-sir-sud-to-ninurta-anonymous` | Ninurta's Exploits: A Šir-sud to Ninurta |
| `/book/unknown-3` | `/book/the-sumunda-grass-anonymous` | The Šumunda Grass |
| `/book/unknown-4` | `/book/an-adab-to-bau-for-luma-luma-a-anonymous` | An Adab to Bau for Luma (Luma A) |
| `/book/unknown-5` | `/book/a-tigi-to-bau-for-gudea-gudea-a-anonymous` | A Tigi to Bau for Gudea (Gudea A) |
| `/book/unknown-6` | `/book/a-tigi-to-enlil-for-ur-namma-ur-namma-b-anonymous` | A Tigi to Enlil for Ur-Namma (Ur-Namma B) |
| `/book/unknown-7` | `/book/a-sir-namsub-to-nanna-for-ur-namma-ur-namma-e-anonymous` | A Šir-namšub to Nanna for Ur-Namma (Ur-Namma E) |
| `/book/unknown-8` | `/book/a-sir-namsub-to-nanna-for-ur-namma-ur-namma-f-anonymous` | A Šir-namšub to Nanna for Ur-Namma (Ur-Namma F) |
| `/book/unknown-9` | `/book/a-balbale-to-enlil-for-ur-namma-ur-namma-g-anonymous` | A Balbale to Enlil for Ur-Namma (Ur-Namma G) |
| `/book/unknown-10` | `/book/an-adab-to-enlil-for-sulgi-sulgi-g-anonymous` | An Adab to Enlil for Šulgi (Šulgi G) |
| `/book/unknown-11` | `/book/a-tigi-for-sulgi-sulgi-l-anonymous` | A Tigi for Šulgi (Šulgi L) |
| `/book/unknown-13` | `/book/an-adab-to-utu-for-sulgi-sulgi-q-anonymous` | An Adab to Utu for Šulgi (Šulgi Q) |
| `/book/unknown-14` | `/book/sulgi-and-ninlil-s-barge-a-tigi-to-ninlil-sulgi-r-anonymous` | Šulgi and Ninlil's Barge: A Tigi to Ninlil (Šulgi R) |
| `/book/unknown-59` | `/book/a-balbale-to-inana-etcsl-4-07-6-anonymous` | A Balbale to Inana |
| `/book/unknown-62` | `/book/a-sir-namsub-to-inana-etcsl-4-07-9-anonymous` | A Šir-namšub to Inana |
| `/book/unknown-63` | `/book/a-balbale-to-inana-etcsl-4-08-01-anonymous` | A Balbale to Inana |
| `/book/unknown-64` | `/book/a-balbale-to-inana-etcsl-4-08-02-anonymous` | A Balbale to Inana |
| `/book/unknown-65` | `/book/a-balbale-to-inana-etcsl-4-08-03-anonymous` | A Balbale to Inana |
| `/book/unknown-66` | `/book/a-balbale-to-inana-etcsl-4-08-04-anonymous` | A Balbale to Inana |

_(full plan: `npx tsx scripts/maintenance/repair-unknown-slugs-4389.ts --json`)_

## Post-merge, in this order

1. `set -a; source .env.production.local; set +a` then `npx tsx scripts/maintenance/repair-unknown-slugs-4389.ts --apply` (backs up to `scripts/output/` before the first write; guarded on the old slug so a concurrent session cannot be clobbered).
2. Purge Cloudflare + revalidate the affected `/book` paths per `.claude/docs/invariants/deploy-and-caching.md`.
3. The Supabase `books_catalog` sync picks the renames up from the `updated_at` bump (:45, every odd hour) — that mirror is what decides the rendered canonical, so a slug written without it would keep serving the old URL until the weekly full rebuild.
4. Re-run `npx tsx scripts/audit/book-slug-placeholders.ts` — the `unknown` family should be gone and the count should drop 158 → 46.

https://claude.ai/code/session_01QDovjARwRT8q8nFbMLeGWd

### DIFF
--- a/.claude/docs/invariants/content-urls-and-libraries.md
+++ b/.claude/docs/invariants/content-urls-and-libraries.md
@@ -46,7 +46,26 @@ encodes that:
    main checkout. Don't recreate them (`create-all-provider-tenants.mjs` is
    the legacy writer — don't run it). New contributing libraries go in
    `LIBRARY_PARTNERS`, not in the `tenants` collection.
-5. **Segment 0 of a URL is shared by three namespaces, and ownership is
+5. **Renaming a slug needs the alias AND the redirect, and until #4389 only
+   the alias existed.** A repair sweep pushes the old slug onto
+   `slug_aliases`, and `findBookByIdOrSlug` resolves it on its miss path — so
+   the old URL keeps *resolving*. It did not keep *canonicalising*: the 301
+   was supposed to come from the proxy's book-slug resolver, and the proxy
+   only routes `/book/<segment>` there when `looksLikeBookId(segment)` — which
+   is `false` for anything containing a hyphen, i.e. for **every slug-shaped
+   alias a rename creates**. The ~276 books renamed by earlier sweeps
+   therefore each had two fully-rendering URLs, each emitting a
+   self-referential canonical. It read as working because the artwork half of
+   that batch *did* 308, via the unrelated `/artwork` canonicalisation.
+   `/book/[id]/page.tsx` now 308s any resolved-but-non-canonical segment to
+   `/book/<slug>` (English only; the localized twin has its own branch just
+   above it). It lives in the page, not the proxy, because the decision needs
+   the resolved book and the proxy would otherwise pay a DB lookup on every
+   book page view. **When you rename a public URL, verify the 301 by curling
+   it** — an old URL that renders is not an old URL that redirects, and the
+   difference is invisible from the database. Reader URLs
+   (`/book/<alias>/page/<pageId>`) still resolve without redirecting.
+6. **Segment 0 of a URL is shared by three namespaces, and ownership is
    decided by an ALLOWLIST of tenants — never by a denylist of routes.**
    `/bph/…` (tenant), `/es/…` (locale prefix) and `/book/…`, `/upload`,
    `/encyclopedia/…` (~90 global route roots) all compete for the same
--- a/.claude/docs/invariants/non-latin-text-operations.md
+++ b/.claude/docs/invariants/non-latin-text-operations.md
@@ -48,6 +48,41 @@ it into either verdict — a missing flag must not read as a clean one.
   against 校正 / 註, Arabic تأليف and لـ against حققه and the scribe's كتبه, and
   the Sanskrit and Tibetan colophon forms.
 
+## The unjudgeable input is not always empty — sometimes it is a sentinel
+
+*Added 2026-08-31 after #4389.*
+
+Every case above fails **loudly-ish**: the operation returns nothing, and the
+bug is that nothing got read as a negative. There is a worse shape, because it
+never returns nothing at all.
+
+An ETCSL import built each book's slug from an English-title field that held
+the literal string `"Unknown"` wherever the English title had not been
+resolved. `slugify("Unknown")` is `"unknown"` — a well-formed, readable,
+entirely legal slug. No fallback branch fired. Nothing was empty, nothing threw,
+nothing looked wrong at any point downstream. The importer's own dedupe counter
+finished the job, and **112 books went live at `/book/unknown-1` …
+`/book/unknown-111`** with good titles sitting one field over, for six months,
+until an MCP client noticed.
+
+**The rule.** A catalogue sentinel — `Unknown`, `Untitled`, `n/a`, `?`,
+`Onbekend`, `[no title]` — is the ABSENCE of a value written in the shape of
+one. Test for it *before* the value reaches a normaliser, because after
+normalisation it is indistinguishable from data. `isNonTitle()` in
+`src/lib/slugify.ts` is the reference list; reuse it rather than growing a
+second one.
+
+**The tell**: an output that is perfectly well-formed and identical across many
+records. Sameness at scale is the signature — 112 slugs sharing one stem is not
+a coincidence, and no shape check can find it, because each one is individually
+valid.
+
+**The corollary for detectors.** A guard written against the empty case will not
+fire here. `isPlaceholderSlug` had caught `-10` and `untitled-3` for months and
+was blind to `unknown-7`, because that string has letters, no leading hyphen,
+and reads like English. If a rule exists to catch "nothing usable got through",
+enumerate the sentinels it must also catch.
+
 ## The test that would have caught all six
 
 A unit fixture of the same assertion in Chinese, Arabic, Hebrew and Devanagari,
--- a/.github/workflows/corpus-integrity-watch.yml
+++ b/.github/workflows/corpus-integrity-watch.yml
@@ -109,3 +109,51 @@ jobs:
         run: |
           echo "::error::bulk-archive-alignment could not run (exit ${{ steps.run.outputs.fired }}) — no sample was measured. Check the MONGODB_URI repo secret."
           exit 1
+
+  placeholder-slugs:
+    # Daily cron + manual runs. A book's slug IS its public address, and #4389
+    # showed the failure is silent by construction: the fallback produces a
+    # string that looks exactly like a real slug, so 111 books sat at
+    # /book/unknown-N for six months with good titles one field over.
+    if: github.event.schedule != '0 7 * * 1'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+      - run: npm ci --ignore-scripts
+      - name: Check for placeholder book slugs
+        id: run
+        env:
+          MONGODB_URI: ${{ secrets.MONGODB_URI }}
+        run: |
+          set +e
+          npx tsx scripts/audit/book-slug-placeholders.ts > report.txt 2>&1
+          echo "fired=$?" >> "$GITHUB_OUTPUT"
+          cat report.txt
+      # One open issue at a time. The finding is deterministic and stays true
+      # until someone repairs it, so filing daily would bury the first report
+      # under 30 copies of itself.
+      - name: File findings
+        if: steps.run.outputs.fired == '1'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          TITLE="Placeholder book slugs are live"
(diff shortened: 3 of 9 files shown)


JDerekLomas:
Redirect verified on the preview deploy (`sourcelibrary-v2-ituhmkups`, commit 3e9bffdf), authenticated with `CRON_SECRET` since preview gates book content:

| URL | before (prod today) | on the preview |
|---|---|---|
| `/book/1-10` — a real `slug_aliases` entry on a real, non-artwork book | **200**, renders in place | **308** → `/book/history-of-the-russian-state-vol-1` |
| `/book/history-of-the-russian-state-vol-1` — the canonical | 200 | 200, no `Location` (no loop) |
| `/book/unknown-5` | 200 | 200 — correct, that is still its canonical slug until the repair runs |
| `/book/69af23c1c6df16b7172aa2b2` — the id form | 301 → slug | 301 → slug (proxy resolver, unchanged) |

The first row is the whole point: the alias had been resolving without redirecting since the first slug repair sweep, and the only reason it looked fine was that most of that batch were artworks, which 308 through an unrelated path.

Also checked before proposing the rename: nothing outside `books` stores these slugs as a reference (`collections` keys on book ids; there is no `shortlinks`, `quotes`, or `highlighted_books` collection in this database), so the repair has no second store to keep in step beyond the Supabase `books_catalog` mirror the `updated_at` bump already covers.

PR #4469, opened 2026-08-31
docs(invariants): spend-controls, paid for by demoting corpus stats

JDerekLomas:
The `/gnite` reflection pass from the spend-ceiling session. Both directions of the ratchet.

## Up

New `.claude/docs/invariants/spend-controls.md`, with one routing line in `CLAUDE.md`.

Last night found **four distinct ways a spend ceiling fails**, and the reason they need writing down is that they are indistinguishable from outside — all four present as "I set a limit and it didn't hold":

1. **Wrong meter** (#3826) — guard read one store, logger wrote the other
2. **Ungated paths** (#4436) — 5 of 7 orchestrator spending phases never asked
3. **Dispatch ≠ consumption** (#4446) — gated creating work, not draining a queue
4. **Committed but unpriced** — OPEN; batch cost is invisible between submit and collect (26% overshoot measured)

`scripts/audit/spend-perimeter.mjs` already fails CI on the mechanical cases. This doc carries only what a check cannot assert — chiefly that **presence of a guard is not coverage by it**, which is how the audit's own first version passed the worker that then drained a queue through the ceiling.

## Fixed in place

The OCR routing line claimed flash-lite was for "everything else", omitting the **non-Latin and unknown-language** branches. That is the *safety* half of the rule — flash-lite doesn't fail on low-resource scripts, it invents plausible text — and the omission is roughly why 4,281 non-Latin books (Chinese, Tibetan, Parthian, Sogdian) were about to be routed to it.

## Down

The corpus-stats paragraph moves to `visibility-and-stats.md`. It's a measurement snapshot that says of itself "re-measure before quoting", it only fires when someone quotes a count, and that doc already owns the subsystem. `CLAUDE.md` keeps the rule and the canonical filter.

**The figures were appended to the target before the pointer went in.** A demotion whose destination doesn't hold the content is deletion with a citation — and I'd just spent a night on exactly that failure shape, so I checked rather than assumed (the pointer *was* dead when I first wrote it).

Net **5,554 → 5,543** words: a routing line added, a wrong line corrected, and still under where it started.

### DIFF
--- a/.claude/docs/invariants/spend-controls.md
+++ b/.claude/docs/invariants/spend-controls.md
@@ -1,0 +1,85 @@
+# Spend controls — the dial, and the four ways it has failed
+
+**Read this when** you are adding or changing anything that calls Gemini from a
+worker, cron, or scheduled job; adding a line to `infrastructure/hetzner-crontab`
+or `vercel.json` crons; touching `scripts/lib/spend-guard.mjs`; or asking "why
+did the daily budget not hold?"
+
+The dial is `system_config.processing_control.daily_budget_usd`, changed **only**
+via `scripts/maintenance/set-dial.mjs` (versioned — it snapshots the prior doc).
+Unset/0 means no paid dispatch: default-closed on purpose.
+
+There is a standing check — `scripts/audit/spend-perimeter.mjs`, wired to CI on
+changes to workers, the crontab, or `vercel.json`. It fails on an ungated
+spending phase, an unclassified schedule, or a worker whose `main()` does not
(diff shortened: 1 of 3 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek**. Half of it is being superseded today: #4770 corrects the same CLAUDE.md line about flash-lite routing. What remains unique is `invariants/spend-controls.md` (absent from main) and the demotion of the corpus-stats paragraph to `visibility-and-stats.md`. Conflicts with a CLAUDE.md that has churned since; the rebase is editorial under the word budget, not mechanical.

PR #4470, opened 2026-08-31
fix(security): revalidation routes fail closed — unset REVALIDATE_SECRET disabled auth entirely

JDerekLomas:
Found live during the #4389 post-merge work: `POST /api/admin/revalidate` accepted a request with **no secret and no session** (112 paths revalidated, 200). The guard was `if (secret && header !== secret) 401` — with `REVALIDATE_SECRET` unset in production, auth was skipped entirely. Same shape in `revalidate-book/[id]`. (`revalidate-authors` is fine — it uses `withAdminAuth`.)

**Impact:** anyone could force origin re-renders of arbitrary paths — a cache-busting lever that spends our compute and could bury the origin under a purge loop. No data exposure.

**Fix:** shared `isRevalidateAuthorized()` in `src/lib/revalidate-auth.ts`, fail-closed: no configured secret ⇒ nothing authorized. Accepts `REVALIDATE_SECRET` **or** `CRON_SECRET` (header or Bearer), so every existing pipeline caller keeps working — the scripts that pass `$REVALIDATE_SECRET` today are passing it empty (it's unset locally too) and were riding the fail-open; they should switch to `$CRON_SECRET`, which they all have.

In scope: the two routes + guard + unit/structural tests (7 passing; the structural half fails CI if the fail-open shape returns). Out of scope: setting REVALIDATE_SECRET in Vercel (unnecessary once CRON_SECRET is accepted), updating script call sites to pass CRON_SECRET (they work today only because of the hole; after merge they need the header — listed in the follow-up comment).

⚠️ Merging this deploys production and immediately 401s any caller not sending a secret. Known callers all have CRON_SECRET available.

https://claude.ai/code/session_01QDovjARwRT8q8nFbMLeGWd

### DIFF
--- a/src/app/api/admin/revalidate-book/[id]/route.ts
+++ b/src/app/api/admin/revalidate-book/[id]/route.ts
@@ -3,6 +3,7 @@ import { revalidatePath } from 'next/cache';
 import { getDb } from '@/lib/mongodb';
 import { findBookByIdOrSlug } from '@/lib/book-lookup';
 import { purgeCloudflareUrls } from '@/lib/cloudflare-cache';
+import { isRevalidateAuthorized } from '@/lib/revalidate-auth';
 
 export const dynamic = 'force-dynamic';
 
@@ -12,10 +13,8 @@ export async function POST(
 ) {
   const { id } = await params;
 
-  // Optional: verify a simple shared secret for pipeline calls
-  const authHeader = request.headers.get('x-revalidate-secret');
-  const secret = process.env.REVALIDATE_SECRET;
-  if (secret && authHeader !== secret) {
+  // Shared secret for pipeline calls — fail-closed (see src/lib/revalidate-auth.ts).
+  if (!isRevalidateAuthorized(request)) {
     return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
   }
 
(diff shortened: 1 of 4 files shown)


JDerekLomas:
Script call sites that must send `x-revalidate-secret: $CRON_SECRET` (or Bearer) after this merges — today each rides the fail-open (passes an unset `$REVALIDATE_SECRET`, i.e. nothing):

- `scripts/lib/triage-apply.mjs:108` (fetch) and `:99` (printed curl)
- `scripts/maintenance/cleanup-orphan-gallery-images.mjs:159`
- `scripts/maintenance/reextract-missed-pages.mjs:330`
- `scripts/maintenance/fix-digitizer-covers-4276.mjs` (documented curl)
- `scripts/create-proposed-collections-2026-08.mjs:370` (printed curl)
- `scripts/split-book-v2.mjs:599` — passes `?secret=$REVALIDATION_SECRET` as a **query param the route never read**, and the env name is misspelled (`REVALIDATION_` vs `REVALIDATE_`); it has only ever worked because of the hole.

These are one-line changes each; happy to sweep them in a follow-up PR after this merges (kept out of this diff so the security change stays reviewable alone).

Issue #4455, opened 2026-08-30
extractPageType screens out three page types the OCR prompt asks for


JDerekLomas:
Split out of #4443, which consolidated the six forked OCR-result parsers and could not fix this
without changing what the collectors write.

## The gap

`VALID_PAGE_TYPES` in `src/lib/types/prompts/defaults.ts` lists 17 values. The OCR prompt in the
same file offers 19, and three of them are not in the set:

```
musical-score    table    cover
```

`extractPageType` screens the model's answer against `VALID_PAGE_TYPES`, so when the model does
what the prompt asked and answers `musical-score`, the canonical parser returns `undefined` and
the page is stored with no page type.

This is the inverse of the bug fixed in #3591 / `tests/unit/page-type-vocabulary.test.ts`. That
one was *reader branch exists, prompt cannot produce the value*. The prompt was then widened to
offer all three — and `VALID_PAGE_TYPES`, which sits between the prompt and the database, was
not. So the value is now produced, rendered by `NotesRenderer.DESCRIPTION_ONLY_PAGE_TYPES`, and
dropped in the middle.

## Why it did not surface in #4443

Four batch collectors carried private copies of `extractPageType` that screened nothing at all,
so they *do* write `musical-score` / `table` / `cover`. Consolidating them onto the canonical
would have silently stopped that, so #4443 kept their behaviour behind
`extractPageType(text, { validate: false })` and left this to be decided on its own.

That flag is the marker for this issue: **every `{ validate: false }` call site should disappear
when this is fixed.**

## Scope

- [ ] Decide the source of truth for the vocabulary — most likely derive `VALID_PAGE_TYPES` from
      the prompt's own `One of:` line, or assert the two agree in
      `tests/unit/page-type-vocabulary.test.ts`, so they cannot drift a third time.
- [ ] Add `musical-score`, `table`, `cover` (and check `digitizer-notice`, referenced by
      `SKIP_TRANSLATION_PAGE_TYPES` but absent from both lists).
- [ ] Drop `{ validate: false }` from the four collectors once the sets agree.
- [ ] Count how many pages are already typed with the three values by the unvalidated collectors,
      and how many engraved-music pages are still typed `text`.

Issue #4443, opened 2026-08-30
OCR-result parsing is forked across 6 scripts and the canonical parser is imported by none of them


JDerekLomas:
Split out of #3979 item 2, which has had no movement since 2026-08-13. Splitting because the one
item on that map that *did* get fixed (`saveRevisionBeforeOverwrite`) is the one that was pulled
into its own issue (#3977 → PR #3981); the four that stayed in the map did not move.

## The defect

`src/lib/types/prompts/defaults.ts` is the canonical OCR-result parser (`extractPageType`,
`parseMultiPageOcr`). **`scripts/lib/` has no twin of it**, so every script-side consumer carries
its own copy and imports nothing.

Measured on `origin/main`, 2026-08-31 — **6 private copies + the canonical = 7 files**:

```
scripts/batch/collect-batch-results.mjs
scripts/batch/collect-multipage-ocr.mjs
scripts/batch/realtime-ocr.mjs
scripts/batch/realtime-reocr-efm.mjs
scripts/split-book.mjs
scripts/workers/batch-collector.mjs
src/lib/types/prompts/defaults.ts        <- canonical, imported by none of the above
```

That is **worse than when #3979 measured it** ("×4+" on 2026-08-13).

## Why this one first

- **It fails invisibly.** A prompt-tag change silently no-ops in the unpatched copies. The
  symptom is *missing metadata*, not an error — nothing goes red. #3979 recorded
  `collect-multipage-ocr.mjs` pinned to prompt v4 while live was v5.
- **It is on the paid path.** These are the batch collectors that read results we have already
  spent money to generate. A parse that silently drops `<page-type>` means the write-guards that
  depend on it (blank-page fabrication, #4149/#4184) are reading a field the parser never
  populated.
- **The fix is already specified and small.** #3979 names it: a `scripts/lib/` twin of
  `defaults.ts`, parity-pinned by a unit test — the `page-image-url.mjs` pattern this repo
  already uses for exactly this TS/JS split.

## Scope

- [ ] Add `scripts/lib/ocr-result-parse.mjs` (or extend an existing lib file) mirroring
      `defaults.ts`'s `extractPageType` / `parseMultiPageOcr`.
- [ ] Parity test pinning the twin against the TS canonical, in the shape of the existing
      `page-image-url` parity test — so a future edit to one fails CI rather than drifting.
- [ ] Replace the 6 private copies with imports. Diff each first: **a copy may have drifted for a
[...]

PR #4483, opened 2026-08-31
fix(pipeline): five collectors parsed a `<detected-images>` format the prompt never asked for (#4456)

### LINKED ISSUE #4456: Script-side parseDetectedImages parses a format the OCR prompt stopped emitting
Found while doing #4443. Not fixed there — it is a different parser, not a fork of the canonical
one, and fixing it changes what four collectors write.

## The defect

`parseDetectedImages` exists in two incompatible shapes:

- **canonical** (`src/lib/types/prompts/defaults.ts`) — `JSON.parse`s the contents of
  `<detected-images>`, which is what the current OCR prompt asks for:
  `<detected-images>[{"description": ..., "type": ..., "bbox": {...}}]</detected-images>`
- **scripts** (`collect-batch-results.mjs`, `collect-multipage-ocr.mjs`, `realtime-ocr.mjs`,
  `realtime-reocr-efm.mjs`, `workers/batch-collector.mjs`) — walks `<image>...</image>` sub-tags
  with `<description>` / `<type>` / `<bbox>` children.

JSON contains no `<image>` tags, so the scripts version finds nothing and returns `[]` on every
page the current prompt produced. The call sites all guard with
`if (detectedImages.length > 0)`, so nothing is written and nothing errors — the collectors just
never record detected images.

## Measured

Sampled 200 production pages holding a `<detected-images>` block (2026-08-31):

| shape | count |
|---|---|
| JSON-shaped | 200 |
| XML `<image>`-shaped | 0 |

188 of the 200 have a populated `detected_images` field — written by the canonical parser on the
`src/` path (`ocr-processor-logic.ts`, `batch-ocr-async/route.ts`), not by these collectors.

So the loss is confined to pages whose OCR arrived through a script-side collector. Worth
quantifying before fixing: pages with `<detected-images>` in `ocr.data`, an `ocr.source` of
`batch_api`, and no `detected_images` field.

## Scope

- [ ] Count the affected pages.
- [ ] Point the five scripts at the canonical JSON parser via `scripts/lib/ocr-result-parse.mjs`
      (the twin added in #4443), rather than porting the XML walker.
- [ ] Decide whether the XML shape is worth keeping as a fallback for pages OCR'd under an older
      prompt, or whether those are better handled by a backfill.
- [ ] Parity-test it alongside the other functions in `tests/unit/ocr-result-parse.test.ts`.

Mayank-PPL:
Closes #4456.

## The defect

`parseDetectedImages` existed in two incompatible shapes:

- **canonical** (`src/lib/types/prompts/defaults.ts`) — `JSON.parse`s the contents of `<detected-images>`, which is what the OCR prompt asks for.
- **scripts** (5 copies) — walked `<image>…</image>` sub-tags with `<description>`/`<type>`/`<bbox>` children.

JSON contains no `<image>` tags, so the five scripts returned `[]` on every page the prompt produced. Every call site guards with `if (detectedImages.length > 0)`, so nothing was written and nothing errored — the collectors simply never recorded a detected image, for as long as they have existed.

## Measured (full scan, not a sample)

`scripts/audit/detected-images-parse-gap.mjs --full`, 2026-08-31 — a 76-minute scan of every page in `pages` holding a `<detected-images>` block:

| | pages | books |
|---|---|---|
| holding a block | **31,210** | — |
| XML `<image>`-shaped | **0** | — |
| `ocr.source: batch_api`, block, **no** `detected_images` | 901 | 75 |
| `ocr.source: batch_api`, block, field present | 728 | 20 |
| `ocr.source: ai`, block, **no** `detected_images` | 3,562 | 439 |
| `ocr.source: ai`, block, field present | 25,992 | 748 |
| `(none)` / `manual` | 27 | 4 |

Two things the numbers say that are easy to get wrong:

1. **`ocr.source` cannot isolate this defect.** `realtime-ocr.mjs` and `realtime-reocr-efm.mjs` stamp `ocr.source: 'ai'` — the same label the canonical Lambda path uses. So the 901 `batch_api` pages are unambiguously a collector loss, but an unknown share of the 3,562 `ai` pages are too. Upper bound 4,470; don't quote the `ai` bucket as if it were all this bug.
2. **The repair set is mostly recoverable.** Of 200 sampled pages with a block and no field, 197 parse cleanly under the canonical parser. 3 do not (malformed JSON or no `description`) — those are model output, not this bug, and a backfill would correctly leave them alone.

## What changed

- All five scripts import `parseDetectedImages` from `scripts/lib/ocr-result-parse.mjs` (the twin added in #4443). The local walkers are deleted.
- `parseDetectedImages` added to the twin as an exact mirror of the TS canonical, using `VALID_IMAGE_TYPES` from `scripts/lib/gallery-image-types.mjs`.
- **No XML fallback.** 0 of 31,210 pages are XML-shaped and no prompt in this repo has ever asked for that shape, so a fallback would be dead code — and it would add a branch the TS canonical does not have, which is exactly the drift `tests/unit/ocr-result-parse.test.ts` exists to prevent. The test pins that both sides ignore the XML shape identically.
- New `scripts/audit/detected-images-parse-gap.mjs` — `--sample N` for a fast shape check, `--full` for the exact count. Exits 2 when the DB is unreachable (UNKNOWN, not clear).
- 13 new tests: full parity across 22 fixtures, plus behaviour pins for the clamp, the all-or-nothing bbox, the vocabulary screen, and reasoning-as-value in `type` (#3419 shape).

### Schema note

The two shapes disagreed about what they *wrote*, not just what they read. The walkers emitted a pixel `bbox: {x1,y1,x2,y2}` plus a `subject` array that is not even on the `DetectedImage` type, and no `detection_source`/`detected_at`. Because they never wrote a row, there is no legacy pixel-bbox data from this path to migrate — and a test pins that the pixel four-tuple is now **dropped, not coerced**: `detected_images.bbox` is read as fractional, so a pixel box read as fractions would crop the whole page.

## Verification

- `npx vitest run tests/unit/ocr-result-parse.test.ts` — 33 pass.
- `npx vitest run tests/unit` — 247 files, 3,275 tests pass.
- `npx tsc --noEmit` — clean.
- `node --check` on all six touched `.mjs` files.
- `eslint` could not run in this checkout (`TypeError: Class extends value undefined` from `@typescript-eslint/utils`) — pre-existing, reproduces on untouched files.

## Out of scope

- **The backfill of the ~4,470-page repair set.** A corpus write sweep is a different risk profile from a parser consolidation, and the mechanism already exists: `src/app/api/_archived/admin/backfill-detected-images/route.ts` does exactly this re-parse with the canonical parser. Reviving that is a deliberate change of its own (note it also writes `page_type` in *validating* mode, which disagrees with what the collectors write — that needs settling first). Filed separately.
- **The `detected_images` clobber question.** An OCR write replaces `detected_images` wholesale, so a re-OCR can downgrade rich `detection_source: 'vision_model'` rows to thinner `ocr_tag` ones. That is pre-existing behaviour on the canonical `src/` path (`ocr-processor-logic.ts` → `write-processor-logic.ts`, no guard anywhere), and this PR makes the collectors consistent with it rather than inventing a guard for five scripts and not the main path. It applies to every OCR write path and belongs in its own issue.
- **The `page-type` vocabulary gap** (#4455) — untouched.


Mayank-PPL:
Added a second commit: the  upward pass. The paragraph in `.claude/docs/invariants/tests-that-are-not-guards.md` records the class this defect belongs to — a parser whose contract is with a *prompt*, where a parity test between implementations proves nothing and `if (parsed.length > 0)` at the write site converts a dead parser into silence. Subsystem-triggered, so the invariant doc rather than CLAUDE.md; the downward pass found nothing in CLAUDE.md this duplicates or contradicts, and the word budget is untouched.

PR #4525, opened 2026-08-31
test(api): off-infra guard flags image files, not IIIF manifests

JDerekLomas:
Follow-up to #4522. The new "no off-infra image hosts" case flagged `iiif.archive.org` on a book response — but the URL in question is a **manifest.json**: an interop document describing the object, the same category as a DOI or a catalogue landing page, and genuinely useful to scholars. The other six archive.org references on that response are Dublin Core `dc_source`, the item landing page, a scanner hostname, a dedup fingerprint, and the `attribution` links added in #4522 — all provenance, none an image endpoint.

So the test was over-broad, not the API. A real IIIF **Image API** request always carries a `/full/` region segment; a manifest never does. The guard now matches image file extensions plus `/full/`, so it still fails on any actual image served from someone else's host.

Tightened rather than relaxed — the thing it exists to catch is unchanged.

**27/27 against production.**

https://claude.ai/code/session_01Ve8GjsAYmoR791UjeYCXtc

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -161,7 +161,7 @@ const nextConfig: NextConfig = {
       // POST is for /api/mcp (JSON-RPC); the Mcp-* headers are its session
       // protocol, exposed so browser MCP clients can read the session id.
       {
-        source: '/api/:root(search|books|gallery|collections|catalog|verify|mcp|image|locus|ngrams|openapi|dataset)/:path*',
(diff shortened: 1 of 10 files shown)


JDerekLomas:
Superseded by a clean branch off main — the original branch predated #4522's squash merge and could not rebase cleanly.

PR #4527, opened 2026-09-01
fix(archive): bound concurrency so a JP2 run cannot OOM the box + off-box liveness watchdog

JDerekLomas:
## A 25-hour silent outage of every scheduled job on Hetzner

Found while answering "are there missing images, and is the archiver draining them?"

On **2026-08-31 06:41** the box went global-OOM and systemd's oom-kill took down **`cron.service` itself**. Until I restarted it, **every scheduled job stopped**: archiving, the hourly `auto-pull` that deploys `main` to the box, health snapshots, batch collection. Nothing alarmed — a dead scheduler writes no logs to notice, and the archive log sat at 0 bytes, which is indistinguishable from "nothing to do" (a failure mode this script's own header warns about).

```
Active: failed (Result: oom-kill) since Mon 2026-08-31 06:41:30 UTC
Memory: peak 14.7G          # box has 15G
Out of memory: Killed process 1485513 (opj_decompress) anon-rss:1013460kB
```

## Cause

JP2 pages are decoded by **`opj_decompress`**, an external binary Node's heap limits cannot bound — measured **~1.0 GB RSS each**. At concurrency 8, a JP2-heavy stretch is ~8 GB of decoders against 15 GB shared with the rest of the pipeline. syslog carries **687 OOM events**, naming `opj_decompress`.

## What this is NOT: a throughput fix

My first draft of this PR claimed lowering concurrency would raise throughput. **Measured today, that is not supported** and I've corrected it:

| run | concurrency | rate | failures | memory |
|---|---|---|---|---|
| Aug 31 (during OOM storm) | 8 | 0.10 p/s | 57% | 14.7 GB → OOM |
| my test, today | 4 | 0.81 p/s (peak 1.41) | 11.5% | 4–6 GB |
| live cron, today | 8 | 1.05 p/s | 9% | 5 GB |

Concurrency 8 with healthy memory performs the same as 4. The Aug-31 numbers were **the OOM storm itself** — decoders being killed and counted as failures — not the concurrency setting.

So 4 is **insurance against the tail**: most pages decode cheaply, and the box only dies when enough large JP2s align. It costs little measurable throughput and removes a failure mode that takes the entire scheduler down with it. Concurrency here is a memory budget, not a throughput dial; the comment now says so with the measurement, so the next person doesn't tune it back up.

## Done outside this diff

- `systemctl restart cron` — the outage is over; this commit stops it recurring.
- Verified no other units failed. Cron has since run normally (08:45 run completed 1,000+ pages).

## Follow-ups (#4528)

An off-box liveness check, and `Restart=always` so an OOM self-heals. The incident was invisible because the thing that would have reported it was the thing that died.

https://claude.ai/code/session_01Ve8GjsAYmoR791UjeYCXtc

### DIFF
--- a/.claude/docs/invariants/measurement-instruments.md
+++ b/.claude/docs/invariants/measurement-instruments.md
@@ -465,3 +465,38 @@ Rules for any third-party search you draw conclusions from:
 
 Same shape as the guard-reads-the-wrong-store entry above: the instrument was
 healthy-looking and pointed at nothing.
+
+## A monitor must not share a failure domain with what it monitors
+
+On 2026-08-31 06:41 the Hetzner box went global-OOM and systemd's oom-kill took
+down **`cron.service` itself**. Archiving stopped, the hourly `auto-pull` that
+deploys `main` stopped, health snapshots stopped — for **25 hours**, and nobody
+knew. It was found only because someone went looking for an unrelated answer.
+
+Every check that would have reported it **was a cron on that box**. The daily
+health snapshot could not run. The archive log did not fill with errors; it sat
+at **0 bytes**, which reads exactly like "nothing to do" — and the very script
+writing it carries a header explaining that a held lock and an empty queue are
+indistinguishable in the log, a distinction it was rewritten to fix. It fixed
+the case where the script *runs*. A script that is never invoked says nothing at
+all, and silence had already been established as normal.
+
+- **Site the watchdog outside the blast radius.** `scheduler-liveness.mjs` runs
+  on **GitHub Actions** — not Hetzner (dies with it), not Vercel (the other
+  single point of failure). If you cannot name a failure that would take out the
+  subject but not the monitor, you have not built a monitor.
+- **Absence of an alarm is not health; it is usually no signal at all.** Prefer a
+  *liveness* assertion ("something reported within N hours") over an *error*
+  assertion ("no failures were logged"). The second cannot fire when the
+  reporter is dead, which is precisely when you need it.
+- **"Could not measure" must be as loud as "broken."** `scheduler-liveness.mjs`
+  exits 2 on an unreachable database and the workflow fails the run — an
+  unreadable store must never look like a quiet, well-behaved pipeline.
+- **Positive-control the alarm, not just the metric.** Force the thresholds and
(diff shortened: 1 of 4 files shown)


JDerekLomas:
Added the watchdog Derek asked for (ntfy alerts when archiving stops) — it belongs with this fix rather than a separate PR, since the outage and its detection are one story.

**`scripts/audit/scheduler-liveness.mjs` + `.github/workflows/scheduler-liveness.yml`**, running every 2 hours on **GitHub Actions** — deliberately not Hetzner (dies with it) and not Vercel (our other single point of failure). The whole lesson of #4528 is that the check must have a failure domain disjoint from the thing it checks.

Signal is `cron_runs`: every Hetzner worker writes a row, so "when did anything last run?" is one indexed query with no new bookkeeping. It asks two separate questions — has *any* cron reported (catches total cron death, the actual incident), and has an `archive-*` cron reported (catches the archiver dying while the rest lives). Alerts push to `ntfy.sh/sourcelibrary-uptime`, the topic the other watchers already use.

Verified both branches against production:
- healthy → exit 0 (`scheduler 0.3h ago, archive-ocr 0.3h ago`)
- forced stale thresholds → exit 1 with the right message

Two things testing caught that production would have:
- The first version read `started_at` on the newest row — which the `scheduler` rows don't carry — and reported a **live** system as "never ran". A watchdog that cries wolf trains you to ignore it; it now falls back to the ObjectId timestamp, which every row has.
- Exit 2 ("could not measure") is surfaced as loudly as a failure, so an unreadable database can't look like a quiet, healthy pipeline — that being the exact shape of the original mistake.

JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek**. Partly absorbed: `archive-acquired.ts` on main defaults to 6×6 citing this PR's OOM measurement (#4656, #4739). Not absorbed: the Hetzner cron still launches at concurrency 8 (`archive-acquired-cron.sh:33`), #4561 is still open, and the off-box liveness watchdog (`scheduler-liveness.mjs` + workflow, which needs DB credentials in Actions) does not exist. The scheduler was rewritten under it, so it needs a rebase and a decision on the watchdog.

Issue #4534, opened 2026-09-01
Tile-stitched page masters are two-thirds white: 483k-page cohort, 79,410 OCR'd Tibetan pages confirmed


JDerekLomas:
## What happened

`fetchIiifNativeRes` rebuilt page masters by requesting IIIF region tiles and
compositing them onto a **white** canvas at `left = col*chunk`, without checking
what came back. `scripts/maintenance/rearchive-iiif-fullres.mjs` sized its stride
from the host's **advertised** `maxWidth` — EAP advertises 2000 and serves 1200,
which is the entire reason EAP is on `SILENT_CAP_HOSTS` in the first place.

Every 2000px cell therefore received a 1200px image pasted at its top-left and
kept an 800px white gutter. 0.6 linear, 0.36 area. The stored master is 63.5%
pure white, every line of text truncated at a gutter and resuming 800px later.

It rewrote **`photo` as well as `archived_photo`**, so the reader, IIIF and the
OCR pipeline all get the gapped image.

```
3888x2592  white=0.635  x[1200,2000)              y[1200,2000)
3504x2336  white=0.623  x[1200,2000)              y[1200,2000)
4752x3168  white=0.603  x[1200,2000) x[3200,4000) y[1200,2000)
```

## Scope

The run was 18–20 July 2026. `image_metadata.upgraded_at` marks the cohort:
**483,134 pages across 1,998 books** — Latin 208,619, Tibetan 80,981, Greek
80,940, Hebrew ~36,000, German 17,493, Chinese 12,647, Syriac 7,886, Arabic
7,283, Ge'ez 5,253.

**The cohort is not the damage** — the same script ran correctly on hosts that
honour their advertised cap. Within Tibetan the marker is a near-perfect proxy
(21/21 guttered pages carry it; 0/210 clean pages do), which gives an exact
count there:

> **80,981 Tibetan pages / 167 books carry the marker. 79,410 have OCR text.
> 74,344 have a published translation** — read off, and written from, images
> that are two-thirds missing.

The other six `SILENT_CAP_HOSTS` (Manchester, e-rara, TU Delft, Kyoto RMDA,
IRHT, HAB) still need measuring; a first underpowered pass hinted the damage may
be confined to EAP, but that is a hint, not a finding.

## Why nothing caught it

R2 serves a real, complete, 200-OK JPEG. The blank-page guard keys on ink
coverage and these leaves are dense. The OCR model reads the fragments as a
[...]

Issue #4536, opened 2026-09-01
FT demolition: finish #3881 — card becomes the only writer, old machinery off then deleted


JDerekLomas:
Derek's directive 2026-09-01: "do it and get it done." This executes the deletion half of #3881 (passes 3–6), which was decided on 2026-08-11 and never executed. Since then two systems have been live at once — the Translation Card (`work_translation_history`, live on book pages since #3910/#3925) and the condemned book-grain machinery (attempt-ledger derive → reconcile valve → `books.first_translation`), still actuating nightly at 05:30 on Hetzner.

Context: the 2026-09-01 session (PR #4524, handoff `2026-09-01-ft-estimate-tibetan-ocr-and-eleven-corrections.md`) measured the old layer's verifier over-claiming priors ~20%, and spent most of its eleven self-corrections inside machinery already sentenced to deletion.

## Plan, in order

**1. Off-switch (operational + PR).** Retire the 05:30 Hetzner cron (`derive-ft-verdict-from-attempts.ts` + `reconcile-first-translation-flag.ts`) with a `#RETIRED-3881` comment, same pattern as the six `#RETIRED-2933` FT crons. Re-snapshot `scripts/workers/crontab.production` in a PR. Badges freeze except through reviewed card edits. Reversible by uncommenting.

**2. Citation harvest (evidence → cards).** The one valuable organ of the old layer is citations: `priors[]` on `first_translation_attempts` (11,323 books; 6,608 with named translator, 3,227 with URL). Harvest into `work_translation_history` entries under the card hygiene rules (#3901: no citation → no entry; defeater policy: durable identifier required to defeat). Verdicts are NOT migrated — per the 2026-08-11 decision they are opinions, not evidence, and the ~57K `gemini_verifier` rows carry a measured ~20% false-found rate. Proposals land as unreviewed card *suggestions* for reviewer merge, not silent bulk writes (card method rule 3).

**3. Deletion PR(s).** Delete the machinery the card replaces: derive/reconcile scripts, the verdict vocabulary and valve, the `src/lib/first-translation/` modules with no reader-facing consumer (verified by grep, not by graph tools). Deletion-class: wants second-dev eyes per repo policy. Scripts with historical value move to `scripts/_archived/`, not deleted from history.

[...]

PR #4545, opened 2026-09-01
Withdraw the Tibetan first-translation claims (#4523)

JDerekLomas:
## Withdraw the Tibetan first-translation claims

**The write is already applied** (2026-09-01, on Derek's explicit go-ahead). This
PR carries the script that did it, so the operation is reviewable and repeatable.

402 Tibetan books carried a first-family verdict (`first_no_prior`); 398 were
rendering a "First Translation" badge on card surfaces. Two independent findings
say the claim cannot rest on this text:

- **The instrument.** [GlotOCR Bench](https://arxiv.org/html/2604.12978v1)
  measures our production OCR model, `gemini-3.1-flash-lite`, at **19% Acc@5 on
  Tibetan clean printed text** with 100% script accuracy — it reliably emits
  Tibetan script and unreliably emits the right Tibetan. Cross-run agreement on
  our own manuscript pages runs 26–56% against ~87–93% for a printed Latin
  control, and some pages came back as Devanagari Hindu scripture (#4523).
- **The images.** 80,981 pages across 167 Tibetan books were archived from
  tile-stitched masters that are 63.5% white canvas (#4534). 74,344 of them
  carry a published translation generated from an image with two thirds of the
  folio missing.

400 of the 402 were `evidence_strength: weak` / `resolver: tier1_catalog` — the
blind-catalog tier `first-translation-reference-set.md` calls "effectively
unsearched".

### Why `needs_review` and not `not_first`

We are not claiming a prior translation exists. We are saying the claim cannot
rest on this text. `needs_review` is restorable per book and is exactly the state
a competent re-OCR should clear.

### What it deliberately does not write

- **`first_translation_attempts`** — that ledger records SEARCHES. We did not
  search. A row there would say we asked, and the nightly reconcile loop reads
  it, so writing there is actuation rather than recording.
- **`is_first_translation`** — already `false` on all 402. The badge renders from
  the VERDICT via `isFirstTranslation()`, so flipping the boolean would have
  changed nothing on the page. Single-writer stays
  `reconcile-first-translation-flag.ts`.
- **Page text, translations, visibility.** The text stays; only the claim about
  it is withdrawn.

Prior verdicts are preserved as `book_events{type:'ft_withdrawn'}` rows — one per
book, a ROW rather than a new column on 400 books (#3969) — so each withdrawal is
individually reversible.

`updated_at` is bumped on purpose: `sync-books-catalog.mjs` selects by
`updated_at > lastSync` and mirrors `first_translation.verdict` into the catalog's
`ft_verdict`. Without the bump the write is real in Atlas and inert on every card.

### Verification

| check | before | after |
|---|---|---|
| Atlas: Tibetan first-family verdicts | 402 | **0** |
| `books_catalog.ft_verdict = first_no_prior` (Tibetan) | 402 | **0** |
| `books_catalog.is_first_translation = true` (Tibetan) | 0 | 0 |
| `book_events{type:'ft_withdrawn'}` | — | **402** |
| API `first_translation=true&language=Tibetan` | — | **0 books** |
| positive control: same query, `language=Latin` | — | **5 books** |

### Two measurement mistakes made and corrected on the way

Both worth reading, because both were the same error — reading absence from an
instrument never shown capable of producing a presence.

1. I checked whether the badge was live by grepping a book page for
   "First Translation", found it, and called that a positive control. Calibrating
   against a *known-badged* Latin book showed it has zero badge occurrences above
   the "Related books" rail too — the detail page does not render that badge
   server-side at all. The string I found was a card in the related-books rail,
   for a Chinese book.
2. The first API check used `firstTranslation=true`; the route reads
   `first_translation`. The unfiltered response read as a clean result in one
   direction and an alarming one in the other. The table above uses the correct
   param with a positive control.

### Follow-up in flight

ISR + Cloudflare revalidation of the 402 book pages is running on Hetzner. The
first pass took 339 HTTP 429s — a bare curl UA at 5 req/s into the proxy's bot
limiter (10 req/60s). Retrying with a browser UA at 6.5s intervals.

Refs #4523, #4534

https://claude.ai/code/session_016sWzVok6mHuuuXb2V5R4fe

### DIFF
--- a/scripts/maintenance/withdraw-ft-tibetan.mjs
+++ b/scripts/maintenance/withdraw-ft-tibetan.mjs
@@ -1,0 +1,167 @@
+#!/usr/bin/env node
+/**
+ * withdraw-ft-tibetan.mjs — withdraw the first-translation claim on the Tibetan
+ * corpus, pending a trustworthy transcription (#4523, #4534).
+ *
+ * WHY (and why this is not a bibliographic demotion)
+ * --------------------------------------------------
+ * A "first English translation" badge asserts something about the BIBLIOGRAPHY.
+ * Every other demotion path in this repo answers a bibliographic question: does
+ * a prior translation exist? This one does not. It answers a prior question —
+ * is the thing we are calling a translation a translation at all?
+ *
+ * Two independent findings say no, for Tibetan:
+ *
+ *  1. THE INSTRUMENT. GlotOCR Bench (arXiv 2604.12978) measures our production
+ *     OCR model, gemini-3.1-flash-lite, at 19% Acc@5 on Tibetan CLEAN PRINTED
+ *     text, with 100% script accuracy — it reliably emits Tibetan script and
+ *     unreliably emits the right Tibetan. Cross-run agreement on our own
+ *     manuscript pages runs 26-56% against an ~87-93% baseline for a printed
+ *     Latin control. Some pages came back as Devanagari Hindu scripture.
+ *  2. THE IMAGES. 80,981 pages across 167 Tibetan books were archived from
+ *     tile-stitched masters that are 63.5% white canvas (#4534). 74,344 of them
+ *     carry a published translation generated from an image with two thirds of
+ *     the folio missing.
+ *
+ * So the verdict is `needs_review`, not `not_first`: we are not claiming a prior
+ * exists. We are saying the claim cannot stand on this text. It is restorable —
+ * this is exactly the state a re-OCR with a competent instrument should clear.
+ *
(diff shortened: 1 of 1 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: closing as superseded by #4524 (018e2df6), which added `scripts/maintenance/retract-tibetan-ft-claims.mjs` for the same #4523 decision, and by #4536, which removed the badge machinery entirely — #4523 confirms 1,467 visible Tibetan books carry zero first-translation badges. The reader-facing half shipped as e4b39b37 / 5838c77a / bcd177ce.

PR #4555, opened 2026-09-01
Put our social accounts where a reader can actually see them

JDerekLomas:
## Why

Derek noticed our social handles weren't on the site. They weren't — anywhere a person could see.

The only place any handle existed was the homepage JSON-LD `sameAs` array in `HomePageSchema.tsx`, listing X and GitHub. That's structured data: search engines read it, humans never do. `GlobalFooter.tsx` carried navigation, partners, the in-memoriam link and licensing, and no social links at all.

Instagram (`source.library`) appeared nowhere in the repo. The only Instagram references in the codebase are in-app-browser detection — about Instagram's webview breaking OAuth, not about our account.

## What

- `SOCIALS` list beside `PARTNERS` in `GlobalFooter.tsx`, rendered between the partner logos and the licence line: X, Instagram, GitHub.
- Inline SVG paths — no new dependency, no fetch, nothing to install.
- Instagram added to the homepage `sameAs` array.
- A comment on each list pointing at the other. Two lists of the same thing drift, and then readers and crawlers are told different things.

## Tenant behaviour

Gated on `useIsEmbedded()` — the same signal that already filters the footer's global-only nav columns (#3364). A partner reading room like `bph.sourcelibrary.org` is the partner's surface, and our accounts aren't theirs to carry. Resolved after mount, so the static HTML the global site serves is unchanged; only a tenant visitor sees the row drop.

## Out of scope

No Bluesky. I couldn't confirm an account exists, and an HTTP check is useless here: a deliberately fake handle returns 200 from Bluesky exactly as a real one does. If there's an account, it's a one-line addition to both lists.

No Facebook — there isn't one.

## Testing

`npx tsc --noEmit` clean. Visual check of the row and the tenant gate would be worth doing on the preview deploy.

https://claude.ai/code/session_01R5GrwxDWJqLLE9H1jYtMq9

### DIFF
--- a/src/components/layout/GlobalFooter.tsx
+++ b/src/components/layout/GlobalFooter.tsx
@@ -30,6 +30,28 @@ const PARTNERS: Partner[] = [
   { name: 'Frond Studio', src: '/partners/frond-studio-white.png', href: 'https://frond-studio.com', width: 400, height: 213, invert: false },
 ];
 
+// Source Library's own accounts. Kept beside PARTNERS so the two lists that
+// render in the footer's lower zones live together. These are also mirrored in
+// the homepage `sameAs` array (src/components/seo/HomePageSchema.tsx) — update
+// both, or search engines and readers see different sets.
+const SOCIALS: { name: string; href: string; path: string }[] = [
+  {
+    name: 'X',
+    href: 'https://x.com/SourceLibrary_',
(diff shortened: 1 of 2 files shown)


JDerekLomas:
Backlog sweep 2026-09-12: **held for Derek** — public footer links and homepage JSON-LD (reader-facing). Unlanded: footer has no social markup and `HomePageSchema` still lists only GitHub and X. Mergeable clean.

Issue #4600, opened 2026-09-03
OCR Lambda has no CI — merging a fix to main does not ship it


JDerekLomas:
Found while verifying #4591 was in effect.

## The problem

`src/lib/ai.ts` is bundled into four AWS Lambda workers. There is **no GitHub workflow and no
`infra/` directory** — deployment is entirely manual, per `.claude/docs/worker-architecture.md`:

```
scripts/aws-lambda/build-lambda.sh
scripts/aws-lambda/package-lambda.sh
aws lambda update-function-code --function-name sourcelibrary-ocr-processor \
  --zip-file fileb://dist/packages/ocr-processor.zip
```

So merging a fix to `main` deploys the **Next.js app** and updates **Hetzner** (hourly auto-pull at
:17), but leaves the Lambdas running whatever was last hand-deployed. Nothing warns you.

## Why it bit

#4591 merged 2026-09-03T00:29Z and fixes the thinking-budget defect in `src/lib/ai.ts` — worth
~$2,000/month across `worker/ocr` (124,953 calls in August), `worker/image-extraction` (10,811) and
`worker/translation` (8,110). All three are **Lambda** paths. Hetzner's auto-pull ran at 00:17Z,
twelve minutes before the merge. **The saving is not in effect and there is no signal that says so.**

This is the "merged ≠ in effect" failure mode, with a longer fuse than usual: the Vercel check goes
green, the PR closes, and the actual hot path is untouched indefinitely.

## Proposed

- [ ] Add a workflow that builds and deploys the Lambdas on push to `main` when `src/workers/**` or
      `src/lib/ai.ts` changes (or, at minimum, one that **fails loudly** when they change and the
      Lambdas were not redeployed)
- [ ] Stamp the git SHA into each Lambda bundle and expose it, so deployed-vs-main is checkable
      (`CODE_VERSION` already exists in `ocr-processor-logic.ts` — surface it in the usage rows)
- [ ] Grant the `sourcelibrary` IAM user `lambda:GetFunctionConfiguration` so deployment state can be
      audited at all. It currently cannot: `AccessDeniedException` on every worker.
- [ ] Add the manual deploy step to the PR template checklist for any PR touching those paths

Related: #4581, #4591.

JDerekLomas:
[...]

Issue #4599, opened 2026-09-03
~250K Gemini calls/month reach no usage log — ~58% of the token bill is unattributed


JDerekLomas:
Surfaced while sizing #4581. This is now **larger than the path #4581 fixed.**

## The gap

| August 2026 | successful `GenerateContent` calls |
|---|---|
| Google (Cloud Monitoring, `serviceruntime` consumed_api) | **404,000** |
| `gemini_usage` (our meter) | **~154,000** |
| **unlogged** | **~250,000** |

Those ~250K calls carry roughly **1,000M output tokens — about 58% of August's billed token
volume.** They come from analysis, eval, audit and enrichment scripts that call Gemini directly and
never write a usage row.

Everything the meter *does* see maps to five endpoints (`worker/ocr`, `worker/image-extraction`,
`worker/translation`, `script/ft-ladder`, and a handful of Hetzner scripts). Everything else is dark.

## Why it matters

1. **Cost is unattributable.** We cannot say what any workstream costs, which is the same problem
   that made the Google line unauditable in the first place.
2. **Thinking budgets are unaudited on exactly these scripts.** #4581 showed an unset
   `thinkingConfig` multiplies billable output 7.6× for no benefit. The batch scripts set
   `thinkingBudget: 0`; `src/lib/ai.ts` now does too. **The ~250K unlogged calls were never checked.**
3. It is why `gemini_usage` reported $499.74 against an $8,389.32 bill.

## Proposed work

- [ ] Enumerate every Gemini call site: `git grep -ln "generativelanguage\|generateContent\|GoogleGenAI\|getGenerativeModel" -- scripts src`
- [ ] For each, record two facts: **does it set a thinking budget**, and **does it log to `gemini_usage`**
- [ ] Route everything through a shared helper that does both, rather than 40 hand-rolled `fetch` calls
- [ ] Add a lint/test that fails on a direct `generativelanguage.googleapis.com` fetch outside the helper
- [ ] Backfill an `endpoint` label convention so per-workstream cost becomes answerable

## Note on budgets

`thinkingBudget: 0` is right for transcription and extraction. It is **wrong** for grounded search
and the FT ladder — `thinkingBudget: -1` silently suppresses grounding, and a fixed budget truncates
(see `scripts/eval/ft-ladder.ts` and `lesson_gemini_grounding_truncation`). The audit must classify
each site, not blanket-apply zero.

[...]

Issue #4581, opened 2026-09-02
OCR worker runs Gemini with thinking ON by default — bill counts it, meter can't see it


JDerekLomas:
## The finding

`performOCRWithBuffer()` in `src/lib/ai.ts` — the path the Lambda OCR workers use — calls
`model.generateContent({ contents, safetySettings })`. **No `generationConfig`, therefore no
`thinkingConfig`.** On Gemini 3.x thinking defaults to ON, and Google bills reasoning tokens at
the output rate.

Every batch script already does the right thing. `scripts/batch/realtime-ocr.mjs:201`,
`scripts/batch/bulk-reocr-local.mjs:426,451`, `scripts/batch/realtime-translate.mjs:98`,
`scripts/batch/batch-transliterate.mjs:134` all set `thinkingConfig: { thinkingBudget: 0 }`.
The guard exists — it just never travelled to the worker call site.
(Same shape as the "a guard travels with the FILE, not the pattern" lesson.)

## Evidence — four independent lines, all agreeing

1. **Code.** The call site passes only `safetySettings` (read 2026-09-02).
2. **Telemetry.** Cloud Monitoring labels **100%** of the `booksplit` project's August output
   tokens `thinking_enabled: "true"` (metric
   `generativelanguage.googleapis.com/generate_content_usage_output_token_count`).
3. **Live probe.** One call today, `gemini-3-flash-preview`, trivial prompt:
   `candidatesTokenCount: 4`, **`thoughtsTokenCount: 83`**, `totalTokenCount: 106`.
4. **Reconciliation.** Metered August spend **$499.74** (`gemini_usage`). Billed August spend
   **$8,389.32**. A **17x** gap, against the 3.4x previously documented.

## Why the meter is blind

`performOCRWithBuffer` records `outputTokens = usageMetadata.candidatesTokenCount`, which
**excludes** `thoughtsTokenCount`. Google bills both. So the bill counts thinking and our ledger
does not — which is how "the pipeline is paused since 2026-06-08" and "$8,389/month" have coexisted
for three months.

Unit economics this produces: August's $8,389.32 bought **224,944** pages of OCR (counted via the
`ocr.updated_at` index, so re-OCR is included) = **$0.0373/page**, against the $0.0034/page the
grant proposal validated. 1,712M output tokens / 224,944 pages = **7,606 output tokens per page**,
where a page of transcription is normally 500-1,500.

## Proposed fix

```ts
[...]

PR #4601, opened 2026-09-03
Reconcile Gemini spend against the vendor; fix two drifted price tables (#4581, #3576)

### LINKED ISSUE #3576: gemini_usage cost_usd: two price tables disagree 3.3×/5× for the same model

JDerekLomas:
## Why

We had three spend instruments and **all three were internal**:

| instrument | what it reads |
|---|---|
| `true-gemini-spend.mjs` | our own meters |
| `spend-perimeter.mjs` | whether every spender asks the dial |
| `daily-health-snapshot.mjs` | `batch_jobs.cost_usd` |

None ever asked Google what it actually charged. That is precisely how a **17×** gap survived three
months — August metered **$499.74** against a billed **$8,389.32** (#4581).

## What this adds

`scripts/audit/spend-reconcile.mjs` — closes the loop against the vendor.

- **Billed tokens** from Cloud Monitoring, per project, per model. Includes reasoning tokens, which
  our meters excluded.
- **Live prices** from the Cloud Billing SKU catalogue —
  `cloudbilling.googleapis.com/v1/services/AEFD-7695-64FA/skus`, 634 SKUs. An **API**; the console is
  not needed, and it is a better source than the pricing page, which omits some models entirely.
- **Diffed against `gemini_usage`**, with meter *coverage* as a first-class number.
- **The rest of the surface** — R2 computed from Cloudflare analytics; every vendor we cannot read
  printed as `UNREADABLE` with the reason. An omitted line reads as $0, so nothing is omitted.
- **Exit 2 on price drift**, so CI can gate on it.

On August it reports `$5,558` estimated billed vs `$499.74` metered (11.1×), and meter coverage of
**37%** — 263,048 successful calls write no usage row (#4599).

## What this fixes

Two price tables had drifted from Google **and from each other**:

| table | model | was | now (list) |
|---|---|---|---|
| `src/lib/ai.ts` | `gemini-3.1-flash-lite` | 0.075 / 0.30 | **0.25 / 1.50** |
| `src/lib/ai.ts` | *default* | 0.10 / 0.40 | 0.75 / 3.75 |
| `scripts/lib/model-pricing.mjs` | `gemini-3.1-flash-lite` | 0.075 / 0.30 | **0.25 / 1.50** |
| `scripts/lib/model-pricing.mjs` | `gemini-3.6-flash` | 1.50 / 7.50 | **0.75 / 3.75** |
| `scripts/lib/model-pricing.mjs` | *default* | 0.50 / 3.00 | 1.50 / 9.00 |

Plus `3.5-flash`, `3.5-flash-lite`, `3.6-flash`, `3.7-flash` added to `ai.ts`, which had no entries
and were silently falling through to `default`.

### This resolves #3576

`model-pricing.mjs` documented the flash-lite disagreement as open, kept 0.25/1.50 as
`DISPUTED_LEGACY_PRICING`, and chose 0.075/0.30 as the *"closest fit to recorded costs."*

**That test was circular** — recorded `cost_usd` was itself computed from 0.075/0.30, so of course it
fit. Google's catalogue gives **0.25 / 1.50**. The disputed pair was correct all along.

## Behavioural change — please read before merging

`gemini_usage.cost_usd` is **computed** from these constants, never billed. Correcting them means:

1. **Rows written after this merge price higher than historical rows.** Do not compare a
   pre-2026-09-03 `cost_usd` to a later one.
2. **The daily-budget dial reads these constants**, so it will trip *sooner* on the same real spend.
   That is the safe direction — the dial has been permissive because the numbers were low — but it is
   a real change to unattended behaviour, and worth knowing on a `$5/day` setting.
3. Defaults are now deliberately **expensive**, so an unpriced model overstates and gets noticed
   rather than hiding under a cheap fallback.

## Traps encoded in the script (each cost real time to find)

- Cloud Monitoring `crossSeriesReducer=REDUCE_SUM` **double-counts** token series ~1.9×. The script
  never passes one; it aligns per series and sums them itself.
- The `thinking_enabled` metric label reads `"true"` **even at `thinkingBudget: 0`**. It is not
  evidence of thought billing and the script never branches on it.
- `gcloud billing projects list` **default-paginates** — six projects are billing-enabled, not the
  five an unpaginated call returns.
- Prices differ by **serving tier**: batch is exactly half, priority output is 1.8× standard. Match
  the tier or the number is wrong.
- A quoted `.env` value silently becomes a malformed Cloudflare `accountTag` and returns *"not
  authorized for that account"* — which reads as a permissions problem and is not one.

## Verification

- **Negative control:** deliberately corrupting one price makes it exit **2** and name the row;
  restoring it exits **0**. Run both ways, not just the passing one.
- An unreadable price table counts as **drift**, not a pass — a silent skip is how an unpriced lane
  hides.
- `npx tsc --noEmit` passes.

## Out of scope, deliberately

- **The Lambda deploy.** #4591's fix is merged but the OCR Lambda has no CI (#4600) — it still runs
  the old code. Nothing here changes that.
- **Enabling the BigQuery billing export.** It would make the Google half exact rather than
  estimated, but it is not retroactive and it is an account-level change for Derek to make.
- **The Cloudflare token scope.** The production token can read R2 objects but not R2 *analytics*, so
  the R2 line prints UNREADABLE with that reason. Adding "Account Analytics Read" would light it up.
- **Atlas**, still the last unmeasured vendor.


JDerekLomas:
## CI caught a real one — and it is the most interesting thing in this PR

`tests/unit/model-pricing-single-source.test.ts` failed, correctly. It asserted
`gemini-3.1-flash-lite` = **0.075 / 0.30** as canonical and held **0.25 / 1.50** as
`DISPUTED_LEGACY_PRICING`.

It had the two backwards — and **because it was a test, it was actively defending the wrong number.**

The original reasoning was that 0.075/0.30 was the *"closest fit to recorded costs."* That test was
circular: recorded `cost_usd` was itself computed from 0.075/0.30, so it could only ever fit. Google's
SKU catalogue settles it at **0.25 / 1.50**.

### What changed in `ce5e6a1a`

- `DISPUTED_LEGACY_PRICING` → `SUPERSEDED_PRICING_BEFORE_2026_09_03`, now holding the rate historical
  rows were **actually costed at** (0.075/0.30, plus `3.6-flash` at 1.50/7.50). Kept rather than
  deleted — those rows are still in the database, and they need to stay re-derivable instead of being
  silently compared against corrected ones. The file's original virtue (don't quietly drop the other
  rate) survives; only which one is canonical has flipped.
- The test now asserts the **resolution** plus the superseded record, and leaves verifying the
  *number* to the thing that can reach the source of truth: `spend-reconcile.mjs`.
- Added a case pinning `BATCH_MULTIPLIER = 0.5`. Batch is exactly half across every model in the
  catalogue, so pricing batch work at the standard rate overstates ~2× — the mirror image of the
  failure this test file was originally written about.

### The lesson, which outlives the constant

A test that pins a **constant** pins whatever was believed the day it was written, and then defends
it against correction. Pin the **property**; let an instrument that can reach the source of truth
check the value. That is the division of labour now: the test asserts *"the superseded rate is still
recorded"*, and `spend-reconcile.mjs` asserts *"the canonical rate matches Google today"* — and exits
2 when it stops matching.

4/4 pass locally; `spend-reconcile` reports no drift in either table.

Issue #4584, opened 2026-09-02
Translation phase fabricates text where OCR declined: live Egyptian funerary lines with no source (Urkunden IV p.24)


JDerekLomas:
## The failure

On a page where OCR **succeeded on the prose and honestly declined on the glyphs**, the translation phase invented the missing content — fluent, plausible, and citable.

Live example: *Urkunden des ägyptischen Altertums* (Sethe, Abt. IV), **p. 24** — https://sourcelibrary.org/q/BiX3lunYwsGV9PQm3Dc (`visible: true` today).

The OCR field ends with a bracketed stand-in:

> `Nur die letzten Zeilen sind erhalten` / `x+1` / `[Hieroglyphic text lines x+1 through 20]`

The translation field renders those twenty unread lines as twenty lines of English:

> x+3. ...I was one who spoke the truth before the Lord of the Two Lands.
> x+9. ...my name shall endure upon this stone.
> x+20. ...peaceful in the field of reeds.

None of it has a source. The OCR layer said "I cannot read this"; the translation layer treated that sentence as content to be rendered rather than as an absence. Worse, x+3 was promoted into the book's `reading_summary.quotes` with page 24 attached, so it surfaces as a featured quotation of the volume.

This is adjacent to the blank-page fabrication guard (shipped, default-on in both writers) and to #4523, but it is a **different door**: the page is not blank, OCR did not fail, and the guard has nothing to fire on. The gap is that a placeholder in the OCR output is not treated as a no-text signal downstream.

## Second, separate failure on the same volume

Degenerate repetition. p. 118 OCR loops `𓂋𓂝𓏤` and the translation dutifully renders "Ra, Ra, Ra, Ra, Ra, Ra…"; p. 182 emits "Father, behold;" dozens of times. That is a loop artifact being published as text.

## What I did and did not measure

Confirmed **by reading the page**: 1 fabrication (p. 24), 2 repetition loops (pp. 118, 182).

[...]

Issue #4195, opened 2026-08-21
OCR prompt v16: define <page-type>, narrow 'blank', kill the DISCURSUS specimen, add a lacuna marker


JDerekLomas:
One prompt revision closing four documented defects at once. Each change below names the issue it fixes and the evidence for it, so this can be reviewed on the merits rather than on my say-so.

The OCR prompt lives in the `prompts` collection (`{ type: 'ocr', is_default: true }`, currently v15) and is read by `getOcrPromptFromDb()` in `scripts/workers/pipeline-orchestrator.mjs`, which passes `content` to the model verbatim apart from `{language_instruction}` and `{language}`. Every OCR call in the system uses it, so this is a behaviour change across the whole pipeline and wants a deliberate version bump plus a read of the eval before/after.

---

## 1. Give `<page-type>` its definition back — #4149

**Problem.** Lines 34–35 of the live prompt, verbatim:

```
- <page-type>X</page-type>
- <columns>N</columns> — number of text columns on this page (omit for single-column pages,
  include for 2+ columns) — classify this page (REQUIRED). One of: title-page, frontispiece,
  dedication, preface, toc, index, errata, colophon, appendix, blank, illustration, diagram, map, text
```

The enum belongs to `<page-type>`. It has been appended to `<columns>`, a tag about counting text columns. `<page-type>` is the only metadata tag in that block with no definition of its own, so `blank` is effectively invisible as an option.

**Evidence.** All 50 confirmed fabricated pages carried `<page-type>text</page-type>`; none carried `blank`. The same merge is in `prompts/ocr/standard-ocr-v5.md` and `v10.md` — the exact versions every fabrication ran under — and in live v14/v15.

**Change.** Move the enum onto `<page-type>`; restore `<columns>` to its own meaning.

## 2. Define `blank` narrowly — #4149 **and** #3591

**Problem, from both directions.** Under-use fabricates (#4149: whole blank leaves given invented pages). Over-use suppresses (#3591 case 3: Kitāb al-Bulhān pages 266, 197 and 4 carry a legible basmala, a Latin cataloguer's note and `للمؤلف`, were classified `blank`, and translation then replaced them with `[Blank page — no translatable content]`).

**Change.** `blank` means **no ink at all**. State explicitly that faint, damaged, stained or partially legible pages are NOT blank — those keep their real classification and use `<unclear>` / `<warning>`.

[...]

Issue #3591, opened 2026-08-04
OCR fabricates text where the page is damaged or unreadable, instead of declining


JDerekLomas:
## Summary

Spot-checking **Kitāb al-Bulhān** (`6953b56577f38f6761bd979d`, Bodleian MS Or. 133) against the scans turned up a coherent defect class: **when the OCR model cannot read a region, it copies or generates plausible text rather than marking the region unreadable.**

This is not a legibility problem. Transcription quality on ordinary prose is *good* — better than my own reading in a blind comparison (see "What is working" below). The problem is the absence of an "I can't read this" option.

Same family as the quote-integrity rules in CLAUDE.md, at write time rather than read time: fabricated page text is a fabricated citation, and it is **more fluent** than real transcription, not less.

### 1. Gap-filling over physical damage — page 60

A 7×2 grid of lunar-mansion entries. A large tear removes the paper across the lower-right cells. The stored OCR supplies complete, fluent text for **all fourteen cells**, including the ones with no paper left.

Possible because the entries are formulaic (*"when the moon descends into X, perform Y… whoever is born male will be Z, if female W, and God knows best"*). The model even emitted a `<warning>` noting the damage, then filled the gap anyway.

**Corroborating tell — a closed-set violation.** The 28 lunar mansions are a fixed canonical list with no repeats. The OCR names **الطرف (al-Ṭarf) twice** in adjacent cells; by position in the sequence that slot should be **الصرفة (al-Ṣarfa)**. A scribe does not duplicate a mansion; a text generator filling a hole does.

### 2. Duplication into a second slot — page 159 (confirmed), page 41 (probable)

**Page 159** carries two headings. The manuscript shows a circle with **one dot** in the first and **two dots** in the second (lot-casting notation), so the headings must differ. The OCR *describes this correctly in its own image description* — "a single large dot", then "the central circle contains two dots" — and then transcribes the **identical heading text for both**: `قارعت الاثنين قارعت خمسة انا`.

The file contradicts itself. Note this needed no ability to read Thuluth: **internal contradiction is checkable even when the script is not.** That is probably the most useful detector here, because it scales to pages nobody can proofread.

**Page 41**: the `<margin>` block is a word-for-word copy of the body paragraph, differing only in its opening clause. A real marginal note exists on that page; I could not read it at available resolution, so this one is probable rather than confirmed.

### 3. Real text suppressed as blank — page 266 (+ 197, 4)

[...]

PR #4605, opened 2026-09-03
<lacuna>: a named marker for regions with no legible source (#4195 item 5, #4584)

JDerekLomas:
Implements **#4195 item 5** ("a named lacuna marker for unreadable REGIONS") together with the OCR/translation prompt revision that uses it, and closes the #4584 fabrication at its cause.

## The tag, and why it is not `<unclear>`

#4195 asks for the marker to be distinct from `<unclear>`, "which currently implies a guessable reading." That distinction turns out to be load-bearing rather than cosmetic:

`<unclear>X</unclear>` wraps a best-guess **reading** and unwraps to body text in every extraction path. So a *description of a gap* carried inside one would be quoted, counted into ngrams, embedded, and exported into PDFs and EPUBs as if it were words printed on the page. That is precisely the #4584 failure: Urkunden IV p.24's unread glyph block was described in prose, and the description came back as twenty lines of invented English funerary formulae — one of which was promoted into the volume's featured quotes.

`<lacuna>` therefore joins `EDITORIAL_WRAPPERS` in **both** stripper twins, so it is dropped content-and-all everywhere text is extracted. The reader renders it in **both** note states on purpose: a silently dropped region reads as a complete page, which is the failure the tag exists to prevent.

## Measured, v15 vs v17, against the exemplars the issues name

Not pages chosen to flatter the change — #3591's and #4149's own cases, plus counter-examples that must *not* move.

| case | v15 | v17 |
|---|---|---|
| #3591 p.60 torn mansion grid | 24,108 body chars (completes all 14 mansions) | 1,286, gaps marked |
| #4584 p.118 hieroglyph rows | 16,264 (the "Ra, Ra, Ra…" loop) | 318, apparatus kept, 8 `<lacuna>` |
| #4584 p.24 hieroglyph block | invented `x+N` lines | one `<lacuna>`, no invented lines |
| #4149 p.30 ordinary prose *(counter-example)* | 2,485 | 2,485 |
| #3591 p.198 cipher *(counter-example)* | declines | still declines |

Harness committed at `scripts/eval/ocr-prompt-v17-acceptance.mjs` so this is re-runnable rather than a claim.

## What I could NOT establish

**#4195 items 2–3 (blank-page narrowing) are not demonstrated.** p.4 still classifies `blank` despite the model describing a faint mark, and p.45 still emits untagged prose. I sharpened the rule with a self-consistency check and re-ran; it did not fix either.

Worse for the measurement: **v15 itself returned body=79 then body=0 on consecutive runs of the same page.** These blank cases are unstable run-to-run, so n=1 cannot separate prompt from noise. I've left those boxes unticked on #4195 — #3469's labelled hard-page corpus is the instrument that would settle it.

## Ordering — please note

The prompt rows are inserted **non-default** (`ocr` v17, `translation` v14). They must not be promoted until this merges, or the pipeline emits a tag the renderer shows as literal text. After merge:

```
node scripts/maintenance/ocr-prompt-v17-lacuna.mjs --promote
```

`--demote` rolls back. Live default stays v16/v13 meanwhile — an interim I wrote earlier today before finding #4195, which achieves the same fabrication fix by overloading `<unclear>`; it is safe because `<unclear>` is fully handled downstream, and v17 supersedes it.

## Out of scope

- #4195 item 4's DISCURSUS specimen swap and items 1/6 are *in* v17 but not separately measured — they are template hygiene, not behaviours with a clean before/after on a single page.
- #3591's fixes 2–4 (self-contradiction detector, closed-set validators) are detectors, not prompt text, as that issue says.
- Repairing already-fabricated text. Prompts only affect future runs.

## Test

`npx tsc --noEmit` clean. 73 unit tests pass, including two new ones pinning that `<lacuna>` is dropped by both stripper twins while `<unclear>` survives as body text. Negative-controlled: an unlisted tag leaks its content verbatim, so the test genuinely fails without the change.

### DIFF
--- a/scripts/eval/ocr-prompt-v17-acceptance.mjs
+++ b/scripts/eval/ocr-prompt-v17-acceptance.mjs
@@ -1,0 +1,91 @@
+#!/usr/bin/env node
+/**
+ * Acceptance run for OCR prompt v17 (#4195 checklist + #4584).
+ *
+ * Runs v15 (current default) and v17 side by side over the exemplars the issues
+ * name, so each claim is checked against the page that produced it rather than
+ * a page chosen to flatter the change. Two of the cases are counter-examples
+ * that must NOT change — a prompt that only ever declines is not an improvement.
+ *
+ * Usage: node --env-file=.env.production.local scripts/eval/ocr-prompt-v17-acceptance.mjs
+ */
+import { MongoClient } from 'mongodb';
+
+const BASE = 'https://generativelanguage.googleapis.com/v1beta';
+const KEY = process.env.GEMINI_API_KEY || process.env.GEMINI_API_KEY_2;
+if (!KEY) throw new Error('no GEMINI_API_KEY');
+const MODEL = 'gemini-3.1-flash-lite';
+const LANG_INSTR = `**Source language:** Detect the primary language from the text. Pages may contain multiple languages — transcribe all of them. Report the primary language in the <language> tag (e.g. <language>Latin</language>).`;
+
+const BULHAN = '6953b56577f38f6761bd979d';   // Kitab al-Bulhan — #3591 exemplars
+const ZUNI = '69a565b95a8a09c1b325e47f';     // Zuni Fetiches — #4149 blank leaves
+const URK4 = '69e013c593b116d24238b3d7';     // Urkunden IV — #4584 fabrication
+
+const CASES = [
+  { label: '#3591 torn mansion grid',   book: BULHAN, page: 60,  expect: 'lacuna, and NOT 14 complete mansion names' },
+  { label: '#3591 legible basmala',     book: BULHAN, page: 266, expect: 'NOT blank — real text transcribed' },
+  { label: '#3591 cataloguer note',     book: BULHAN, page: 197, expect: 'NOT blank' },
+  { label: '#3591 li-l-muallif',        book: BULHAN, page: 4,   expect: 'NOT blank' },
+  { label: '#3591 cipher counter-ex',   book: BULHAN, page: 198, expect: 'must STILL decline (counter-example)' },
+  { label: '#4149 blank leaf',          book: ZUNI,   page: 45,  expect: 'page-type blank, no body/header/sig/page-num' },
+  { label: '#4149 blank leaf',          book: ZUNI,   page: 72,  expect: 'page-type blank, no body' },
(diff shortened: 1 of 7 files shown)


JDerekLomas:
## Retraction — the measured table above does not replicate

I built the k-run instrument (PR #4610) and re-ran this comparison at **k=5** instead of k=1. Two of the figures I posted are single draws from very wide distributions and are withdrawn.

| case | what I reported (k=1) | replication (k=5) |
|---|---|---|
| torn-grid p.60 | "24,108 → 1,286" | v15 1,561 ±313 → v17 7,160 **±10,222** — not established |
| hiero-block p.24 | "invented x+N lines → one `<lacuna>`" | v15 3,962 **±6,884** → v17 455 ±372 — not established |

With SDs of 10,222 and 6,884, those single runs told me nothing. Retracted.

**One row survives, and it is large.** The glyph rows: v15 16,264 ±0 → v17 348 ±4, a change 5,000× the pooled SD. v15 is *perfectly reproducible* — it emits the identical "Ra, Ra, Ra…" loop on all five runs. Ordinary prose is also unchanged (2,486 → 2,494), so the arm is not simply suppressing output everywhere.

**And a correction in the other direction.** I reported that items 2–3's blank narrowing "did not fix" p.4. Also n=1: across five runs v17 classifies it `text` every time, stably. **The narrowing works.** I said it didn't because I looked once.

## The finding that changes the recommendation

**v17 over-declines.** On two pages it converts readable text into gap markers — p.197's cataloguer note ("Nihil hic deesse videtur", perfectly legible Latin) becomes a `<lacuna>`; the p.198 cipher shrinks from 295 to 119 chars.

That is exactly the trade this issue warned about — *"do not simply make `blank` easier to reach. That trades #4149 for #3591"* — running in the direction I wasn't watching, because one run per arm only shows the fabrication side.

**So v17 should not be promoted as it stands.** It needs another pass targeting the over-declining, then a re-run. The rows stay non-default meanwhile.

## Also: my decision rule was flawed

Clause (b), "agreement must not fall", misfires when the baseline failure is a deterministic loop. v15 scored agreement=1.000 on the glyph rows while being reproducibly wrong, so the biggest real win in the set was scored "agreement fell". Next revision should use gap-marker rate as the substantive criterion and keep agreement as a floor. I have left the rule unchanged in the committed script and documented the flaw — re-scoring after seeing the numbers is what pre-registration is for.

PR #4610, opened 2026-09-03
eval: paired k-run prompt A/B — and it retracts the v17 numbers I published

JDerekLomas:
Adds the instrument that was missing when #4195 and #4584 could not be scored — and runs it. **The first result is that most of my own single-run claims from earlier today do not replicate.**

## Design

- **Paired** — both arms see the same pages. Page-to-page variance dwarfs the arm effect; pairing removes it.
- **Repeated measures** — k runs per (page, arm). This is the whole point: with k=1 there is no way to separate a prompt effect from the sampler.
- **Unit of analysis is the page.** The k runs are repeated measures *within* a page, not independent observations; aggregating raw runs would fake precision.
- **Controls both ways.** Positive controls are pages where the defect is known present and the metric must fire; negative controls are clean pages where nothing should move. A metric that cannot fire is not a measurement.
- **Pre-registered** outcomes and decision rule, fixed in the file before the run.

## Solving the missing-labels problem

We have no ground-truth transcriptions, so "is this fabricated?" is not directly observable. The way around it is already validated in this repo by #4523: **cross-run agreement**. A genuine reading is reproducible (87–93% measured on printed Latin); invention is not (31–35% on Tibetan cursive). No human labels required.

The logic is one-directional, and that matters: low agreement is strong evidence of invention, but high agreement is *weak* evidence of correctness, because a model can be reproducibly wrong. That caveat bit on the very first run — see below.

## What replication showed (v15 vs v17, k=5)

| case | v15 | v17 | verdict |
|---|---|---|---|
| hiero-rows *(pos)* | 16,264 ±0 | 348 ±4 | **real** — 5,000× pooled SD |
| ordinary-en *(neg)* | 2,486 ±0 | 2,494 ±4 | unchanged |
| torn-grid *(pos)* | 1,561 ±313 | 7,160 **±10,222** | not established |
| hiero-block *(pos)* | 3,962 **±6,884** | 455 ±372 | not established |
| cipher *(neg)* | 295 ±64 | 119 ±67 | **regression** |
| cataloguer *(neg)* | 67 ±23 | 30 ±0 | **regression** |
| faint-mark *(pos)* | type=`blank` | type=`text`, 5/5 | blank narrowing works |
| blank-45 *(pos)* | 32 ±43 | 105 ±15 | worse — more untagged prose |

## Retractions

The two headline figures in **PR #4605** and in my comment on **#4195** — torn-grid "24,108 → 1,286" and hiero-block "invented lines → one lacuna" — were **single draws from distributions with SDs of 10,222 and 6,884**. They are retracted. I have corrected both.

Conversely, I reported that #4195's blank narrowing "did not fix" the faint-mark page. That was also n=1: across 5 runs v17 classifies it `text` every time. The narrowing works and I said it didn't.

## The finding that matters

**v17 over-declines.** On two negative controls it converts readable text into gap markers — the cataloguer's note ("Nihil hic deesse videtur", perfectly legible Latin) becomes a `<lacuna>`. That is precisely the trade #4195 warned about — "do not simply make `blank` easier to reach; that trades #4149 for #3591" — running in the direction I was not watching, because with one run per arm I could only see the fabrication side.

**v17 should not be promoted as it stands.** One large genuine win, two regressions, and most cases unresolved.

## Known flaw in the rule, left unfixed on purpose

Clause (b) ("agreement must not fall") misfires when the baseline failure is a *deterministic loop*: v15 scored agreement=1.000 on the glyph rows while emitting the identical "Ra, Ra, Ra…" every run — perfectly reproducible, perfectly wrong. A better rule uses gap-marker rate as the substantive criterion and treats agreement as a diagnostic floor.

That change is **not** applied here. The verdicts were produced under the rule as pre-registered, and rewriting it after seeing the numbers is the exact failure pre-registration prevents. It is documented in the file header and should be fixed prospectively, in the next revision.

## Cost

100 calls per full run on flash-lite — a few cents. Cheap enough to be the default before any prompt promotion.

### DIFF
--- a/scripts/eval/lib/paired-stats.mjs
+++ b/scripts/eval/lib/paired-stats.mjs
@@ -1,0 +1,69 @@
+/**
+ * Paired / two-sample statistics for eval harnesses.
+ *
+ * PROVENANCE: lifted verbatim in method from `scripts/eval/stats-cross-model.mjs`
+ * (#3235), which has used these three tests on paired per-page deltas since
+ * 2026-07. Extracted here so a second harness does not hand-roll a weaker rule —
+ * `prompt-ab.mjs` originally compared arm means against a pooled SD, which is an
+ * ad-hoc threshold with no error rate attached, while this machinery was already
+ * in the repo.
+ *
+ * `stats-cross-model.mjs` still carries its own copies; migrating it to import
+ * from here is a separate change (one concern per PR) and is the obvious
+ * follow-up. Until then, if you fix a bug here, fix it there too.
+ *
+ * The bootstrap PRNG is seeded so a CI is reproducible across runs — a
+ * confidence interval that moves when you re-run the report is not auditable.
+ */
+
+let seed = 0x5eed;
+export const resetSeed = (s = 0x5eed) => { seed = s; };
+const rand = () => (seed = (seed * 1103515245 + 12345) & 0x7fffffff) / 0x80000000;
+
+export const mean = (a) => (a.length ? a.reduce((s, x) => s + x, 0) / a.length : null);
+
+/** Exact two-sided sign test: 2 * P(X <= min(k, n-k)), capped at 1. */
+export function binomTwoSided(k, n) {
+  if (!n) return 1;
+  const logC = (nn, kk) => { let s = 0; for (let i = 0; i < kk; i++) s += Math.log(nn - i) - Math.log(i + 1); return s; };
+  const lo = Math.min(k, n - k);
+  let p = 0;
+  for (let i = 0; i <= lo; i++) p += Math.exp(logC(n, i) - n * Math.LN2);
+  return Math.min(1, 2 * p);
+}
+
+/** 10k-resample bootstrap 95% CI on the mean of `xs`. */
+export function bootstrapCI(xs, iters = 10000) {
+  if (xs.length < 2) return null;
+  const means = [];
+  for (let i = 0; i < iters; i++) {
+    let s = 0;
+    for (let j = 0; j < xs.length; j++) s += xs[Math.floor(rand() * xs.length)];
+    means.push(s / xs.length);
+  }
+  means.sort((a, b) => a - b);
+  return [means[Math.floor(iters * 0.025)], means[Math.floor(iters * 0.975)]];
+}
+
+/**
+ * Bootstrap 95% CI on the DIFFERENCE OF MEANS between two independent samples
+ * (the k runs of arm A vs the k runs of arm B on ONE page). Resamples each arm
+ * independently, which is the correct structure here: the runs are independent
+ * draws from each arm's sampler, not paired with each other.
+ *
+ * Returns { delta, ci, decisive } — `decisive` is true when the CI excludes 0,
+ * which is the criterion that replaces the old "bigger than the pooled SD".
+ */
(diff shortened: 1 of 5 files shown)


JDerekLomas:
## Update: replication reverses the result, and the prior art was already here

Two corrections, both against my own work in this PR.

### 1. This duplicates existing infrastructure, and mine was worse

Asked what prior tests exist, I found the repo already has, and has had for months:

- **`PREREGISTRATION-prompt-ablation.md`** and **`PREREGISTRATION-repeat-instability.md`** — the house conventions this PR's header claims to introduce. The ablation preregistration is a 2×2 factorial with reference arms, per-hypothesis numeric thresholds, a cost table, and a named threats-to-validity section. It is better than my design.
- **`prompt-ablation.mjs`** — a multi-arm prompt runner with `--runs=N`, `--arms`, `--dry-run` cost estimation, arm P reading the live prompt from the DB, durable JSONL outputs.
- **`stats-cross-model.mjs`** — exact two-sided sign test, Wilcoxon signed-rank, and a 10k-resample bootstrap 95% CI on paired per-page deltas, with a seeded PRNG so intervals are reproducible.
- **`lib/production-prompt.mjs::withIntervention()`** — the anchor-insert-or-throw helper I rewrote as `once()` in the v17 maintenance script.

My pooled-SD threshold was an ad-hoc rule with no error rate attached, standing in for machinery that already existed. It is now replaced by the real thing, extracted to `lib/paired-stats.mjs` and credited. Migrating `stats-cross-model.mjs` to import from it is the obvious follow-up, deliberately not bundled here.

This is the fourth duplication of existing work in this session.

### 2. The finding does not replicate — the loop is not prompt-controlled

A second k=5 run of the same two arms reversed the headline:

| | run 1 | run 2 |
|---|---|---|
| p.118 v15 | **16,264 ±0** | 554 ±65 |
| p.118 v17 | 348 ±4 | **16,232 ±0** |
| p.24 v15 | 3,962 ±6,884 | **16,271 ±0** |
| p.24 v17 | 455 ±372 | 834 ±48 |

**~16.2k is the output-token cap.** A degenerate repetition loop runs until truncation, and it lands on **either arm**. Within a batch of five it is all-or-none — ±0, agreement 1.000 — which is precisely what makes a single batch look decisive when it is not. Both arms were verified to load distinct prompts by `content_hash`, so this is not an assignment bug.

So the "5,000× pooled SD win" I reported for v17 and its mirror image in this run are **the same artifact**. Neither is evidence.

`lib/production-prompt.mjs` documents this exact class of error from a previous incident — *"an alarming 'fabrication' finding that was purely an artifact of the harness."* I reproduced it.

### What the next revision needs

Body-length means are the wrong estimator: a looped run is a categorical catastrophe, not a long reading, and averaging it with normal runs describes neither. Required instead:

1. a repetition classifier labelling each run looped / normal,
2. **loop rate** reported as a Bernoulli outcome with a proper interval,
3. length statistics computed on normal runs only,
4. tens of runs per arm — k=5 cannot estimate a rate this volatile.

**No v15-vs-v17 claim on looping pages survives, in either direction.** #4605 stays blocked. What this PR now contributes is the instrument, the corrected statistics, and two archived result sets that disagree — which is itself the useful artifact.

PR #4612, opened 2026-09-03
Reader: an honest 'not reliably legible' state for attempted-but-unreadable OCR (#4523)

JDerekLomas:
Part of the #4523/#4534 Tibetan re-OCR work — the mechanism steps 2–4 hang on.

## The gap this fills

Serving *good* re-OCR is trivial (write `ocr.data`, the reader renders it). But there was **no honest way to withhold a *bad* transcription.** The only lever was blanking `ocr.data`, which the reader shows as "Not transcribed yet" — a lie (it *was* attempted) that also erases the untried-vs-unreadable distinction. So today a fragmented Tibetan read, or the old *fabricated* Gemini OCR, gets served as if trustworthy.

Concretely: on one live book the old OCR rendered invented Hindu scripture ("O Vishnu… the great deeds of Shiva") on a Tibetan Buddhist biography. We need a state that says, plainly, *we tried and can't vouch for a reading* — and shows the scan instead.

## What this adds

- **`ocr.unreadable` (+ `unreadable_reason`)** on `OcrData` — set when a transcription was attempted but isn't reliably legible (the re-OCR adjudicator flags pages where two independent BDRC recognizers disagree). `data` is **retained for provenance**, never rendered.
- **Reader** (`ReaderV2Bits`): an `unreadable` page withholds **both** panes (a translation of untrustworthy OCR is untrustworthy too) and shows the scan as authoritative — gated *before* the empty-check, since `data` is non-empty. `PairedEdition` guards its collapsed supplementary-OCR disclosure too.
- **`PaneEmptyState`**: honest **"Not reliably legible"** wording (en + es), no request-translation CTA.
- **Counters** (`page-counts.ts` + its `.mjs` twin + `hasOcr`): an unreadable page does **not** count toward `pages_ocr` — the stat reflects legible pages, not attempts.
- **`apply-reocr-verdicts.mjs`**: consumes the eval-side adjudication (`SERVE` / `MARK_UNRELIABLE` / `TEXTLESS`). SERVE snapshots old text to `page_revisions` then writes new; MARK sets the flag and retains old text; TEXTLESS untouched. Reversible, **dry-run default**, every action reported.

## Relation to existing mechanisms
- **`recitation_blocked`** exists but is never reader-wired (a recitation-blocked page still renders its text). This is the reader-wired equivalent.
- The **#4546 Tibetan banner** (`PageMetadataPanel`) is a blanket per-language notice that does **not** withhold text. Left as-is; this is per-page and actually suppresses. A follow-up can make that banner defer to the per-page flag.

## Verification
- `npx tsc --noEmit` clean; `check-imports` clean.
- Adjudicator already run over 685 re-OCR'd pages: 407 SERVE, 248 MARK, 34 TEXTLESS; the clean *Bardo Thödol* scores BOOK_OK (407 serve-quality pages), the rougher demo book correctly BOOK_UNRELIABLE. SERVE pages show both models reading identical text; MARK pages show them diverging.

## Out of scope
- Writing verdicts to production (this is the *capability*; applying it is a separate reviewed run once re-OCR completes).
- The public first-translation badge withdrawal (Derek owns that).


JDerekLomas:
Follow-on to this PR's `notReliablyLegible` state, from the #4523 takedown (PR #4756, comment on #4523 has the full numbers).

**This state covered less ground than it looked like it did.** `ocr.unreadable` withholds both panes *in the reader* — and only in the reader. The flag is a flag: search, embeddings, the quote API, IIIF, the exports and the MCP tools all read `pages.translation.data`, see nothing, and served the translation of the unreliable transcription anyway. **22,099 pages** were in that state. `translation.data` is now unset on all of them (text preserved in `page_revisions`), so the withholding is uniform rather than reader-only.

Two additions here:

1. **A distinct state.** `notReliablyLegible` says the transcription is untrustworthy and withholds both panes. The new `translationWithheld` says the transcription is *good and shown* and only the English came down — a re-OCR'd page must not wear the first sentence, or we would be disclaiming the transcription we just fixed. English and Spanish in `src/lib/reader-strings.ts`; the copy is Derek's to approve.
2. **The general lesson.** A flag that one surface honours is a filter list of the surfaces you thought of. Where the point is to stop serving text, remove the text: `#4756` moves it off the document entirely, and every consumer stops at once.

Verified with the probe harness added in #4756 (`scripts/audit/withheld-translation-surfaces.mjs`): 20/21 request-path surfaces reached the text before, 0/21 after.

Issue #4613, opened 2026-09-03
Generation parameters are provenance: record thinking, temperature and tier on AI output


JDerekLomas:
## The claim that is no longer true

`.claude/docs/data-provenance.md` opens with:

> Every OCR transcription, translation, summary, index, chapter extraction, and image extraction in
> Source Library can be traced back to the exact prompt, model, trigger, and job that produced it.

That was the right list when it was written. It is now incomplete, and #4581 proved why: **two pages
with identical `model` and `prompt_version` can hold materially different text**, because the
generation parameters that produced them differ and nothing records them.

## What is recorded, and what is not

`pages.ocr` carries: `data`, `language`, `model`, `updated_at`, `source`, `prompt_version`.
(`pages.translation` additionally has `prompt_hash`, `prompt_id`, `prompt_name` — OCR does not, which
is a smaller gap worth closing in the same pass.)

**Not recorded anywhere:** `thinkingBudget`, `temperature`, `maxOutputTokens`, serving tier.

## Why it matters — measured, not theoretical

Two production OCR paths ran the **same model and the same prompt** with different parameters:

| | `src/lib/ai.ts` (Lambda, `worker/ocr`) | `scripts/batch/realtime-ocr.mjs` (Hetzner) |
|---|---|---|
| temperature | **1.0** (unset → model default) | **0.1** |
| maxOutputTokens | unset → 65,536 | 16,384 |
| thinking | **on** (unset) until #4591 | `thinkingBudget: 0` |

The temperature default is not an assumption — `GET /v1beta/models/gemini-3-flash-preview` returns
`temperature: 1`, `topP: 0.95`, `topK: 64`, `outputTokenLimit: 65536`. **We ran transcription at
temperature 1.0 on the busiest path.**

And the effect on output is large. The #4581 A/B (20 real pages, same images, only `thinkingBudget`
varying) produced **74% word agreement on Latin script and 43% character agreement on Classical
Chinese** between arms. Whatever else that says about OCR stability, it means thinking-on and
thinking-off pages are **different populations**.

In August alone, `source=ai` (the Lambda path, thinking on, temp 1.0) wrote **124,117** pages and
`source=batch_api` (thinking off) wrote **47,846** — all tagged with the same handful of
`model` + `prompt_version` combinations, and indistinguishable in the data.

## The consequence that is worse than bookkeeping

[...]

PR #4614, opened 2026-09-03
docs(provenance): generation parameters are provenance, and we don't record them (#4613)

JDerekLomas:
Doc-only. Flags a gap that #4581 exposed and #4613 will fix.

## The problem

`.claude/docs/data-provenance.md` opens by claiming every AI output traces back to *"the exact
prompt, model, trigger, and job."* That list is incomplete — **`thinkingBudget`, `temperature` and
`maxOutputTokens` are recorded nowhere, and they change the output text.**

Two production OCR paths ran the **same model and the same prompt** with different parameters:

| | `src/lib/ai.ts` (Lambda, `worker/ocr`) | `realtime-ocr.mjs` (Hetzner) |
|---|---|---|
| temperature | **1.0** — unset, so the model default | **0.1** |
| maxOutputTokens | unset → 65,536 | 16,384 |
| thinking | **on** until #4591 | `thinkingBudget: 0` |

The temperature default is verified, not assumed: `GET /v1beta/models/gemini-3-flash-preview`
returns `temperature: 1`. We ran transcription at 1.0 on the busiest path.

The #4581 A/B — 20 real pages, same images, only `thinkingBudget` varying — measured **74%** word
agreement on Latin script and **43%** character agreement on Classical Chinese between the two arms.
Whatever else that says, thinking-on and thinking-off pages are **different populations**, and
`model` + `prompt_version` cannot tell them apart. In August, `source=ai` wrote 124,117 pages and
`source=batch_api` wrote 47,846, all under the same handful of model/prompt combinations.

## Why it's worth a doc change now rather than waiting for the fix

`page_revisions` is used as a **double-OCR corpus** for agreement, calibration and page-difficulty
work. If two revisions were produced under different generation parameters and nothing records that,
then measured "disagreement" mixes model instability with an unrecorded config change — and every
metric built on it inherits the confound.

Anyone reading that corpus this month needs to know. The note gives them the rough proxy (`source`
correlates with the call site) and is explicit that it is a correlation, not a record.

## Scope

One paragraph added to the top of `data-provenance.md`. No code. The fix — recording the resolved
generation config on the output, and adding `prompt_hash` to `ocr` to match `translation` — is
**#4613**, which also notes the overlap with #4599: both have the same root, that there is no single
chokepoint an AI call must pass through.


JDerekLomas:
## Second correction to the same doc — found by validating my own claim

A background job checking one of my assertions in #4613 got killed, so I re-ran it. Good thing.

The doc's prompt-field block listed `prompt_id`, `prompt_hash` and `prompt_name` as present on
**both** `page.ocr` and `page.translation`, describing the hash as a *"cryptographic verifier"*.

Measured against production:

| field | `page.ocr` | `page.translation` |
|---|---|---|
| `prompt_version` | 300/300 | 300/300 |
| `prompt_hash` | **0/300** | `source=ai` **100%** · `source=batch_api` **19%** |
| `prompt_id` | **0/300** | same |
| `prompt_name` | **0/300** | same |

(300 most-recently-OCR'd pages; 600 most-recent translations for the by-source split.)

So **an OCR page cannot be cryptographically verified against the prompt text that produced it** —
only against a version *string*. Anyone writing a provenance check against `ocr.prompt_hash` reads
`undefined`, and if they treat absence as pass, they verify nothing. That is the same failure shape
as the `if (parsed.length > 0)` case in `tests-that-are-not-guards.md`: a missing value that reads as
success.

This also **sharpens a claim I made in #4613**, where I wrote that `translation` "has" these fields.
It has them on one of two write paths. Corrected there too.

Pushed as `cd486452`. Same file, same class of defect — the doc asserting provenance that isn't in
the data — which is what this PR is about. CI re-running.

JDerekLomas:
## Withdrawing "never" — my own claim from an hour ago was too strong

The commit above said `page.ocr` **never** carries `prompt_hash` / `prompt_id` / `prompt_name`. My
evidence was 0/300 recent pages. A single indexed probe then found a counterexample:

```
source=pipeline_preview   updated_at 2026-05-29T11:22:22.856Z
prompt_hash e1a8c7ee5b0a1a0239813eb881fb3491
```

So the field **is** written somewhere in history. What is actually established:

- **No current writer emits them** — 0/300 most-recently-OCR'd, 0/800 sampled across May 2026
- **They are not absent from history** — at least one `pipeline_preview` row carries a real hash
- **Exact historical coverage is UNMEASURED**, and is now labelled that way in the doc.
  `ocr.source` has no index, so a per-source count is a full scan of ~20M documents. I started one,
  watched it scan, and killed it — a documentation footnote does not justify that load on
  production. Better to say "unmeasured" than to publish a number I did not measure.

The practical guidance is unchanged and arguably firmer: **presence is rare and unreliable, so a
provenance check must not treat absence as pass.**

Pushed as `b53ca393`. CI re-running.

---

*Third correction to this doc in one session, and the second to my own text. Worth noting the
pattern: the first two came from measuring, and this one came from a **killed background job** —
which is what prompted me to re-check a claim I had already published. A killed job is not a
neutral event; it can be the only thing standing between an assertion and its evidence.*

Issue #4593, opened 2026-09-03
gemini_usage (Mongo) writers dropped created_at — every date-scoped spend read, including the $5/day dial, sees $0


JDerekLomas:
Found during the 2026-09-03 spend audit (context: #4581, which covers the thinking-token half of the blindness).

## Defect
The Mongo `gemini_usage` collection is written heavily — **166,322 rows between Aug 1 and Sep 2** (128,731 ocr, 17,730 extract_images, 11,593 ft_ladder_skeptic, 8,113 translation, plus translate/index/summary) — but **every one of those rows is missing `created_at`**. The last row that HAS `created_at` is from 2026-03-24. So:

- Any date-filtered query (`created_at: {$gte: ...}`) over the store returns **zero rows** and reads the meter as silent/$0. This audit initially concluded the store was dead since March — wrong, and exactly the failure mode any dashboard or the $5/day dial's spend check hits.
- Rows are currently findable only via `_id` timestamps (`ObjectId.createFromTime`).
- The Supabase `gemini_usage` store (primary sink, `timestamp` column) is unaffected and live.

## Fix shape
1. Find the writer(s) that stopped setting `created_at` (~late March; `supabase-usage-logger.mjs`'s Mongo fallback path is the first suspect) and restore the field.
2. Backfill `created_at` on the dateless rows from `_id` timestamps (cheap, exact to the second).
3. Point whatever enforces the daily spend dial at a read that can actually see the rows — and give it a positive control (a day known to have spend must read > $0), per the four-failure-modes lesson.

Related: #2247 (pricing table validation), #3685 (meter at the boundary), #3576 (two price tables disagree). This issue is only the missing-date defect; don't fold the others in.

🤖 Filed from the spend-audit session — https://claude.ai/code/session_01JsexXhYKKtEXN73km7dahV

JDerekLomas:
Measured against the live collection on 2026-09-05, **this issue is inverted** — and the real defect it uncovered is somewhere else. PR #4657 carries both halves.

## The `created_at` claim

[...]

PR #4622, opened 2026-09-03
docs: an instrument that reads only our own records can't find our records wrong

JDerekLomas:
Two doc changes from the #4581 spend audit. No code.

## 1. The generalisable lesson → `invariants/measurement-instruments.md`

We had **three** spend instruments, all well-built, all encoding real scar tissue:

| instrument | reads |
|---|---|
| `true-gemini-spend.mjs` | every meter we write |
| `spend-perimeter.mjs` | whether every spender asks the dial |
| `daily-health-snapshot.mjs` | `batch_jobs.cost_usd` |

**None ever asked Google what it charged.** So for three months they agreed with each other at
~$500/month against a billed **$8,389.32**. No amount of cross-checking *between* them could have
surfaced it — the missing quantity (reasoning tokens, billed at the output rate) was absent from
every source they read.

Two compounding failures, both recognisable in other domains:

- The cost was **computed from a constant, not observed** — so a wrong constant produced wrong
  numbers that were internally consistent and looked fine.
- The constant had been **"validated" against those same computed costs**. `model-pricing.mjs` chose
  `0.075/0.30` as the *"closest fit to recorded costs"* — where the recorded costs were generated by
  `0.075/0.30`. A constant fitted to numbers it produced is a mirror, and it fits perfectly forever.

**Rule:** when an external party is authoritative for the quantity, at least one instrument in the
set must reach outside and reconcile. Plus two corollaries this session shipped in
`spend-reconcile.mjs`: report **coverage** as a first-class number (it found only 37% of Gemini calls
write a usage row — #4599), and print **UNREADABLE with a reason** rather than omitting a vendor,
because an omitted line reads as $0.

## 2. A stale stat in `CLAUDE.md`

Line 183 quoted the Vercel bill as `~$3,069/mo` when warning against bulk imports through Vercel
functions. It has been **$1,750.15** since the #3648 and WAF work landed. Fixed in place.

The *argument* is unaffected — invocations are still a measured line item — but a routing line that
quotes a number 75% too high is the kind of thing that gets a doc distrusted wholesale.

## Budget note

`CLAUDE.md` is **5,639 words** against its ~5,500 target. Nothing was added to it here — the lesson
went to the invariant doc it routes to, and the stat was fixed in place with no word growth. Flagging
it so the next person who wants to *add* knows they need to demote something first.


JDerekLomas:
Added the session handoff (`a47e39bc`) — `.claude/handoffs/2026-09-03-spend-audit-thinking-tokens.md`, via `git add -f` per the CLAUDE.md convention for public-worthy technical postmortems.

Technical content only. The credential findings from this session are in the **private** ops repo and are deliberately absent here.

JDerekLomas:
Backlog sweep 2026-09-12: merged main (both sides append to measurement-instruments.md; kept both). One correction while resolving: the paragraph quoted "only 37% of calls write a usage row" — #4657 showed that was a one-store read; the doc now says 72% and cites #4593/#4657. CLAUDE.md stays at 5,562 words.

PR #4629, opened 2026-09-03
Withdraw the fabricated translation on Urkunden IV p.24 (#4584) — already applied

JDerekLomas:
Records the script that removed the live fabrication Derek approved withdrawing today.

**What was serving.** The OCR behaved correctly — it read the German apparatus and recorded the glyph block as unread, `[Hieroglyphic text lines x+1 through 20]`. The translation phase rendered those twenty unread lines as twenty lines of English funerary formulae with no source. One reached `reading_summary.quotes` and was served on a `visible: true` book as a quotation of Sethe — with an AI-written gloss about the ideal of Maat, for a sentence that does not exist.

**Applied to production 2026-09-03**, verified on a fresh read and on the live site after a Cloudflare purge:

```
invented lines remaining: 0
<lacuna> marker present:  true
p.24 featured quotes:     0  (book now has 14)
purge: success:true — 0 hits for the Maat line and 0 for x+N on the live URLs
```

**Method** borrowed from `quarantine-fabricated-ocr.mjs` (#4149): assert on a live read, preserve the prior text to `page_revisions`, then remove it from the read paths. Nothing destroyed.

**Surgical, not a delete** — unlike #4149 where the OCR is wrong and no correct text exists, here the OCR is right and only part of the translation was invented. The German-apparatus translation is real and is kept.

**The replacement is `<lacuna>`** from #4605, registered in `EDITORIAL_WRAPPERS` in both stripper twins, so the gap description cannot be quoted, counted, embedded or exported. A flag would have left the text in place for every consumer that reads `translation.data` and checks none.

**Scope, stated honestly.** All 310 pages scanned for the same signature (OCR placeholder + ≥5 enumerated lines); p.24 was the only hit. The detector does **not** catch the p.118 glyph loop or the p.182 repetition — separate defects, untouched.

Also dogfoods the prior-art guard from #4618: the file carries a `PRIOR ART:` line naming `quarantine-fabricated-ocr.mjs` and why it could not simply be reused.

### DIFF
--- a/.claude/handoffs/2026-09-03-harkhuf-and-the-fabrication-class.md
+++ b/.claude/handoffs/2026-09-03-harkhuf-and-the-fabrication-class.md
@@ -1,0 +1,79 @@
+# Harkhuf, Yam, and a fabrication class — 2026-09-02/03
+
+Started as "what do we have on Yam and Harkhuf?", became a session about the
+translation phase inventing text where OCR honestly declined. Written for a cold
+reader picking this up.
+
+## Where it stands
+
+**Live harm: withdrawn.** Urkunden IV p.24 was serving 20 invented lines of
+Egyptian funerary formulae, and one of them was a *featured quote* on a visible
+book. Removed 2026-09-03, prior text preserved to `page_revisions`
+(`source: withdraw-fabricated-translation-4584`), Cloudflare purged, verified 0
+hits on the live URLs. Script: `scripts/maintenance/withdraw-fabricated-translation-4584.mjs`
+(PR #4629).
+
+**The original request: half done.**
+- Breasted, *Ancient Records of Egypt* I (`6a989f98ba191f96aa9bab9c`) — the book
+  that actually carries Harkhuf §§325–336 and Pepi II's dwarf letter, in English.
+  **398/398 pages archived to R2, 62 OCR'd, hidden.** Verified ELIGIBLE for the
+  full OCR pass; blocked only by the $5/day dial being spent. It finishes on its
+  own when the dial resets — no action needed. Then QA and flip visible.
+- Sethe, *Urkunden des Alten Reichs* (`6a989fa2ba191f96aa9bad2d`) — the
+  hieroglyphic text edition. **Deliberately NOT processed.** Its glyph plates are
+  exactly the material that fabricates. Keep as a facsimile.
+
+## Merged today
+
+| PR | what |
+|---|---|
+| #4587 | standing "unverified transcription" notice on hieroglyphic pages |
+| #4610 | paired k-run prompt A/B instrument + retraction of my own numbers |
+| #4618 | prior-art guard (PreToolUse/Write), `scripts/eval/INDEX.md`, `EXPERIMENTS.md` |
+| #4605 | `<lacuna>` tag — stripped from every quotable/countable/exportable surface |
+
+Open: #4629 (withdrawal script), #4623 (research-agenda amendment).
+
+## Do NOT redo
+
+- **Do not promote OCR prompt v17 / translation v14.** They exist as
+  non-default rows and `--promote` refuses without `--ab-rerun-clean`. At k=5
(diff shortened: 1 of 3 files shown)


JDerekLomas:
## Corpus scan done — the class is bounded, and mostly a false alarm

Scanned **5,576 non-Latin-script books with translations (1.28M translated pages)** for the #4584 signature. Both controls asserted first — the positive control caught an earlier regex that **missed the known case** and would have returned a reassuring near-zero.

**12 pages flagged. Verified by hand, one at a time: 11 are false positives.**

The instructive one: *Seven Tablets of Creation* II p.41 looked like a textbook hit and is actually **exemplary**. The OCR transcribed the visible cuneiform and the translation rendered it faithfully — 𒀭𒈨 → "divine powers", 𒈗 → "king", 𒂗 → "lord" — with the `[ ... ]` breaks preserved on both sides. Three *Babylonian Liturgies* pages simply **repeat** the placeholder (`[Cuneiform text]`, `[illegible]`), which is precisely the behaviour we want. The Amduat pages' "[Transliteration … left to right]" is a heading introducing a transliteration the OCR did produce; Clement and Oxyrhynchus are the editions' own bracketed apparatus.

**2 real, now withdrawn:** *Book of the Dead* Vol. II pp.33–34. OCR transcribed **no** hieroglyphs (`[Hieroglyphic text lines 1-28]`) and the translation supplied numbered claims about what those lines say. **Milder than Urkunden p.24** — the content sits in AI-marked `<note>` tags rather than being served as the source's words, and none reached a featured quote — but it still tells a reader what an unread line 29 contains.

Same method: assert on a live read → preserve to `page_revisions` → replace with `<lacuna>`. p.24 correctly self-skipped as already withdrawn, so the idempotency guard works. Cloudflare purged.

**Bottom line: the fabrication class is small and now bounded** — 3 pages across 2 books corpus-wide, on this signature. The detector runs at **~8–17% precision**, so it is a work list and never an authority.

**What this does NOT cover**, stated plainly: the signature requires an English bracket placeholder plus ≥5 enumerated lines. It cannot see the other shapes in Urkunden IV — the p.118 glyph loop, the p.182 repetition — nor prose fabrication with no enumeration, nor Latin-script books. The corpus-wide B1 rate remains unmeasured; this bounds one signature, not the class.

JDerekLomas:
## Hand-review complete: 6 more fabrications confirmed and withdrawn

Per Derek's request, hand-read all 122 candidates from the v3 detector (OCR + full translation, not just the flagged snippet — several needed the raw uncollapsed text to see the actual invented span).

**6 confirmed genuine fabrications, withdrawn and CDN-purged:**
- Book of the Dead Vol. 2 — pp.19, 54
- Book of the Dead Vol. 3 — pp.220, 288
- Sefer ha-Zohar — p.415 (worth noting: the translation's own `<meta>` tag admitted it — *"this translation reconstructs the likely thematic content based on the provided vocabulary"*)
- Babylonian Liturgies — p.333

Worst case: BD Vol.3 p.220 had 15 fully invented numbered "verses" for a page where OCR declined every single line.

**~91 false positives** — mostly trivial declines (one illegible word/seal amid an otherwise fully-transcribed page) or correctly-hedged full declines describing physical damage rather than claiming content (the Herculaneum papyri and most of the Zohar cluster are exemplary here).

**2 left unwithdrawn as low-confidence** (Maḥzor pp.192, 251 — standard fixed liturgy / a garbled but plausible attempt at a genuinely difficult piyyut).

**A ~25-book Tibetan tantric cluster deliberately left out of this batch** — it corroborates the already-tracked #4523 finding (model can't reliably read cursive dbu-med) rather than being a fresh instance of this bug; recommend folding it into that issue's remediation instead of treating it as 25 more one-off withdrawals here.

**Measured detector precision: 6/122 ≈ 5%.** Confirms this detector is a work list, never an authority — any future run needs the same hand-review pass before withdrawing anything.

Full writeup: `.claude/handoffs/2026-09-03-harkhuf-and-the-fabrication-class.md`.

Issue #2107, opened 2026-05-27
Human-in-the-loop review site for image extraction + curation tasks


JDerekLomas:
## Goal

A volunteer-friendly review site (provisionally `/review` or `/volunteer`) where humans rate samples from the image-extraction + curation pipeline. Outputs are training signals for filter rules and a quality dashboard.

Builds on the rating prototype at `scripts/output/_tmp-skip-filter-rate-v2.mjs` (local HTML page with thumbs up/down + keyboard nav + localStorage). Productionizes that pattern as multiple queues backed by Supabase logging + consensus aggregation.

## Why

The image-extraction pipeline has several silent-failure modes that pure AI cannot catch on its own:
- OCR hallucinates `<image-desc>` tags on bleedthrough/stained pages (issue #2092 spot-check)
- Page-typer mislabels engraved frontispieces as `title-page`
- `gallery_quality` scoring systematically downgrades frontispieces in `defaults.ts:202`
- dHash deduplication can collapse distinct editions of the same emblem
- Many quality decisions (cover image, collection fit) are inherently editorial

Volunteers (and the Derek + other core curators) can resolve these in seconds per item — much faster than designing perfect heuristics, and much cheaper than running another model pass.

### MVP: 1. OCR hallucination check
- Sample: pages with `<image-desc>` markup that the skip filter would currently KEEP
- Question: \"Does the OCR description match something actually visible on this page?\"
- Rating: ✅ matches / ❌ hallucination / ⚠ partial / 🤷 unclear
- Output: per-page hallucination flag → extends the skip filter

### 2. Page-type / frontispiece check
- Sample: pages typed `title-page`, `frontispiece`, `illustration`, `blank` from books with illustrations
- Rating: ✅ correct / 🔄 should be {dropdown} / 🤷 unclear

### 3. Gallery quality scoring
- Sample: random gallery_images biased toward 0.5-0.8 band
- Rating: 👑 hero-worthy / 👍 yes / 👎 no / 🤷 unclear

## Data model

\`\`\`sql
create table volunteer_ratings (
  id uuid primary key default gen_random_uuid(),
  queue text not null,
  item_id text not null,
  rating text not null,
  detail jsonb,
  volunteer_id uuid,
  volunteer_label text,
  created_at timestamptz default now(),
  user_agent text
);
\`\`\`

[...]

PR #4652, opened 2026-09-04
feat(review): report what readers of the original languages found (#2107)

JDerekLomas:
The read side the queue was missing.

A rating store nothing reads is a suggestion box. Three "this is not a page" notes sat unread in this system for a month, and 104 ratings changed nothing in the library because no consumer existed. A queue earns its asking only if the answers surface.

## It does not average the verdicts

The point of the split vocabulary is that `transcription_off` is an **OCR** defect and `translation_drift` is a **translation** defect. Folding them into one score reproduces exactly the confusion the queue exists to remove.

So `faithful_pct` is computed **over pages whose transcription held** — the only population where "is the English faithful?" is a well-posed question — and transcription failures are reported *beside* it, never inside its denominator.

Three numbers, kept apart because they mean different things:

| | |
|---|---|
| `checked` | pages at least one person judged |
| `decided` | pages where enough people agreed to call it (`CONSENSUS_TARGET`) |
| `faithful_pct` | of settled pages whose transcription held, how many read faithfully |

Reporting only `checked` would overstate what we know.

## The display replaces an effort counter with coverage

`/review`'s only feedback was three numbers that can only go up — ratings, notes, yours. An activity count reports effort, not progress: it cannot distinguish a corpus getting better from people clicking more, and it never finishes.

This shows progress toward **35 pages per language** — the sample that buys ±10% at 95% confidence — so a language can visibly be *done*. That boundedness is the thing that makes people start; everyone assumes this work is infinite.

## Implementation notes

- Reads are **paged explicitly**: supabase-js truncates a select at 1,000 rows silently.
- **Note-only rows are excluded** from the tally — a note is a report, not a vote.
- Language is read from the **candidate pool**, not the rating, so a rating can never carry a language the sample did not stratify on.
- `revalidate = 300` — this is a slow-moving number, not a live counter.
- Renders nothing when there is no data, rather than showing a row of zeros.

Refs #2107


JDerekLomas:
Backlog sweep 2026-09-12: brought up to main, CI green (test, field-write-lint, Vercel). **Held for Derek** — this is reader-facing (`/review` summary panel and the `/check/[token]` page carry new authored copy), which is on the hold list. Nothing on main supersedes it; #4650 landed the queue's write side and this remains its read side. Ready to merge on your say-so.

Issue #4162, opened 2026-08-21
Embedding spend is invisible to the budget dial — neither embedder logs to gemini_usage


JDerekLomas:
## The gap

`budgetAllowsDispatch` (`scripts/lib/spend-guard.mjs`) sums today's spend from the two `gemini_usage` stores and stops paid dispatch at `system_config.processing_control.daily_budget_usd`. It is the repo's one brake on Gemini spend.

**Neither embedder writes to `gemini_usage`.** `scripts/workers/embed-gemini.mjs` and `scripts/workers/embed-page-texts.mjs` both call the Gemini embedding endpoint directly and log nothing. So:

- embedding spend is invisible to the dial — the ceiling is enforced against OCR and translation only;
- it is invisible to every cost report built on `gemini_usage`, including the infrastructure-cost docs;
- and it is invisible after the fact, which is how ≈$11.65 of Spanish embedding (#4095) could be spent and then only reconstructed by measuring the corpus and multiplying by the published rate.

The dial does not report that it is partially blind. It just returns a smaller number — the same shape as the incident in its own header comment, where reading one of the two stores showed $9.00 on a day that billed ~$2.3K.

## Scale

`gemini-embedding-2-preview` is $0.20/1M input tokens. Measured 4.29 chars/token, input capped at 8,000 chars/page:

| corpus | per full pass |
|---|---|
| Spanish `page_texts` (38.6K pages) | ≈$5.83 |
| English `page_translations` (3.9M pages) | **≈$180** |

An unlogged $180 job is well past the point where the dial should be able to see it.

## What to do

1. **Log embedding usage.** `embedTexts()` in `scripts/lib/page-embedding-text.mjs` is the single call site for both workers — the natural place. `batchEmbedContents` does not return `usageMetadata`, so the token count has to come from `countTokens` or from the measured chars/token ratio; either is far better than nothing, and `cost_usd` is already documented as a computed estimate rather than billed truth.
2. **Then the gate means something.** #4160 made `embed-page-texts.mjs` consult the dial; once embedding is logged, the dial can actually account for what embedding spends rather than only for what runs alongside it.

## Not urgent, but do not let it go quiet

Nothing is presently running a large embed. The reason to file it rather than leave it is that this class of gap is only ever discovered by someone reconstructing a bill by hand — which is exactly what happened here.

[...]

PR #4658, opened 2026-09-04
fix(meter): every Gemini call in src/ records what it spent (#4599)

JDerekLomas:
Addresses #4599. **Stacked on #4657 — merge that one first.** Its commit is included here, so until it merges the diff below shows both; afterwards this PR collapses to its own commit. (Base is `main` rather than the stacked branch on purpose: the `test` workflow only runs on PRs targeting `main`, and a PR whose checks never ran is not a PR that passed.)

## The measurement

August 2026, both usage stores read against Cloud Monitoring:

```
successful GenerateContent calls at Google ....... 417,936
calls that wrote a gemini_usage row .............. 300,846
calls that wrote nothing ......................... 117,090   (28%)
```

An unlogged call is spend attributable to no workstream, invisible to the daily dial, and — the reason #4581 ran for months — uncheckable for a runaway thinking budget, because there is no row to check.

(The issue's figures, ~250K unlogged / 58% of the bill, came from reading one of the two stores. #4657 has the arithmetic.)

## The fix is structural, not diligent

Adding a log line to forty call sites only works until the forty-first. So `getGeminiClient()` now returns a client whose models log every `generateContent` / `generateContentStream` themselves. Forgetting to log is no longer possible — only forgetting to **label**, which surfaces as `unlabelled` in the attribution table instead of vanishing. Every direct `new GoogleGenerativeAI(...)` in `src/` is gone; the two that remain are inside the chokepoint.

Two ways out, both requiring a stated reason:

- `{ selfMetered: true, reason }` — for lanes that already write a richer row (`src/lib/ai.ts` and the OCR/translation workers, the index routes, word-alignment, quality-scoring…). **A second automatic row would double the measured spend and close the dial on money never spent**; a unit test pins that it doesn't happen.
- `getUnmeteredGeminiClient(key)` — for a contributor's own key, where a row would charge someone else's spend to us (`/api/contribute/*`).

Sites the SDK wrapper can't reach were done by hand: `/api/identify` (two raw REST calls), `identify-rerank`, the podcast script/TTS paths, and two unattended workers (`transliterate-greek`, `ocr-correct-grounded`). In `transliterate-greek` the book id is now passed into the page loop rather than read from `page.book_id` — the projection never selected it, so the row would have carried `undefined`.

`logGeminiCall` also stops connecting to Mongo on the Supabase path (it did so unconditionally, for a title lookup it usually doesn't need). That connection is what made metering too expensive to put on a request-path route.

## A third usage store

`logAiUsage()` writes Mongo **`ai_usage`** for the request-path features — librarian, explain, ai_search_expand, voice. August: 8,784 rows, $77.00, of which the librarian is $76.48. **No spend instrument was reading it.** `spend-reconcile` now prints it on its own line and in its own unit: a librarian row is a *turn*, and an agentic turn is several Gemini calls, so it can never be added to a call count.

## The standing guard

`scripts/audit/gemini-usage-perimeter.mjs`, wired into the existing spend-perimeter workflow. It fails on a new call site that spends without recording, and on any `src/` file constructing the SDK outside the chokepoint. Current state:

```
Gemini call sites scanned .......... 203
write a usage row .................. 68
declared exempt (usage-ok/self) .... 4
write nothing ...................... 131  (131 baselined)
```

The 131 are one-shot maintenance, enrichment, eval and analysis scripts, all hand-run. They are **listed in a baseline, not waved through** — the guard is green today and red on anything new, and a guard that has always been red teaches nothing. Removing a line as you fix it is the intended workflow; adding one is not (declare `// usage-ok: <reason>` at the call site instead).

## What this does not do

- It does not make the rows *correct* — right tokens, right price, right label. `spend-reconcile` answers that against the vendor and `gemini-thinking-and-meter` guards the token definition.
- It does not touch **thinking budgets** on the traffic-driven routes. #4599 is explicit that those want a per-route quality decision, not a blanket zero, and turning thinking off changes output on the creative paths.
- **Embeddings stay unmetered** and say so at the call site: an embedding response carries no `usageMetadata`, so there is nothing to record from. What embedding costs is measured on the writer side (#4162).
- It does not instrument the 131 baselined scripts.

## Cost and behaviour notes for review

- The metered path **awaits** its log (~one Supabase round trip) on a call that already took seconds. An un-awaited write in a serverless function can be killed with the response, and a meter that drops rows under load drops them exactly when spend is highest. `logGeminiCall` swallows its own errors, so metering can never fail a request.
- Streaming logs when the aggregated response resolves, attached to that promise rather than awaited inline (awaiting would buffer the stream).
- Module-level clients in routes are **lazy**: resolving a key at import time would turn a missing `GEMINI_API_KEY` into a build failure, where the old `new GoogleGenerativeAI(key || '')` merely failed at call time.

## Verification

- `npx vitest run` — 259 files, 3,478 tests, green, including 7 new ones on the wrapper (label, thinking tokens counted as output, failure path still throws, unlabelled still logs, streaming, and both opt-outs).
- `npx tsc --noEmit` — clean.
- `scripts/audit/gemini-usage-perimeter.mjs --ci`, `spend-perimeter.mjs --ci`, `gemini-thinking-and-meter.mjs` — all exit 0.

https://claude.ai/code/session_011TYUQFktrtiVxy1V4GRE3W


JDerekLomas:
Backlog sweep 2026-09-12: re-merged main after #4657 landed (the stacked commit collapsed as described). The new perimeter guard found three call sites that arrived on main between Sept 4 and 12 (classify-v5, translation-prompt-ab, transcribe-notation) — hand-run eval scripts of the same class as the baseline, registered there rather than declared exempt. Full unit suite 3,774 green, both perimeter guards exit 0, preview build Ready.

PR #4660, opened 2026-09-04
eval: measure the ground truth itself, and fix the 6% of it we were deleting (#4523)

JDerekLomas:
Track B item 1 of the OCR measurement program. Nothing downstream is quotable until we know how wrong the reference is, so this measures it — and finds that the largest source of reference error was our own cleaner, not Wikisource's volunteers.

## How wrong is the reference?

Wikisource pages carry their own second opinion. `level 3` = one human transcribed the page; `level 4` = a second, **different** human re-read it against the scan and corrected it. `reference-error-rate.mjs` replays each page's revision history and measures exactly what that second reader changed — through the same `normalizeForScript` folding the bench scores through, so the number lands on the bench's own scale.

No hand transcription, and deliberately no VLM re-transcription: refereeing a bench about VLMs with the system under test is not a control.

n=69 independent validations (pinned set + a matched level-4 harvest per wiki), median **0.00%**, 39 references exact:

| | level-3 reference error | level-4 residual | reported engine gap |
|---|---|---|---|
| Greek (n=18) | **0.07%** CI [0.02, 0.13] | 0.00% (5/5 exact) | 1.5pp |
| German (n=15) | **0.06%** CI [0.02, 0.11] | 0.01% | — |
| Latin (n=36) | **1.15%** CI [0.17, 2.47] | 0.48% | 1.4pp |

Greek and German engine differences are ~20x the reference noise and stand. **The Latin comparison does not** — reference error and the reported Kraken-vs-Gemini gap are the same size. "Latin is a tie" turns out to be arithmetic, not a shrug. Latin's distribution is skewed rather than uniformly bad: most references are exact, a handful omit a whole printed block (an apparatus criticus, a clause), so median 0.00% / mean 1.15%.

## The bigger error was ours

`cleanPageText` blanked a template together with the text it wrapped. `{{SperrSchrift|D'Glocke het zwölfi gschlage.}}` is a printed line, not scaffolding. Measured: **6.0% of Greek reference letters, 3.8% of Latin, 0.9% of German** deleted — 5-80x the human reference error — with deeply nested apparatus markup surviving as literal braces on top of that. The Poemander reference was 1,178 characters of `{{κσχασ|εδάφιο=1|σημείωση=μου] μεν A, om Turn. Fluss.}}` soup where the page prints 383 characters of Greek.

**Re-scoring the stored Gemini outputs against corrected references — no re-runs, no API cost:**

| | before | after |
|---|---|---|
| Greek | 97.8% cond / 89.7% uncond @92% | **99.6% / 87.1% @88%** |
| Latin | 96.3% / 81.7% @85% | 96.0% / 81.4% @85% |
| German | 99.4% / 96.0% @97% | unchanged |

One page went **59.7% -> 100.0%**: a perfect transcription had been charged 40 points for our own markup. Coverage falls because a table-of-contents page that the corrupt reference waved through now fails the guard honestly.

**The 1.8pp Greek correction is larger than the 1.5pp Greek engine gap it was used to judge.** Every Bench 2 Greek arm must be rescored before that result is quoted again. The Kraken ws outputs live in #4651 and were not available to rescore here.

## Cleaner rules, each from a page that broke

- keep **every** positional argument, not the longest — a table-of-contents line is nothing but those (one such reference had been cut from 2,000 chars to a 23-char heading)
- emit a sort-key/display pair once
- take the fragment **this** page prints from `{{hws}}`, not the whole word
- named parameters are configuration and drop; numbered ones (`|2=`) are content and stay
- `<ref>` and `{{CRef}}` are treated identically — a validator moving a footnote between the two read as a 13.6% "human correction" and was the largest delta in the first run

Footnote text stays excluded, deliberately and now measurably: it is printed on 9 of 10 footnoted pages (88.6% recall in the Gemini arm against a 97.9% body control), but the tenth is a modern editorial gloss with no syntactic tell, and a reference asserting unprinted text is charged to the **engine**.

## Scope

In: the measurement, the cleaner fix, the pinned references refreshed in place, the level-4 expansion samples, 8 unit tests, an EXPERIMENTS.md entry.

Out, deliberately: matched-k re-runs (item 2), applying Holm in the reporting path (item 3, currently a printed warning with a hardcoded "9 comparisons" that can drift), the escalation threshold sweep (item 5), and rescoring the specialist arms — which needs #4651's outputs and is the first thing to do after this merges.

## Verification

- 18 unit tests pass (`tests/unit/wikisource-text.test.ts`)
- `check-imports` clean
- Every outlier in the first run was traced to a specific page and read in raw wikitext before being kept or dismissed; two of the three largest "human corrections" turned out to be cleaner artifacts and are now regression tests.

https://claude.ai/code/session_014WmTDpvthUQFPnPnVLPC56


JDerekLomas:
## Update: the Greek open question, answered — and one implication in the description walked back

I collected the Kraken ws arm (it had finished on Hetzner; it is already in #4651 as `kraken-ws`, 100 rows, so nothing is duplicated here) and rescored **both** arms against the corrected references.

**Per-arm means, ws tier:**

| | old refs | new refs |
|---|---|---|
| Gemini Greek | 97.8 cond / 89.7 uncond @92% | **99.6 / 87.1 @88%** |
| Kraken Greek | 99.3 / 91.4 @92% | 98.9 / 91.0 @92% |
| Gemini Latin | 96.3 / 81.7 @85% | 96.0 / 81.4 @85% |
| Kraken Latin | 97.7 / 82.5 @84% | 97.6 / 82.4 @84% |
| Gemini German | 99.4 / 96.0 @97% | unchanged |
| Kraken German | 94.7 / 44.2 @47% | 94.8 / 44.2 @47% |

Read alone, that looks like the Greek ranking reverses. **It does not, and the description's "must be recomputed before it is quoted again" is right for the absolute numbers but overstated as a threat to the ranking.** Paired over pages both arms attempted and both aligned under both reference versions:

| | paired n | kraken − gemini, old refs | new refs |
|---|---|---|---|
| Greek | 20 | −0.37pp (kraken ahead on 3/20) | −0.74pp (kraken ahead on 2/20) |
| Latin | 25 | +1.06pp (11/25) | +1.08pp (11/25) |
| German | 13 | −4.64pp (1/13) | −4.63pp (1/13) |

Two things follow, and the second is the more useful one:

1. **Reference error largely cancels in a paired arm-vs-arm comparison** — both arms are scored against the same reference, so a corrupted one penalises both. It does *not* cancel for absolute claims: "Gemini reads Greek at 97.8%" was wrong by 1.8pp; it reads at 99.6%.
2. **The "Greek: Kraken 99.3 vs Gemini 97.8" comparison was unpaired, and it reverses under pairing — with the old references, not because of the fix.** The two arms were averaged over different page sets. Paired, Gemini was already ahead by 0.37pp and Kraken won only 3 of 20 pages. This is the same defect as the coverage-conditioned comparison the bench already documents, one level down: matched coverage is not the same as matched pages.

So the Greek specialist result should not be quoted in either direction from per-arm means. The paired numbers above are the ones that survive, and on them Greek is a Gemini win by well under a point — inside the 1.4pp band, not outside it.

(The `kraken-ws` rows live in #4651 and are not committed here; these numbers were computed against that branch's copy of `scorecard-outputs-2026-09-04.jsonl` and should be re-verified once it merges.)

JDerekLomas:
Follow-up stacked on this branch: **#4665** — Track C, the reference-free fabrication detector. Merge this one first; #4665's base is this branch and it will retarget to `main` automatically.

JDerekLomas:
Backlog sweep 2026-09-12: merged main; the only conflict was the generated `scripts/eval/INDEX.md`, regenerated with `build-index.mjs`. The ground-truth references auto-merged against Bench 2 (#4632). `wikisource-text.test.ts` (18) passes. This branch also carries #4665 (fabrication detector), which was merged into it earlier today. Merging — nothing on main supersedes the reference-error measurement or the cleaner fix.

PR #4661, opened 2026-09-04
fix(pipeline): "complete" meant "has some OCR", so 13,329 books stopped at page 25

JDerekLomas:
Found while asking why 20 books in the Forum of Conscience shortlist were unreadable. They weren't skipped for cost — the pipeline had marked them finished at page 25.

## The bug

Phase 9 asked whether a book had *enough* OCR to look real: anything above **10%** coverage was stamped `complete`. The preview pass transcribes the first **25 pages**. A 250-page book therefore sat at exactly 10%, cleared the floor, and finalized as done — after which it was invisible to every later pass, because the OCR queue only reads `archive_complete`.

## The scale

| | |
|---|---|
| Books stamped `complete` with ≤30 of >50 pages OCR'd | **13,329** |
| …visible to readers | **13,318** |
| Pages they hold | 1,870,238 |
| Pages transcribed | 322,861 (17.3%) |
| **Pages never read** | **1,547,377** |

**Control** — because a number that large is usually a definition problem, not a bug: 11,387 `complete` books have ≥90% OCR. The status is normally honest. This is a distinct failure, not how the field is always written.

The reader-facing cost is worse than the count. A book showing 25 of 250 pages reads as a *curatorial choice* — "we hold it, we didn't translate it" — rather than a queue the pipeline dropped. That's why it stayed invisible: it never looked broken.

## The fix

`scripts/lib/finalize-decision.mjs` — `decideFinalize()`, pure and therefore testable. The logic lived inline in a 5,400-line worker where it could not be.

**Two thresholds, deliberately not one:**
- **≥90% → complete.** The canonical readable bar already used by `homepage_stats` — not a second invented one.
- **<50% → back to `archive_complete`**, the state the OCR queue actually reads.
- **50–90% → complete, on purpose.** Those books have usually hit pages that will never transcribe (damage, blanks, RECITATION blocks). Requeuing them forever would churn the queue and park thousands of books that are as finished as they will ever be. The gap between the thresholds is the point.

**Bounded**, because a requeue that never progresses is an infinite loop: a book is requeued only while its OCR count is still growing. When a lap adds no pages it parks for a human, with both numbers in the message.

The first test is the exact shape of the bug — 250 pages, 25 transcribed — so reintroducing a percentage floor fails CI. The near-complete band and the artwork/empty-book cases are pinned too, since those are what a careless fix would regress.

**Full suite: 3,620 tests passing.**

## Not done here, deliberately

**The 13,329 existing books.** Finalize only revisits `cover_selected`, so they stay exactly where they are. Moving them back to `archive_complete` would queue roughly **$8,000** of OCR and translation behind whatever valve opens next — that is actuation two hops upstream of the spend, and it's Derek's decision, not a side effect of a bug fix.

The count and the query that produced it are in the lib header, so that call can be made with the real number in front of it.

https://claude.ai/code/session_0116Jc7Z7ZzYmAhCPKQNLJpC


JDerekLomas:
Merging as part of the 2026-09-12 backlog sweep, after bringing the branch up to main.

**Conflict resolution:** main's #4563/#4565 fix added an untranslated-finalize requeue that also counts in `pipeline_auto.finalize_requeues`. The two loops now use separate counters (`finalize_ocr_requeues` for this PR's OCR-incomplete loop) so neither consumes the other's retry budget. `finalize-decision.test.ts` (12 tests) and the full unit suite pass.

**What merging actuates:** nothing immediately. Phase 9 only revisits books in `cover_selected`, so the 13,329 books already stamped `complete` at ~25 pages stay where they are — moving them is the ~$8K decision this PR leaves to Derek. From the Hetzner box's next hourly pull of main (:17), a book that reaches `cover_selected` with <50% OCR is returned to `archive_complete` (bounded by the stall check) instead of being stamped complete.

Issue #3958, opened 2026-08-12
1,520 live books carry language_review: true and nothing consumes the queue


Mayank-PPL:
Split out of #3942 (fixed in #3956).

## The gap

**1,520 live books** (`visible: true && pages_count > 0`, measured 2026-08-12) carry `language_review: true` and nothing consumes that queue. The flag is set by `resolveLanguage()` on any caller-vs-source conflict at import (#2185) and by `scripts/maintenance/detect-language-mislabels.mjs`, and it is documented as meaning "this record must NOT auto-publish pending human review." All 1,520 are published.

## Why it matters — this is not hypothetical

One of the two records reported in #3942 was **already in this queue with the correct answer attached**:

```json
// 69b418ae2b0edf3eaa2dd9aa — "Hebrew Kabbalistic Collection"
language: "German",
language_review: true,
language_review_detail: {
  "detected": "Hebrew", "current": "German",
  "confidence": "high", "bucket": "lang_drift",
  "flagged_at": "2026-08-09T06:30:22.223Z"
}
```

The detector found it on **9 August**, named the right language with high confidence, and the book stayed catalogued German until a reader hit it through MCP and filed #3942 on **12 August**. The detection worked; the queue had no consumer. A flag nobody reads is indistinguishable from no flag at all — and worse, it makes the corpus *look* instrumented.

## What is wanted

- A triage surface for `language_review: true` — an admin list, or a batch review script in the shape of `scripts/audit/ft-demote-queue-screen.mjs` (screen cheaply first, review what survives).
- Cheapest high-confidence lane: where `language_review_detail.confidence === 'high'` **and** the OCR model's own per-page `<language>` declarations corroborate the detected value on a mid-book sample, that is two independent signals against one catalogue string. On the Hebrew MS that reads 328/422 pages Hebrew against 1 German — not a close call.
- Resolution must clear the flag (`language_review: false` plus a `field_provenance.language` entry), and **must bump `updated_at`**, or `sync-books-catalog.mjs` never mirrors the correction to `books_catalog` (that defect is written up in #3956).
- Sample the queue before building: how many of the 1,520 are real drift versus multilingual volumes where a single scalar was always going to be wrong?

## Prior art in-repo

[...]

PR #4696, opened 2026-09-10
feat(language): set books.language from page text, sampled across the interior

JDerekLomas:
## Why

`books.language` picks the OCR model. `getModelForBook()` (`src/lib/types/ai-models.ts`) routes Latin-script languages to flash-lite and non-Latin/unknown to full flash — so a book mis-catalogued as English buys the cheap Latin-script model for its entire run.

Nothing in the repo sets `language` from **our own** page text. `src/app/api/import/ia/route.ts` can only pass IA's `ocr_detected_lang`, which IA publishes for roughly a third of items, and the direct importers don't read even that.

## The measurement that prompted it

Audited the 124 books of the 2026-09-09 acquisition wave. IA carries `ocr_detected_lang` for **42**; 37 agreed with what we'd stored and **5 were wrong**, the OCR right every time:

| stored | actually | book |
|---|---|---|
| English | German | Novalis, *Fragmente* |
| English | French | La Bruyère, *Les Caractères* |
| Italian | French | Pleyte & Rossi, *Papyrus de Turin* |
| English | Latin | *Caeremoniale Episcoporum* |
| English | Greek | *Greek Papyri in the British Museum* |

~12% error in catalogued language wherever it is checkable. The remaining **82 had no page-derived signal at all**.

## What's here

- **`scripts/lib/page-language.mjs`** — generalises the sampling / `<language>`-tag / script-census primitives from `scripts/audit/ft-english-badged-classify.mjs`. That script answers only English-vs-not for badge adjudication and writes nothing; it is load-bearing there and is left untouched, keeping its own copy. Normalisation delegates to `language-normalize.mjs` (the pinned twin with a parity test) rather than a fresh code map — an ad-hoc one that knew 3-letter codes but not 2-letter ones produced 30 false "conflicts" in the audit above before anyone checked.
- **`scripts/maintenance/detect-language-from-pages.mjs`** — two phases, so the paid step is a deliberate separate command. `--plan` picks the sample, writes a page-id list and prices it; the caller then runs `realtime-ocr.mjs --page-ids-file=…` (OCR is **not** reimplemented); `--apply` reads the OCR and writes `language` plus typed provenance.

## Two traps encoded on purpose

**The front of a book lies.** The sample skips ~15% front and ~5% back (#3524). A scholarly edition of a Latin text opens with an English title page, preface and introduction: the first dozen content pages of Quignones' *Breviarium Romanum*, Feltoe's *Sacramentarium Leonianum* and Little's *Opus Tertium* all read English while pages 112, 113 and 63 are solidly Latin. Sampling the first N content pages measures the **apparatus**; it nearly demoted three legitimate first-translation badges. Verified on this wave: every sample runs 15%→92% of its book, zero sampled only in the front.

**Script names a writing system, not always a language.** A heavy non-Latin census vetoes a Latin-script tag, but script is only allowed to *name* the language where it can (Greek, Hebrew, Syriac, Armenian, Tibetan). Devanagari and CJK go to review rather than guess between Sanskrit/Hindi or Chinese/Japanese.

## Write discipline

A `clear` verdict writes. A `review` verdict **never** overwrites a stored value — it records both claims and sets `language_review`, matching the conflict contract in `src/lib/resolve-language.ts`. `pages.ocr` is read via `ocr.data` throughout (it is an object; `String()` yields 15 chars and silently fails every length filter).

## Scope / out of scope

In: the library, the maintenance script, and the `--plan` costing. Out: changing the direct importers to read `ocr_detected_lang` (they should, and it is the obvious follow-up — kept separate so this PR stays reviewable), and the bare-string-vs-typed `field_provenance.language` inconsistency between the direct importers and the routes, which this script reads both sides of but does not migrate.

## Test

`--plan` over the 82 books: 2,050 pages, $4.55 at the measured rate, spread verified. `--apply` results to follow in a comment once the sample OCR completes.

https://claude.ai/code/session_0121EoJp3hd4FGrsQgYcoi2m

### DIFF
--- a/scripts/import/founding-tail-direct.mjs
+++ b/scripts/import/founding-tail-direct.mjs
@@ -17,6 +17,7 @@ import { makeBookDoc, makePageDoc } from '../lib/book-docs.mjs';
 // the `born|died|fl|circa|ca` date strip — and is left in place deliberately:
 // swapping it would change this script's dedup keys. See #4444 for the plan.
 import { normalizeTitle } from '../lib/dedup-normalize.mjs';
+import { resolveIaLanguage } from '../lib/ia-language.mjs';
 const COMMIT = process.argv.includes('--commit');
 
 const SETS = [
@@ -82,11 +83,16 @@ async function main(){
     for(const k of Object.keys(iaCatalog)){const v=iaCatalog[k];if(v===null||v===undefined||(Array.isArray(v)&&v.length===0))delete iaCatalog[k];}
     const licenseUrl=meta.licenseurl||meta.license||null,rights=meta.rights||meta.possible_copyright_status||null;
     const contributor=typeof meta.contributor==='string'?stripText(meta.contributor):null,sponsor=typeof meta.sponsor==='string'?stripText(meta.sponsor):null;
+    // Language: IA's own signals outrank the hardcoded value in SETS above. This script previously
+    // wrote `language: b.language` with `field_provenance:{language:'caller'}` — literally accurate,
+    // and the defect: a caller value describing the WORK silently overrode the source's record of
+    // what this PRINTING is (#2184). resolveIaLanguage applies the route's precedence.
+    const lang=resolveIaLanguage(meta,b.language);
     const bookId=new ObjectId(),bookIdStr=bookId.toHexString();
     const slug=await uniqueSlug(db,slugify(`${b.title} ${b.author}`));const now=new Date();
     const photo=i=>`https://archive.org/download/${b.ia}/page/n${i}/full/full/0/default.jpg`;
     const thumb=i=>`https://archive.org/download/${b.ia}/page/n${i}/full/pct:15/0/default.jpg`;
-    const bookDoc=makeBookDoc({_id:bookId,id:bookIdStr,slug,title:b.title,display_title:null,author:b.author,language:b.language,published:b.published,field_provenance:{language:'caller'},categories:['History','Political Philosophy'],ia_identifier:b.ia,thumbnail:thumb(0),pages_count:pageCount,pages_ocr:0,pages_translated:0,content_type:'book',text_role:'original',
+    const bookDoc=makeBookDoc({_id:bookId,id:bookIdStr,slug,title:b.title,display_title:null,author:b.author,language:lang.language,...(lang.original_language?{original_language:lang.original_language}:{}),published:b.published,field_provenance:{language:lang.provenance},...(lang.conflict?{language_review:true}:{}),categories:['History','Political Philosophy'],ia_identifier:b.ia,thumbnail:thumb(0),pages_count:pageCount,pages_ocr:0,pages_translated:0,content_type:'book',text_role:'original',
       dublin_core:{dc_identifier:[`IA:${b.ia}`,...(iaCatalog.ark?[String(iaCatalog.ark)]:[]),...(iaCatalog.oclc_id?[`OCLC:${iaCatalog.oclc_id}`]:[]),...(iaCatalog.lccn?[`LCCN:${iaCatalog.lccn}`]:[])],dc_source:`https://archive.org/details/${b.ia}`,...(iaCatalog.publisher?{dc_publisher:iaCatalog.publisher}:{}),...(iaCatalog.description?{dc_description:iaCatalog.description}:{}),...(iaCatalog.subjects?.length?{dc_subject:iaCatalog.subjects}:{})},
       catalog_metadata:iaCatalog,...(iaCatalog.place?{place_published:String(iaCatalog.place)}:{}),...(iaCatalog.publisher?{publisher:String(iaCatalog.publisher)}:{}),
       image_source:{provider:'internet_archive',provider_name:'Internet Archive',source_url:`https://archive.org/details/${b.ia}`,identifier:b.ia,license:licenseUrl||'publicdomain',license_url:licenseUrl,rights,...(contributor?{contributing_library:contributor}:{}),...(sponsor?{sponsor}:{}),access_date:now},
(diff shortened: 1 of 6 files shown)


JDerekLomas:
## `--apply` results on the 82 books

Sample OCR: 2,050 pages, 1,914 ok / 9 err / 127 skipped (120 recitation refusals, 6 hallucination, 1 empty), 3,382s, $4.55 at the measured rate. All 82 books now carry `field_provenance.language.chosen_from: 'page_ocr_detected'`; none is left without a language.

**2 written · 66 confirmed · 14 flagged for review · 0 too thin**

### Written
| book | was | is | evidence |
|---|---|---|---|
| Ratio studiorum (Pachtler) | German | **Latin** | modal "latin" 21/21 content pages |
| Garbe, *Vaitāna Sūtra* | English | **German** | modal "german" 23/23 content pages |

### The run changed the design — four corrections, all in the follow-up commit

**A same-family refinement is not a correction.** Der Renner's pages read "Middle High German" against a stored "German"; *La Règle du Temple* reads "Old French" against "French". Writing those would also have pushed both books off `getModelForBook()`'s Latin-script list and onto the **dearer** model — a silent cost increase dressed as a metadata fix. `languageFamily()` already existed for exactly this; they now resolve as confirmation with the finer label kept in provenance.

**A parallel-text edition has two languages, and picking one is editorial.** A runner-up at ≥25% of the sample now goes to review. It found real ones:

- Della Casa, *Galateo* — Italian 62% / Latin 38% (the title says *volgare e latino*)
- Budge, *Bar Hebraeus* — Syriac 59% / English 29%
- Fossey, *La magie assyrienne* — French 70% / Akkadian 30%
- *Statuten des Deutschen Ordens* — Latin 40% / German 36%
- Gorfinkle, *Maimonides* — English 68% / Hebrew 32%
- Warren, *Celtic Church* — English 67% / Latin 33%

**Agreement must not outrank a review verdict.** Gorfinkle's Maimonides *vanished* from the report: the bilingual verdict returns the modal language, that equalled the stored value, and the `agrees` check short-circuited first — logging "confirmed English" for a Hebrew/English parallel edition.

**A token must read as a language name.** The OCR emitted `|xam` for *Specimens of Bushman Folklore* — a real language (ǀXam) under a transcription no consumer knows.

### Prior art found late, now disclosed in the header

`scripts/audit/detect-book-languages.mjs` (#4117) already aggregates the per-page tag over **all** tagged pages with its own bilingual threshold (default 0.10), and `scripts/audit/language-review-triage.mjs` (#3958) triages the review queue. **Both deliberately never write** — "clearing a flag on a published book is a public metadata decision." Two consequences:

- `language` is no longer written over a `visible` book without `--allow-visible`. This script's justification for writing at all is the case those two cannot reach: a book with no OCR yet, so no tag to aggregate, which is why it commissions a sample first.
- Flags now carry `language_review_detail` in the shape the triage script reads, so they land **in** that existing queue rather than beside it. Worth knowing: that queue holds ~1,519 live books, nothing drains it, and its default query is `visible: true` — so a flag on a hidden book is only reachable via its `--all`.

### Also worth knowing

The first OCR attempt spent **$0**: Atlas DNS dropped out and `realtime-ocr.mjs` stopped itself after 20 consecutive failures rather than grinding through 2,000 pages of errors.

Spread sampling verified on the real data: every sample runs 15%→92% of its book, zero sampled only in front matter.

JDerekLomas:
## Follow-up: the direct importers now read IA's detected language

The gap turned out to be wider than "they used the weaker signal". **They read no source signal at all.** `founding-tail-direct.mjs` wrote `language: b.language` with `field_provenance: {language: 'caller'}` — literally accurate, and the defect. Of the 17 scripts that fetch `archive.org/metadata`, **none** read its language:

| script | how it sets language |
|---|---|
| founding-tail-direct.mjs | `s.lang` (hardcoded) |
| founders-collected-direct.mjs | `s.lang` |
| jefferson-canon-direct.mjs | `s.lang` |
| founding-influences2-direct.mjs | `s.lang` |
| bncf-aldine-direct.mjs | `entry.original_language \|\| 'Latin'` |
| early-america-direct.mjs | `'English'` |
| new-world-wave.mjs | `b.language` |

Meanwhile the route has flipped that precedence since #2184, because a caller value describing the *work* silently overrode IA's record of what a given *printing* is.

### Added

- **`scripts/lib/ia-language.mjs`** — IA's two signals in priority order (`ocr_detected_lang` above `metadata.language`, both above the caller), normalised through the existing pinned twin rather than a fresh code map. On disagreement the source takes `language`, the caller is demoted to `original_language` as a work hint, and `conflict` is set.
- **`tests/unit/ia-language.test.ts`** — pins the precedence, using the real disagreements measured on this wave as fixtures (10 tests).
- **`founding-tail-direct.mjs`** — the de facto IA template, now using it.

### Two judgement calls worth reviewing

**Not a twin of `resolve-language.ts`.** That would be the complete fix, but it depends on `displayLanguage` from `src/lib/language-utils.ts`, which **disagrees** with `scripts/lib/language-normalize.mjs` on period-qualified names — `displayLanguage` strips the qualifier ("Middle High German" → "High german"), `normalizeLanguageToken` keeps it. A twin has to settle that first, and that is a separate question from reading the field. Flagging it rather than quietly picking one.

**`language_review` added to the `makeBookDoc` whitelist.** The field-sprawl guard (#3969) blocked it and was right to ask: it is a real field — ~1,519 live books carry it, a Sunday cron refills the queue — that the route has written at import since #2185 and the direct path could not. So the same conflict was being silently dropped depending on which door a book came through.

### Verification

Passed a deliberately wrong caller language — `"Latin"` for Fowler's *Roman Festivals* — and IA's signal won: the book resolved to **English**. 262 test files / 3,581 tests pass. `founding-tail-direct.mjs --dry-run` still clean (52 already-held, 0 failed).

### Still open, deliberately

The other 14 one-off importers in the table above are untouched: most are completed waves that will not run again, and editing them changes what a re-run would store. The helper is the documented path for the next one. Happy to sweep them in a follow-up if you'd rather they were consistent now.

Issue #4737, opened 2026-09-11
Mac IIIF archive daemon: e-rara answers 403 to the laptop, breaker trips every cycle, 0 pages archived


JDerekLomas:
Found 2026-09-11 while sizing a laptop archive lane for MDZ (#4404).

**Tell:** `/tmp/archive-iiif-loop.log` on Derek's Mac ends every cycle with `[CIRCUIT BREAKER] 25 consecutive failures — stopping` and `Done in 94.9s: 0 archived (0 MB), 30 failed`. The failures are all `HTTP 403` and the books are e-rara (`www.e-rara.ch`) carried under `image_source.provider: 'iiif'` — e.g. `6a94053b1cd9a117d3088f35`, `6a906edd32545072610bd687`. So `org.sourcelibrary.archive-iiif` (`scripts/workers/archive-iiif-daemon.sh`, `--providers=iiif`) is doing nothing useful and the e-rara `iiif` lane is effectively parked.

**What is known:** the Mac's egress is Cloudflare WARP (`warp=on`, colo LAX). e-rara is the host the Mac lane exists for (it 429s datacenter IPs), so if it now 403s the WARP address too, the lane has no working egress. MDZ served the same address cleanly at the same moment (10 concurrent full-res fetches, all 200 in 3–4 s), so this is host-specific.

**Not done:** no retry, no user-agent change (CLAUDE.md: never route around a block). Needs someone to check e-rara from a non-WARP residential address and, if the block holds, to contact e-rara — the same shape as the Wellcome 403 (#4311).

Related: `org.sourcelibrary.archive-erara` (the dedicated e-rara launchd job) shows no pid in `launchctl list`; whether it is also failing was not checked.

Issue #4716, opened 2026-09-10
Atlas gap acquisition wave: 494 canonical works, hidden import + full-res archive, no OCR


JDerekLomas:
## What
The theme atlas (https://sourcelibrary-atlas.vercel.app/next) lists 494 canonical works verified absent from the library (re-checked against the local catalogue + page index 2026-09-10, 494 of ~560 candidates confirmed absent; the rest were held after all and folded back into the maps).

This wave puts them through the **canonical acquisition route** (`.claude/docs/import-workflow.md`) with **no OCR or translation** — Derek decides that separately. Imports are hidden (`visible:false, hidden:true`), categorised `atlas-map`, and archived full-res to R2 by the existing Hetzner archiver.

## Route
- **Tier 1/2 (140 works, pre-resolved pre-1930 IA items):** laptop script `sourcelibrary-atlas/scripts/map-pipeline/acquire.mjs --only-direct`. Per work: IA metadata check → rights screen (scan ≥1930 and not PD → skip; imagecount <12 → skip, PDF-only uploads) → pre-check on `ia_identifier` / `dublin_core.dc_identifier` / `source_fingerprint` → `POST /api/import/ia` (Bearer CRON_SECRET; the route runs `acquisitionGate` + `checkDuplicate`) → `acquisition_queue` row `status:'acquired'` so `archive-acquired.ts` (Hetzner :45) archives full-res. Failed direct attempts fall through to the queue.
- **Tier 3/4 + fallbacks (~350 works):** `acquisition_queue` rows `status:'pending'`, `category:'atlas-map'`, `priority:5`; the hourly Hetzner `acquire-gap-batch.mjs` resolves (IA/MDZ/Gallica/e-rara/OAI), LLM-verifies, imports hidden, archives. Unresolvable rows end `no-source` / `no-match`, which is the record. Works whose author could not be parsed from the canonical name land as `status:'no-author'` and need a human pass.
- Dry-run log: `scratchpad/gaps/acq-dryrun-direct.jsonl` (83 would-import, 45 ia-no-images, 11 ia-error, 1 rights-unknown). Live log: `scratchpad/gaps/acq-results.jsonl`.

## Invariants honoured
- Bulk stays off Vercel functions except this one-time ~85-request tier-1/2 pass; the remainder runs on Hetzner via the existing cron.
- Dedup: fingerprint `ia:<id>` + edition-aware title/author gate in the route; queue upsert by `sn`.
[...]

Issue #4404, opened 2026-08-30
[Epic] Grow the R2 corpus to preservation grade at +$250/mo — sequence, constraints, and why measurement comes first


JDerekLomas:
**Goal (Derek, 2026-08-30):** grow the archived-to-R2 corpus as large as possible, at preservation grade, so the library is ready the moment OCR and translation are unpaused. Budget decision: **+$250/month recurring R2** (~17.9M pages, ~71,500 books at the measured 0.93 MB/page).

Nine issues touch this. They are not independent, and doing them in the wrong order wastes the budget. This is the sequence and why.

## The constraint, stated plainly

**Supply is not the limit. Speed is not the limit. Money is** — and right now, *measurement* is.

- `import_candidates` holds **3,527,297 rows, 3,436,156 with a usable manifest, 26,911 imported (0.78%)**. At ~250 pages/book the full pool is ~860M pages ≈ 763 TB ≈ **~$11.7K/month**. We could not afford to finish it if we fetched perfectly.
- Measured object size: **947 KB master + 4 KB thumb = 0.93 MB/page**, so **1M pages ≈ 0.89 TB ≈ $14/month recurring, forever**.
- Therefore the drain must be **budget-driven, not drained to exhaustion**, and the budget is denominated in pages — which requires knowing how many pages actually lack a master.

### 0. Unblock — the archiver is livelocked
- **#4395** / **PR #4396** ← *green, mergeable, blocking everything below*

The per-host limiter released every waiter at once: 60 concurrent callers against a nominal 5/s limit produced **55 requests in one second** (11×). MDZ throttled a rate we never intended to send. Worse, a 429 was counted as a *block*, so each hourly run aborted → slept → retried at the identical rate. Net **~10 books/hour** against a 3,211-book backlog.

Nothing else on this list matters until this merges.

### 1. Measure — you cannot spend a budget against a number three subsystems disagree about
- **#4239** (one metric) — **PR #4403** lands the definition
- **#4190** (`pages_archived` lies in both directions, and `archive-bulk` selects by it)
- **#3186** (~2.1M pages below master resolution)

[...]

PR #4739, opened 2026-09-11
archiver: --newest-first + --hosts, and a laptop lane for MDZ (#4404)

JDerekLomas:
## Why
MDZ (`api.digitale-sammlungen.de`) rate-limits Hetzner per IP: a lone curl from the box got a 429 in 60 ms while the hourly archiver ran, and that lane has converged on ~0.2 pages/s (~21K pages/day) against a 3.8M-page MDZ backlog (10,715 `mdz` + 4,809 `bsb` books un-archived). Derek's Mac is a different egress address: ten concurrent full-res MDZ fetches from it returned 200 in 3–4 s at the same moment. Step 3 of epic #4404 ("calibrate the MDZ rate") has a cheaper cousin: a second address.

## What
Two flags on `archive-acquired.ts`, both off by default, so the Hetzner cron is unchanged:
- `--newest-first` — sort `{archive_attempts: 1, _id: -1}`; the two lanes meet in the middle instead of fetching the same books, and fresh acquisitions (e.g. the atlas wave, #4716) stop waiting behind the June backlog.
- `--hosts a,b` — archive only books whose first un-archived page lives on one of these hosts. The queue's `source` label is not the host (60 newest "iiif"/"mdz" rows: 25 MDZ, 17 IA, 13 Gallica, 4 e-rara, 1 SBB), and each host has its own lane and its own blocks (e-rara 403s the Mac, #4737; IA times out through its tunnel). Skipped rows are left untouched (no attempt, no note).

`archive-acquired-mac.sh` + `org.sourcelibrary.archive-acquired-mac.plist`: a launchd KeepAlive loop for the lane — checks `~/bin/bulk-net-guard.sh` every cycle (fails open), bounds each cycle at 50 min (gtimeout or perl `alarm`, macOS has no `timeout`), parks 6 h on exit 3 (HostBlocked), MDZ only by default. Rate stays at the in-process `DOMAIN_LIMITS` 2/s.

## Measured
Smoke test from the Mac (`--batch 40 --concurrency 2 --page-concurrency 4 --newest-first --hosts api.digitale-sammlungen.de`, 150 s): **209 pages ok, 0 failed, 1.9 pages/s**, all MDZ. Hetzner's lane the same hour: 667 pages in 50 min (0.22 pages/s).

Both lanes are idempotent per page (`archiveIiif` only fetches pages lacking `archived_photo`), so an overlap costs a duplicate fetch, never a bad write. `npx tsc --noEmit` clean.

## Out of scope
- Raising MDZ's per-process rate above 2/s: the comment in `iiif-utils.mjs` says raise only with evidence from a clean run; this PR produces that evidence, it does not act on it.
- The e-rara 403 on the Mac (#4737) and the BSB rate-allowance email (#4404 step 2).
- The launchd agent is installed on Derek's Mac pointing at the worktree checkout until this merges; after merge, re-point the plist's script path to `~/sourcelibrary/scripts/catalog-coverage/archive-acquired-mac.sh` and `launchctl kickstart -k gui/$(id -u)/org.sourcelibrary.archive-acquired-mac`.

### DIFF
--- a/.claude/handoffs/2026-09-11-laptop-archive-lane.md
+++ b/.claude/handoffs/2026-09-11-laptop-archive-lane.md
@@ -1,0 +1,54 @@
+# Laptop archive lane for MDZ, and the network guard that could not see the network — 2026-09-11
+
+PR: https://github.com/Embassy-of-the-Free-Mind/sourcelibrary-v2/pull/4739 (open, green, review-gated).
+Epic: #4404 (R2 growth). Side issue: #4737 (e-rara 403s the Mac; the Mac IIIF daemon archives nothing).
+
+## What was built
+- `scripts/catalog-coverage/archive-acquired.ts`: `--newest-first` (sort `_id:-1`, so a second lane meets
+  the Hetzner lane in the middle and fresh acquisitions stop waiting behind the June backlog) and
+  `--hosts a,b` (archive only books whose first un-archived page lives on those hosts; the queue's
+  `source` label is not the host — "iiif" rows point at MDZ, IA, Gallica and e-rara). Both off by default.
+- `scripts/catalog-coverage/archive-acquired-mac.sh` + `org.sourcelibrary.archive-acquired-mac.plist`:
+  launchd KeepAlive loop, MDZ only, bounded 50 min per cycle (perl `alarm` — macOS has no `timeout`),
+  falls back to the main checkout's `.env.production.local` from a worktree, re-checks the network guard
+  every minute and SIGTERMs the archiver on DENY (positive-controlled with `--park`), and sets
+  `BULK_NET_ALLOW_ONLY=1` so it runs only on an allow-listed network.
+
+## Why
+MDZ rate-limits per IP: a lone curl from Hetzner got 429 in 60 ms while the hourly archiver ran; that lane
+converges on ~0.2 pages/s (~21K MDZ pages/day) against a 3.85M-page MDZ backlog. Ten concurrent full-res
+fetches from the Mac at the same moment all returned 200 in 3–4 s. Measured lane rate from the Mac at the
+unchanged 2/s in-process cap: 1.8–1.9 pages/s (209 pages / 150 s smoke; 106 pages in the first launchd
+minute), 0 failures.
+
+## The incident inside the session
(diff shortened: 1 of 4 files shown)


JDerekLomas:
Two follow-up commits after the first launchd run: (1) the runner falls back to the main checkout's `.env.production.local` when started from a worktree (the file is gitignored, so the worktree had none — first cycle died on `startsWith` of an undefined URI); (2) the runner re-checks `bulk-net-guard.sh` every minute and SIGTERMs the archiver on DENY — a 50-minute cycle that consults the guard only at its start would keep pulling through in-flight wifi if a plane boards mid-cycle. Positive control: parking the guard while the lane ran stopped the archiver within the minute. The guard's own deny-list (per-machine, `~/bin`) was widened at the same time; that file is not in this repo.

JDerekLomas:
Third follow-up, and a correction to the second: the SSID deny-list cannot fire on this Mac. macOS 26 redacts the network name from `networksetup`, `ipconfig getsummary`, `system_profiler` and `ioreg` without Location permission, so the guard reported "not associated with a Wi-Fi network" on a live link and failed open — the lane ran ~15 min on an airport-lounge network (top saved SSIDs "_LAX Free WiFi" / "AF Lounge", egress colo LAX) before this was caught. Derek's instruction: "only do it on wifido".

The runner now sets `BULK_NET_ALLOW_ONLY=1`, and the guard (per-machine, `~/bin/bulk-net-guard.sh`, not in this repo) gained an allow-list keyed on the network SIGNATURE — default gateway's hardware address + DHCP server — which is readable by anyone and stable for a home router. `--pin wifido` while at home writes it; until then the lane parks (verified: `DENY … no allow-list`, 0 archiver processes). Once the allow-list exists it binds for every caller of the guard, so the older Gallica/e-rara/Harvard launchd jobs inherit the same home-only rule. The tether check (default-route service name) is unaffected and still works.

JDerekLomas:
Fourth and last for tonight: the per-machine guard now **fails closed** on a Wi-Fi network whose name macOS redacts (the "not associated" branch checks `ipconfig getsummary` for `SSID : <redacted>` and denies), so the lounge case cannot recur on this Mac even for callers that don't set `BULK_NET_ALLOW_ONLY`. Negative control run on the lounge network: DENY. Handoff in `.claude/handoffs/2026-09-11-laptop-archive-lane.md` on this branch. Lane stays parked until Derek pins wifido at home.

Issue #3825, opened 2026-08-09
Translation prompt v13: closed tag vocabulary, verbatim original-notes, no glossary lines


JDerekLomas:
Plan for the next revision of the DB-owned "Standard Translation" prompt (currently v12). Grounded in the 2026-08-09 tag census: a 20,000-page random sample per side (translation.data / ocr.data) plus a full consumer sweep of src/ and scripts/ (~60 consumer files). Related: #3811 (reader term+note fix), #3820 (embedding leak, fixed in #3821), #3822 (DTS raw tags).

## Tag inventory (docs-containing % of sample; consumers from the code sweep)

**Layer 1 — transcription (words physically on the page; content must never be deleted):**

| tag | translation | OCR | consumers |
|---|---|---|---|
| header | 24.9% | 59.6% | locus, front-matter, contains-works, misalignment detection |
| page-num | 11.1% | 68.7% | locus reconstruction, drift detection, import splitting |
| margin | 24.7% | 24.9% | reader (chip → unwrap), exports |
| column-break | 16.1% | 16.3% | reader column split, editor, exports |
| sig | 6.8% | 19.5% | metadata extraction only |
| unclear | 3.2% | 4.0% | reader, exports, locus separator |
| insert | 2.4% | 2.6% | reader; ⚠ miscounted as wrapper by collapse detector |
| catchword | ~0 | ~0 | ngram apparatus strip only — effectively dead |

**Layer 2 — translator's apparatus (toggleable):**

| tag | translation | OCR | consumers |
|---|---|---|---|
| note | 61.7% | 3.2% | reader chip, flashcards, exports; 12.2% fabrication rate in "original:" notes (#3308) |
| term | 49.1% | 3.8% | reader chip, book index, flashcards |
| gloss | 36.8% | 4.3% | reader, flashcards, exports |
| image-desc | 1.8% | 18.4% | reader caption, image-extraction gating, grounding |

**Layer 3 — machine housekeeping (candidates to extract to fields at write time, never stored in text):**

| tag | translation | OCR |
|---|---|---|
| vocab | 3.3% | 96.7% |
| page-type | 1.4% | 96.3% |
| language (+lang alias) | 1.2% (+0.03%) | 95.4% (+2.8%) |
| script | 0.3% | 63.9% |
| scan-quality | 0% | 48.8% |
| meta | 34.8% | 40.0% |
| summary / keywords | 58.4% | — |
[...]

PR #4758, opened 2026-09-12
Translation prompt v15: all five #3825 items + pre-registered v13-vs-v15 A/B (unpaid half)

JDerekLomas:
## What this changes

Finishes the **unpaid half** of #3825 and stages the measurement that decides whether v15 becomes the default. Nothing has been translated; the default is still v13.

**1. `scripts/maintenance/translation-prompt-v15-verbatim-original.mjs`** — the merged item-2 seeder now carries all five #3825 items, each an anchored `once()` edit that throws on a miss:

| # | rule | inserted where |
|---|---|---|
| 1 | closed tag vocabulary — use ONLY the defined tags, never invent one; no HTML, no `<greek>`/`<quote>` | before "Metadata tags" |
| 2 | verbatim original-notes (**unchanged from #4740**) | before "Image descriptions" |
| 3 | no standalone glossary lines; `<keywords>` is the only end-of-page list | bullet on "Do NOT use" |
| 4 | OCR housekeeping tags (`vocab`/`lang`/`page-type`/`page-num`/`sig`/`scan-quality`) are input, never output | before "Do NOT use" |
| 5 | duplicate 8/9 numbering → 8/9/10/11; style + em-dash rules scoped to all authored prose, with a source-punctuation carve-out | in place |

v13 7,233 chars → v15 11,188. **Applied 2026-09-12 as a non-default row** (`6aa5767fbcba38d6ce451d50`) so the A/B can read it; `--replace` re-seeds if review changes wording.

**2. Consumer-surface check** (a tag-vocabulary change is a contract change — #3825 names four surfaces). What was checked and what was found, also recorded in the script header:

- **NotesRenderer** (`src/components/reader/NotesRenderer.tsx`) — already strips every tag item 4 removes (`lang`/`page-type`/`page-num`/`scan-quality`/`vocab`) and renders every tag item 1 keeps. Its regex for free-text "Vocabulary:" blocks (~line 180) is exactly the shape item 3 forbids; it becomes dead code for new pages, not broken code.
- **build-book-index** (`src/lib/build-book-index.ts`) — reads `<vocab>` from `ocr.data` ONLY, `<term>`/`<keywords>` from `translation.data`. All three survive v15. Not affected by item 4.
- **learn route** (`src/app/api/learn/route.ts`) — flashcards need `<term>X</term>` immediately followed by `<note>`/`<gloss>`. v15 keeps that shape inline. **Finding:** a trailing glossary block also satisfied the adjacency, so item 3 may reduce flashcard *yield* per page. That is why the harness gates on *inline* terms (floor 20%) and reports total terms descriptively.
- **download routes** (`src/app/api/books/[id]/download/route.ts` → `src/lib/export-markdown-html.ts`) — placeholders for note/margin/gloss/term/unclear, strip list for the housekeeping set. **Finding:** an invented tag is in neither list, so today it survives into exported HTML as literal markup. Item 1 is a strict improvement here.
- **`<header>`/`<page-num>` on the translation side** (11.1% of translation docs per the census) feed nothing: locus anchors, front-matter and drift detection all read `ocr.data` (`git grep '<page-num>'` — 20 hits, none on `translation.data`).

**3. `scripts/eval/translation-prompt-ab.mjs`** — the measurement. Primary outcome: **verified-note rate**, from `scripts/lib/page-terms-parse.mjs` (the `build-page-terms.mjs` verifier; not a second implementation). Co-primary: **note emission** — a prompt that stops writing notes has not won. Gates: invented tags, housekeeping leakage, glossary blocks, em-dashes, body length (floor 10%, excluding the two things v15 removes by design), inline terms (floor 20%). Paired, **one page per book**, 8 strata × 40 (script × print/manuscript — the manuscript axis is a declared provider proxy; `books` has no such field, checked). k=1 per (page, arm), argued in the header against `prompt-ab.mjs`'s k≥5: the estimand is a rate over 320 pages, not stability on ten. `--draw` (free) / `--run` (**paid, refuses without `--approved-usd ≥ estimate`**) / `--score` (free) / `--judge-packet` (free, blinded pairs for the Claude regression judge).

**4. `scripts/eval/PREREGISTRATION-translation-prompt-v15.md`** — hypotheses, sample, outcomes and the decision rule fixed before any paid call. **Flip threshold: ≥ 3 percentage points** on the verified-note rate, with the paired CI excluding zero, emission not down >20%, body not down >10%, no gate firing, and the blind 30-pair judge not finding v15 worse at >1.5×.

**5. Pinned draw** `results/translation-prompt-v15-sample.json` — 320 pages from 320 books, 793,641 OCR chars, OCR text included so run and score read the same input.

**6. lib additions, reusable:** `sampling.sampleOnePagePerBook`, `paired-stats.bootstrapRatioCI` (cluster bootstrap on a ratio of per-page counts), `paired-stats.seededRand`.

**7. `tests/unit/translation-prompt-ab-metrics.test.ts`** — every metric has a positive and a negative control (15 tests). `--score` and `--judge-packet` were exercised end-to-end on a synthetic arms file with a known answer.

## Cost — awaiting explicit approval

| run | model | calls | estimate |
|---|---|---|---|
| all on lite | `gemini-3.1-flash-lite` | 640 | **$1.47** |
| `--route` (recommended) | production routing: 480 on `gemini-3-flash-preview`, 160 on lite | 640 | **$2.64** |

`--route` is recommended because the flip ships on that routing; a prompt effect measured on a model six strata never use is a weaker basis.

## Out of scope, deliberately

- **Flipping the default.** Derek's call, after the run; the harness recommends, never writes.
- The paid run and the report. Blocked on spend approval.
- Dropping `heading`/`footnote`/`caption` from the write-time repairer (`sanitizeTranslationTags`) — #3825 says "once v13 is default"; not yet.
- The v14 `<lacuna>` rule (#4584) is in neither arm. If v14 is promoted first: `--from 14 --version 16` and re-run.
- Migrating `stats-cross-model.mjs` to `lib/paired-stats.mjs` (its own follow-up, per that file's header).

## Verification

- Seeder dry-run: all 8 anchors matched exactly once; applied; default confirmed still v13, rows `[(13, True), (14, False), (15, False)]`.
- `vitest`: 30/30 (new metrics tests + `page-terms-parse`). `eslint` clean on changed files. `tsc --noEmit` clean.
- Draw smoke at n=1/2/3/40 per stratum; spend gate confirmed to refuse with and without `--route`.


JDerekLomas:
**Paid run approved and started (Derek, via the parent session): flat on `gemini-3.1-flash-lite`, $1.47, 640 calls — NOT `--route`.**

Why flat, recorded so it is not re-derived: the routing of non-Latin books to `gemini-3-flash-preview` rests on OCR evidence (lite over-relies on linguistic priors when *visual* decoding is hard). Translation is text-to-text on `ocr.data`; no translation model comparison exists in `scripts/eval/EXPERIMENTS.md`. A single model is also the cleaner design for a *prompt* effect — `--route` would confound arm with model across strata.

The separate question this surfaced — is full flash justified for **translation** on non-Latin books, or are we paying double on inherited OCR evidence? — is **#4759**, which proposes reusing this harness with a model per arm. Not started here; linked for the record.

The result will be stated as measured on flash-lite; whether it extrapolates to the strata that ship on full flash will be addressed explicitly, not assumed. Decision rule unchanged.

JDerekLomas:
## Result — do NOT flip v15 as-is; seed v16

Full report: `scripts/eval/results/translation-prompt-v15-report-2026-09-12.md` (commit c2edec01). 320 pages / 320 books, both arms flat on `gemini-3.1-flash-lite`, 640 calls, 0 errors, **$1.26** actual.

| # | criterion | result | pass |
|---|---|---|---|
| 1 | paired CI excludes zero | Δ **+28.9 pp** [+14.7, +44.7]; sign test 13–0 | yes |
| 2 | gain ≥ 3 pp | **66.7% → 96.3%** verified | yes |
| 3 | emission not down >20% | original-notes/page 0.42 → 0.77 (**+82%**) | yes |
| 4 | body not down >10% | −26.1% as written — **two v13 runaway loops** (MAX_TOKENS, 96K/138K chars) that v15 refused; excluding them **−0.3%** | no (as written) |
| 5 | no gate fires | invented tags −95%, housekeeping −91%, inline terms +22% | yes |
| 6 | blind judge ≤ 1.5× | v15 worse **8**, v13 worse **1** — 4 are headers not reproduced (item 4 by design), **3 are interpretive notes not written** | no |

**The real finding:** interpretive (non-`original:`) notes fall **1.27 → 0.81/page (−36%), in every stratum**. The verbatim rule's "if you cannot copy it exactly, OMIT the note" is being read beyond original-note quotations. The judge saw it ("Pietro Bembo, the cardinal", "references Quran 22:36" gone).

**Keep:** the verbatim rule itself is a large, clean win — Hebrew 8% → 93% verified, Arabic 29% → 100%; on those scripts v13's citations were mostly fabricated — and it did not game the metric (more notes, not fewer). Items 1/4 land. Loops on corrupted pages become `<warning>`s.

**v16:** v15 + one sentence scoping the omit clause to the quoted phrase; interpretive notes explicitly still wanted. Re-run only the judge (header class excluded from the brief) and an interpretive-note floor (~$1.30). Not started.

**Extrapolation to full-flash strata:** measured on lite only; direction probably robust, magnitudes not assumed — see the report's last section and #4759.

Rule not rewritten; Amendment 1 (loop rate as an outcome) is prospective. v15 stays `is_default:false`.

Issue #4780, opened 2026-09-13
IA OCR ingest: spot-check findings — dehyphenate at ingest, Unicode-aware gate tokenizer, refuse language-mismatched items


JDerekLomas:
## Spot check of the ingested Archive text (2026-09-13, Derek's request)

Six books sampled one page each across the accepted range (0.85–0.96), four read against the scan by eye; then three defects quantified over 1,047 ingested books / 26,621 pages. Parent: #4763.

**By eye.** Washington *Writings* 1931 (0.92): exact, footnote markers included. Society for Psychical Research 1913 (0.90): accurate; a junk first line from margin bleed (`‘= ¥ tw, ~`), a few dropped spaces (`Iam sureI canbe`). Fronto Loeb 1920 (0.87): English pages accurate; the dotted lacuna lines come out as `NOX, PUR IE 12O2E Rees TREES gees!`. *De officiis* notes 1888 (0.88): English commentary fine, italic Latin lemmas 10–20% misread (`prolici`, `inkumatum`, `exercilatus`), the one Greek word is invented. Top band excellent, middle good, the bottom band is where bilingual classical editions live.

## Three defects, measured

| defect | share | consequence |
|---|---:|---|
| **Line-end hyphenation retained** (`am-\nmunition`, `ex-\npositions`) | **22,320 of 26,621 pages (84%)** | split words are invisible to search and quotes; Gemini OCR joins them |
| junk first line of symbols (margin bleed) | 202 pages (0.8%) | cosmetic |
| **English-labelled book whose Archive text is detected as another language** | 4 books, incl. Aristotle *De Anima* (Greek text) accepted at 0.91/0.92 | the gate's tokenizer is `[a-z0-9']` — **Greek, Cyrillic, Hebrew are invisible to the score**, so a bilingual edition is judged on its English apparatus alone while its Greek pages may be garbage |

## Fixes
1. **Dehyphenate at ingest** in `ia-ocr-ingest.mjs`: join `xxx-\nyyy` when the next line starts lowercase (the XML gives line structure); keep the raw leaf in `page_revisions`. **Backfill** the 142K written pages with the same rule (idempotent; per-book via `book_events`, sweep_log row).
[...]

Issue #4763, opened 2026-09-12
IA OCR ingest, English run 2026-09-12: offset artifact, the 198K-page no-reference pool, other languages, and why text-only flash-lite cleanup does not pay


JDerekLomas:
## Where the free-OCR lane stands (2026-09-12)

`scripts/import/ia-ocr-ingest.mjs` (#4727) ran dry over the English pool on Hetzner (2,076 IA books with untranscribed pages; results `hetzner:/root/sl-ia-cache/_runs/2026-09-11-en-dry.txt`, XML cache 22 GB):

| bucket | books | fillable pages |
|---|---:|---:|
| ACCEPT (agreement ≥ 0.85 vs our Gemini pages) — **apply running now** | 620 | 65,977 |
| REJECT | 741 | 136,545 |
| no reference (< 5 Gemini pages, cannot calibrate) | 447 | **197,914** (674 transcribed) |
| no `_djvu.xml` on the item | 261 | — |

Derek's ask (this morning): a proper path for importing IA OCR for English books without OCR, possibly other languages, with a way to test quality; and flash-lite text cleanup as a cheaper middle band than image OCR. The quality test exists (per-book agreement against the Gemini sample). Four findings change the plan:

### 1. The biggest REJECT cluster is an alignment artifact — fix in PR (linked below)
292 rejected books sit at agreement 0.10–0.20 (43K fillable pages). Probed four: 500/515 reference pages score 0.15 at leaf offset 0 and **0.70–0.94 at offset −1** (the XML `<OBJECT>` sequence starts one leaf later than our `/page/n<k>` index on those items; the PAGE param on the first object reads `_0001` where the accepted control's reads `_0000`). The PR scores every book at offsets ±3, takes the offset most reference pages prefer (≥ 60% must agree, else UNSTABLE), and applies it to the fill. Jung 1919 and Collectanea Chemica 1893 flip to ACCEPT at 0.94/0.91; the 1801–1803 long-s books stay rejected on real quality. A full English dry run with the patched script is running: `hetzner:/root/sl-ia-cache/_runs/2026-09-12-en-dry-offset.txt`.

The same PR fixes the apply crash (`ocr: null` pages reject a dotted `$set`; MongoServerError 28) with a pipeline `$mergeObjects` write.

### 2. The no-reference pool is the biggest, and it costs ~$5 to unlock
[...]

PR #4783, opened 2026-09-13
ia-ocr quality: dehyphenate at ingest, Unicode gate tokenizer, language-mismatch guard, backfill (#4780)

JDerekLomas:
Fixes the three defects from the spot check in #4780 (parent #4763). One concern: the quality of the ingested Internet Archive text and the gate that admits it.

## In scope
1. **Dehyphenation at ingest** — `scripts/lib/dehyphenate.mjs`: join `xxx-\nyyy` only when the next line starts lowercase; uppercase/digit continuations (`Anglo-\nSaxon`, `1-\n2`) and paragraph breaks are left untouched. Applied to every leaf *after* loading (cached `.leaves.json` files get it too) and before scoring. The one known false join (`self-\npreservation` → `selfpreservation`) is documented in the helper and pinned in `tests/unit/dehyphenate.test.ts`; the spot check measures how often it happens.
2. **Unicode tokenizer** for the agreement score — `[\p{L}\p{N}']` with the `u` flag instead of `[a-z0-9']`, curly apostrophes folded. Greek/Cyrillic/Hebrew now count, so non-Latin garbage lowers the score instead of vanishing from it.
3. **Language-mismatch guard** — verdict `LANG_MISMATCH` (counted as rejected) when IA's `ocr_detected_lang` normalises (`language-normalize.mjs`) to a language the book neither has as `language` nor lists in `languages[]` (a facing-page edition is tagged that way, per `language-fields.md`). Both values are logged on every scored line.
5. `--ids <file>` on the ingester: re-score already-filled books (skips the "still has untranscribed pages" filter).
6. **Scorer fix found by the re-score** — Google-scanned items carry `<HIDDENTEXT/>` (no words) on many leaves (528 of 841 on Ante-Nicene Fathers vol. 8); since the ±3 offset search landed, a reference page whose own leaf is empty but whose neighbour has text scored 0 at the winning offset (`?? 0`), so 0.86 books re-scored at 0.000. Verified the OLD tokenizer reproduces the 0.000, i.e. scorer not tokenizer. Now only reference pages with an IA leaf at the chosen offset enter the median and the offset share (an empty leaf is never filled anyway). Ante-Nicene back to 0.866 over 251 pages (ingest had recorded 0.863 over 252).
7. **Backfill** `scripts/maintenance/dehyphenate-ia-ocr.mjs`: dry by default; per book via `book_events` (never a corpus scan on the unindexed `ocr.source`); only `ocr.source: 'ia_djvu'` pages; `saveRevisionsBeforeOverwrite` before the write (`reason: 'ia_ocr_dehyphenate'` — the raw leaf survives in `page_revisions`); `recordSweepAction` per book; the update matches on the exact current text so a page rewritten by another job between read and write is skipped and counted, never clobbered; idempotent (second run changes 0 — asserted by running it dry after the apply). 4h Mongo timeout.

- Rolling back De Anima or the other language-mismatched books: the Greek reads correctly (see #4780); the re-score under the new tokenizer decides, per book, by Derek.
- A dictionary check for compound hyphens: needs a wordlist per language; the false-join rate is measured in the spot check first.
- Dropping junk first lines (optional item 4 in the issue) — cosmetic, 0.8% of pages.

## CI note
The red `spend-perimeter` check is pre-existing on `main` (`scripts/eval/ia-ocr-cleanup-exp.mjs` lacks its `usage-ok` line there); PR #4779 carries the one-line fix. This PR adds no model call.

## Verification
- `npx vitest run tests/unit/dehyphenate.test.ts` — 8 passing.
- Backfill dry run on 4 books: `{"books":4,"pages_scanned":11,"pages_changed":10,"joins":54}` — example joins `anti-\nquity → antiquity`, `un-\nequal → unequal`.
- Spot check (one page per book, 12 books stratified by band, read against the scan) before and after the backfill, and the low-band re-score table, are posted on #4780 as they complete.

https://claude.ai/code/session_01UawoeZL26cWLL5QnqiVtkr


JDerekLomas:
Merged #4779 (65735e0d), which carries the `usage-ok` declaration for `scripts/eval/ia-ocr-cleanup-exp.mjs` — the spend-perimeter failure on this PR was inherited from main (verified: this PR's head does not contain that line). Updating this branch from main so the check re-runs clean.

PR #4807, opened 2026-09-13
feat(pipeline): hold the 197 wrong-leaf IA books out of every lane; untagged IA text audited tolerant (#4790)

JDerekLomas:
Makes the Archive-sourced OCR text (`ocr.source: 'ia_djvu'`, ~151K pages in 954 books) safe to flow through the derived lane. Handoff: `sourcelibrary-ops/handoffs/2026-09-13-ia-text-pipeline-safe.md`. Follows #4791 (#4790).

## Job 1 — untagged text: what actually breaks (audited, nothing fails closed)

| consumer | on untagged text | verdict |
|---|---|---|
| `blank-page-guard.mjs` | strips META tags, measures the rest; not on the IA write path at all (the ingester never calls it — ABBYY returns no words on blank leaves) | no effect |
| `cover-scoring.mjs` (+ TS twin) | `page_type` null → prose heuristics; no `title-page` bonus, `looksLikeTitlePage` needs 3 markdown headings IA text never has | **degrades silently** — mitigated: 19,816 of 22,279 pages ≤ 25 in these books are model-tagged (89%); only 2,463 early pages are IA |
| `english-source-detect.mjs` | `declaresEnglish` → null → function-word frequency, `basis` says so | tolerant, labelled fallback |
| `language-content-classify.mjs` | strips tags then stopword-classifies | no effect |
| `ocr-result-parse.mjs` | `extractPageType` → undefined; only called by the model batch collectors, never for IA pages | no effect (so `pages.page_type` is simply absent on IA pages) |
| `page-counts.mjs` | reads the `page_type` FIELD, not the tag; absent ⇒ ordinary content page, which is right (IA never fills leaves with < 20 tokens, so no blank leaf carries IA text) | no effect |
| `scholarly-pdf.mjs` / `scholarly-typst.mjs` | strip tags, pass text through; running heads and page numbers stay inline as body text | no effect (noise, not failure) |
| `strip-editorial-wrappers.mjs` | nothing to strip | no effect |
| `title-page-ocr.mjs` | no typed page → first prose page, and it labels the result `untyped_fallback` | tolerant, labelled fallback (audit scripts only) |
| pipeline phases (3.5 counts, 3.7 / 4 / 7 select on the `page_type` field) | IA pages are content pages: translated, indexed, chaptered; TOC detection falls back to the content regex | no effect |

**Decision: make nothing tolerant — it already is; add the pinning test** (`tests/unit/untagged-ocr-tolerance.test.ts`) so a future tag reader that refuses untagged text is caught there. Not chosen: a `<page-type>text</page-type>` backfill (151K rows asserting a type no model checked, buying only the cover bonus) or a classifier (the point of these pages is that no model touched them).

## Job 2 — the wrong-leaf books are HELD (applied 2026-09-13 22:40 CEST)

**197 CLASS C books / 50,911 IA pages** now carry `pipeline_auto.status: 'held'` + `pipeline_auto.hold {reason: 'ia-wrong-leaf-4790', issue, held_at, held_from_status, release}`. Held from: complete 48, images_complete 37, translate_partial 34, needs_attention 24, ocr_complete 23, translate_submitted 15 (jobs in flight), archive_complete 12, translate_complete 3, failed 1.

Why a status and not a filter: every worker already selects on `pipeline_auto.status`, so a held book leaves every lane with no query changed. What can still lift it is a WRITER — a rollback or batch write-back that sets the status unconditionally — so the marker is what the writers consult: the orchestrator's `setPipelineStatus` refuses (always enforced, no observe mode), and `NOT_HELD` is spread into the direct `updateOne` filters in enrich-worker, translate-worker (7 sites) and batch-collector.

**How a future run knows they are still held:** the book says so (`status: held`, the marker with its release sentence), `book_events` (`pipeline_hold` / `pipeline_release` / `pipeline_hold_reasserted`), `sweep_log` (`pipeline-hold-2026-09`), and `scripts/audit/pipeline-hold-drift.mjs` runs daily (crontab.production 07:55): CLOBBERED (marker, wrong status) / ORPHANED / LEAKED (translated after the hold) must be 0; RELEASABLE lists books with an `ia_ocr_leaf_repair` event newer than the hold. First run: 197 held, clean. Release: `hold-pipeline-books.mjs --release-held --reason ia-wrong-leaf-4790 --apply` restores each book's prior status (`--to ocr_complete` to re-flow instead).

Already-derived artifacts on the 197 (built BEFORE the hold, mostly from the 25-page Gemini preview): summary 9, chapters 15, index 19, quality_score 19, ai_metadata 10, 56 books with some translation (2,893 of the 50,911 IA pages), Supabase `page_translations` 6,251 rows / 29 books, `book_embeddings` 14. Translations self-heal after the joint repair (stale-translation predicate); the rest need the lane-3 clearing from `derived-metadata-lane.md` — **not done here, listed for the release step.**

**The 30 empty-target pages** (11 CLASS A books; 17 in one) are flagged `ocr.unreadable: true, unreadable_reason: 'ia_wrong_leaf_empty_target_4790'` (`flag-ia-empty-target-pages.mjs`, revision + sweep_log + recount): reader withholds, counts exclude, stale sweep withholds any translation. Clearing them (`repair-ia-ocr-leaf-offset.mjs --clear-empty`, a model read per page) stays Derek's call. While doing this: `translate-core.translatablePageFilter` / `isTranslatablePage` and the two translate-worker page queries now skip `ocr.unreadable` pages — a #4523 gap (they were translated, then withheld an hour later).

## Job 3 — re-score of the pre-Unicode-tokenizer books

Dry run on Hetzner over the 276 books first ingested before 6958027e whose reference or IA pages carry Greek/Hebrew/Cyrillic runs (`/root/sl-ia-cache/_runs/pipeline-safe/rescore-candidates.log`). Result posted as a comment when it finishes. Nothing applied.

## Not in scope, said plainly
- **Releasing the IA-filled books to flow.** 480 of the 954 are already `complete` (their derived artifacts came from the preview pages); only 4,878 of 150,603 IA pages are translated. Re-flowing them means resetting statuses and paying for translation — Derek's call, and a different PR.
- CLASS B (42 books, text right, images shifted) is the image repair's problem; not held.
- The lane-3 clearing for the 197 (above).

## Verification
`npx tsc --noEmit` clean; eslint 0 errors on changed files; vitest 50/50 across the two new tests + page-counts + stale-translation; hold + flag dry runs matched the apply counts (197 / 30); drift audit clean after apply.

https://claude.ai/code/session_0179nSZhk5h3uUcrYvz9GTFJ


JDerekLomas:
Job 3 re-score finished (dry, nothing applied): 276 pre-tokenizer books with non-Latin content — ACCEPT 225, REJECT 8, UNSTABLE 1, LANG_MISMATCH 2, REF_SHIFTED 40; **0 fall below their language cutoff** (en/fr 0.80, la/de/it 0.85), median +0.01. Full table, the derived-artifact inventory for the 197 held books, and the per-OCR-run alignment finding (preview pages aligned, batch pages shifted) are on #4790.

Issue #4837, opened 2026-09-14
enrich-worker: two stall shapes killed the enrichment lane for 5 days — a finished run that never exits, and groundQuotes() wedging the event loop on a 2,707-page book


JDerekLomas:
## What happened

The Hetzner `enrich-worker` (Phase 6 summary+index) has been effectively dead since **2026-09-10 00:28 UTC**. Two distinct stalls, found 2026-09-14 while trying to get the body-techniques books through to image extraction:

1. **A finished run that never exited.** PID 1362566 started 2026-09-10 00:00:03, printed `[ENRICH] Done — enriched=23, chapters=30 … 1693s` at 00:28 and was still alive 4d18h later at 0.2% CPU / 23 MB RSS, holding `/tmp/sl-enrich.lock`. The scheduler reported `running=[enrich-worker]` every 2 min the whole time, so it read as healthy. Last lines of that run: `Entity sync failed: Client must be connected before running operations` — a dangling handle (timer/socket) after the client closed is the likely reason the event loop never drained. Killed by hand.

2. **The respawned run wedged at ~99% CPU with no log output for 60 min.** PID 2616247, spawned 18:48 UTC, 8 books in flight; 7 of them printed `Quote grounding: dropped N/M batch quotes` by 18:51 and then nothing — none advanced to `summary_indexed`, all 8 sat at `summarizing`. The eighth was **Abulafia, Complete Writings (`69b93e229ebaf5a80f5531fb`, 2,707 pages, 139 batches)**.

   The wedge is `groundQuotes()` → `findBestMatch()` (`scripts/workers/enrich-worker.mjs` ~L806–L900). For every quote whose `quote.page` does not match, it falls back to scanning **every page of the book**, and `findBestMatch` is itself O(words × window × quoteWords) with a `slice().join()` allocation per window. For a 2,707-page book with hundreds of batch quotes that is on the order of 10¹¹ string ops, **synchronously**, so the event loop is blocked and the other 7 books' `await`s (revisions, book write, entity sync, embedding) can never resume. `withTimeout` cannot fire either — it is a timer on the same blocked loop. Killed by hand at 19:50; the 30-min orphan recovery put the 8 back to `translate_complete`.

## Why it matters

- 23 envelope books were queued behind this at `translate_complete` (body-techniques, Yam, Nubia…); corpus-wide, nothing has moved past `translate_complete` in five days.
[...]

PR #4849, opened 2026-09-15
fix(pipeline): stop the image re-dispatch loop (#4839) and the enrich-lane stalls (#4837)

### LINKED ISSUE #4839: Image extraction re-dispatch loop: a Lambda job that never reaches total is reaped, rolled back and re-submitted whole every ~35 min — 1,385 cancellations and 3,209 re-paid pages in 7 days
## The loop

Three cooperating mechanisms turn one never-completing page into a book that is re-extracted every ~35 minutes, indefinitely:

1. **Lambda/SQS image jobs never finish when a page produces no write result.** `jobs.progress` reaches `completed = total − k` (k = 1…8) and stays there; `failed` stays 0. Observed 2026-09-14 on four freshly-dispatched books: Talhoffer 268/273, Orchésographie 144/145, Scienza e pratica d'arme 107/108, Oraibi Soyal 100/108 — all within 90 s of dispatch, then no heartbeat. `processImageExtractionPage` does send a failure write for "page not found" / "no image URL" cases, so the missing results are the *other* failure shape — most likely a Lambda timeout/exception on the page that never reaches the write queue (DLQ). Nothing on the job side records it.
2. **`scheduler.mjs` zombie reaper** cancels any `processing` job with no heartbeat for 30 min (`cancel_reason: 'scheduler: zombie reaper (no heartbeat for >30min)'`) and unsets `book.job`.
3. **`pipeline-orchestrator.mjs` orphan detector** then sees `images_submitted` with no `job` and rolls the book back to `chapters_complete`; **Phase 8 in the same run re-dispatches the whole book** — every page again, not the missing k.

Because the same page fails the same way, the cycle is closed.

## Measured (jobs collection, 7 days to 2026-09-14 21:00 UTC)

- **1,385** `image_extraction` jobs cancelled; **13** completed; 0 failed.
- **3,209 pages** completed inside cancelled jobs — paid, then re-dispatched.
- `6a4a4c83547be098cf67f38c` *Disputatio Iuridica* — **304 cycles**, 5 pages per cycle, 2 complete every time, 09-07 → 09-14 (still `images_submitted`).
- `6a4a53e7e4a335875e1d5319` — **305 cycles**, 3 pages/cycle, 1 completes.
- `6a9f3f7283c3b522416e21e2` *Extirpación de la idolatría* — 95 cycles; `6a9f3f6c83c3b522416e1a03` *Historia del Nuevo Mundo* — 82 cycles; *Voyage au Darfour* 13; *Voyage à Méroé* 15.
- 5 jobs stalled right now (no update > 15 min, completed < total).

The small books cost pennies per cycle, which is why nothing alarmed. The four body-technique books dispatched tonight are 108–273 pages each — ≈ 630 pages ≈ $2 per cycle between them, so they were finalized by hand (see below) instead of being left to loop.

Also seen: the Hetzner realtime worker (`image-extract-worker.mjs`) and the orchestrator's Phase 8 both select `chapters_complete`; Oraibi Powamu was taken by both within the same minute (realtime finished it; the Lambda job then got reaped at 96/108). Two dispatchers for one status is a second, smaller source of double spend.

## Suggested fix

- **Finalize on drain, not on exact count.** When `completed + failed < total` and the job has had no heartbeat for the reaper window, mark it `completed_with_errors` and record the missing page ids on the job — the orchestrator already advances books on `completed_with_errors`. That alone closes the loop.
- **Re-dispatch only the missing pages** on rollback (the job carries `config.page_ids`; subtract the pages that already have `detected_images` from this job's model).
- **Bound the retries**: `pipeline_auto.retry_count` exists for exactly this; after N cycles park at `needs_attention` with the page ids.
- **Find the missing write results**: check the image-extraction DLQ / Lambda error logs for the four books above — the failing page ids are `config.page_ids` minus pages with `detected_images.model` from tonight.
- Pick one dispatcher per status, or make Phase 8 skip when the realtime worker is enabled.

## Hand-finalized 2026-09-14 (so this issue's loop stops on them)

`6aa7cf7efd6dd3b07d31d70f` Talhoffer, `6aa7cfa5fd6dd3b07d31dce1` Soyal, `6aa7cf64fd6dd3b07d31d3e7` Orchésographie, `6aa7cf74fd6dd3b07d31d602` Scienza — job cancelled with a reason naming this issue, `pipeline_auto.status` → `images_complete` with `pipeline_auto.images_note` recording the k missing pages, `detected_images_count` set from pages. Output-backed (≥ 92 % of candidate pages carry detections); nothing claimed that did not run.

JDerekLomas:
Fixes #4839 and #4837 — both found by the 2026-09-14 body-books image run, both live breakage that costs money every hour.

## #4839 — the image re-dispatch loop is THREE mechanisms, not one

Measured against production today, not inferred:

**1. The completion predicate counts the wrong thing.** `progress.completed` was recounted as "target pages carrying `detected_images`". A candidate page the model looked at and found **no illustrations on** never gets that field — the writer deliberately refuses to overwrite existing detections with an empty array. So the page is extracted, paid for, and permanently uncounted, and the job can never reach its total.

> Checked on 40 of 40 recent jobs: `total − completed` equals the number of zero-detection pages, exactly. In all 12 stalled jobs, every "missing" page had been extracted during that run.

Completion now counts pages attempted during the run (`image_extraction_updated_at >= job.created_at`) — the test the extraction Lambda already used to decide a page was current. The two sides of the job now agree on what "done" means.

**2. The orphan detector undoes the advance step.** The writer unsets `book.job` the moment a job completes. The Phase 4 orphan detector runs *before* Phase 8's completion check, so a **correctly finished** book looked exactly like an orphan: rolled back to `chapters_complete` and re-submitted whole.

> **59 of the last 60 successfully COMPLETED image jobs were re-dispatched this way.** One book cycled 6 times in 3 days.

It now leaves `images_submitted` books whose named job has already finished, for Phase 8 to advance.

**3. Nothing bounded the retries.** Stalled jobs are now finalized on **drain** rather than on an exact count (`scripts/workers/lib/image-job-drain.mjs`), with the pages that never reported recorded on the job row — an abandoned page must be findable, never a silent skip. A job that drained with *nothing at all* is still cancelled (zero results means a broken lane, not a finished one), marked so Phase 8 can count it, and a book whose dispatches keep coming back empty is parked at `needs_attention` after 3 — a recorded give-up, not a quiet advance.

### Waste removed

~460 reaper cancellations/day (19–28/h). Sampling 40 of them by their own `gemini_usage` rows — the only per-job record that cannot be overwritten — **all 40 had already paid for every page they were cancelled over**, 376 pages across those 40.

## #4837 — two stall shapes that killed enrichment for 5 days

- **`groundQuotes()` wedged the event loop** at 99% CPU on Abulafia (2,707 pages). A quote whose stated page does not match falls back to scanning every page, O(words × window × quoteWords) with a `slice().join()` per window — synchronously, so the per-book timeout could not fire and seven other books' awaits froze behind it. Moved to `scripts/workers/lib/quote-grounding.mjs`, **matching semantics unchanged**, plus: a page prefilter that is a strict upper bound on any window's score (so it is lossless), a 60s per-book budget that **reports** what it did not reach, and yields between pages so timers can fire.
- **A finished run held the lock for 4d18h.** Fire-and-forget entity sync and embedding writes are now tracked and drained *before* the Mongo client closes — they were failing with `Client must be connected` on every run, so entity sync has silently not been happening — and the worker exits explicitly.
- **Liveness.** The scheduler could only see "holds its lock", which cannot distinguish a wedged worker from a busy one; it logged `running=[enrich-worker]` through both stalls. Workers now beat a file from an unref'd 60s timer (a blocked loop cannot touch it); the scheduler kills a lock holder silent for 10 minutes, escalating TERM→KILL, and records it in `cron_runs` as `killed_stalled`.

## Tests

20 new unit tests; full suite green (4,034 passed), `tsc --noEmit` clean. They pin the drain classification (including the exact 12/8/4 shape that looped), that an older extraction cannot close a newer job, the completion predicate, prefilter losslessness, the budget report, and that grounding actually yields.

## Out of scope, deliberately

- **Re-dispatching only the missing pages.** With drain-finalize, a partially reported job completes with its gaps recorded, so the whole-book re-submission path is no longer reached; subtracting pages would add a second notion of "already done" for no current gain.
- **Two dispatchers for one status.** `image-extract-worker.mjs` (realtime) and Phase 8 both select `chapters_complete`, and a book can be taken by both within a minute. Real, smaller, and independent — worth its own issue.
- **6 `pending` image jobs from ~68 days ago.** The reaper only looks at `processing`, so they sit in the active-job count forever. Harmless at the current cap (18/50), noted here rather than widened into this PR.

## Deploy note

The worker half reaches production when Hetzner's hourly `git pull` picks up `main`. The `src/lib/job-completion.ts` half is the **writer Lambda** and needs the usual hand-deploy (`npm run lambda:deploy -- write-processor`) — until then the Hetzner-side drain covers the same jobs, which is why the guard was put on the side that can observe progress.

After merge, the before/after is one query: `jobs` where `type: 'image_extraction', status: 'cancelled'` per hour — currently 19–28/h. I did not apply a one-off fix to the books looping right now, because without this code deployed they would simply be rolled back again on the next orchestrator run.


JDerekLomas:
### Measured on Abulafia itself, not on synthetic text

Read-only, against the live book (`69b93e229ebaf5a80f5531fb`, 2,707 pages / 2,674 groundable):

| | |
|---|---|
| 20 quotes grounded (10 real excerpts deliberately mislabelled with the wrong page + 10 that match nothing) | **603 ms** |
| the same work without the prefilter, extrapolated from 60 real pages | **~660 s** |

The unfiltered fallback costs ~740ms per 60 real pages per quote, so those 20 quotes alone would take about 11 minutes of **synchronous** CPU — and the wedged run had hundreds of batch quotes across 139 batches, which is the hour of 99% CPU with no log output seen on 2026-09-14.

7 of the 10 real excerpts landed back on the page they were taken from; the other 3 fall below the unchanged 0.8 / length thresholds, and all 10 non-matching quotes were correctly refused. The prefilter is a strict upper bound on any window's score, so it cannot change which quotes match — only how many pages get scanned.

JDerekLomas:
### One-pull caveat on the stall detector

The scheduler seeds the heartbeat file when it spawns a worker, and the worker then beats every 60s. Both come from the same Hetzner `git pull`, but there is a single window where a worker spawned by the **new** scheduler is still running the **old** worker code: it would never beat, and after 10 minutes of silence the scheduler would kill it mid-run.

It is self-correcting (books stuck at `summarizing` are recovered by the existing 30-minute orphan sweep) and cannot recur after that pull. A worker spawned by the *old* scheduler is never at risk: no heartbeat file exists, and `heartbeatIsStale(null, …)` returns false by design — only an observed, stale beat convicts.

Issue #4868, opened 2026-09-15
Nothing alarms on embedding coverage: add a daily rows-added-vs-records-created check


JDerekLomas:
Three separate embedding gaps (#4865, #4866, #4867) all ran for weeks with no signal. The only tell available today is the newest `created_at` per Supabase table, read by hand.

**Why it is silent by construction:** an unembedded book and a book nothing matches both return an empty list, so no read path can distinguish them. Coverage has to be measured on the write side.

**Wanted:** a daily audit, same shape as the Gemini metering gap detector already running on Hetzner at 08:30 — for each store, rows added in the last 24h against records created in the same window, plus absolute coverage against its denominator:

| store | denominator |
|---|---|
| `book_embeddings` | live books (`visible && pages_count > 0`) |
| `page_translations` | pages of live books with OCR or translation |
| `artwork_embeddings` | artworks with an enrichment description |
| `gallery_text_embeddings` / `clip_embeddings` | `gallery_images` |
| `page_texts` | translated pages per language |

Two things the script must do, both learned the hard way today:

- **Assert the probe fires before reporting a gap.** My first pass reported "0% of books have page vectors" because `page_translations` has no `id` column, every request errored, and a missing `Content-Range` header was read as zero. A positive control on a known-embedded row caught it. An audit that cannot tell *broken* from *nothing there* is not measuring anything (`measurement-instruments.md`).
- **Do not read `page_translations.updated_at` as a write timestamp.** It mirrors the Mongo source's `updated_at` and reports the table as months stale while it is being written to right now — documented in `.claude/docs/embeddings.md`, verified again today.

Report-only; file findings rather than acting on them.

JDerekLomas:
**Merged** — PR #4876, on `main` as `b6bcaf32`, deployed to Hetzner and run there to confirm it works from that checkout:

```
EMBEDDING COVERAGE

  page_translations           4,874,374 rows   no write timestamp (see header)
  book_embeddings                66,241 rows   21 in 24h
  artwork_embeddings             22,266 rows   0 in 24h
  gallery_text_embeddings       221,546 rows   74 in 24h
  clip_embeddings               340,794 rows   12,653 in 24h
  page_texts                     59,031 rows   no write timestamp (see header)

  ✔ no findings.
```

`page_translations` is up **+272,488 rows** since yesterday — the #4866 backlog run feeding through, which is what this would have shown in August if it had existed.

[...]

Issue #4867, opened 2026-09-15
2,194 translated live books have no book-level embedding — enrich Phase 6.5 did not write


JDerekLomas:
## Measured 2026-09-15

`book_embeddings` holds 35,974 rows. Joined against live books on **both** id forms (`id` and `_id` — 16,343 books carry a re-minted `_id`, so a single-form join would overcount):

- 16,455 of 41,891 live books have a book-level vector (39.3%);
- of the 25,436 without one, **23,242 have no translation** and are explained by #4866;
- **2,194 have translated pages and still have no row.** That is not explained by the dial or by the translation gap.

Their `pipeline_auto.status`:

```
translate_complete 1383   complete 435   needs_attention 131   archive_complete 94
failed 38   images_complete 26   translate_partial 24   held 42   parked 13
loop_quarantine_hold 6   metadata_enriched 1   summary_indexed 1
```

**540 of them already have a summary**, i.e. enrichment ran far enough to produce the text Phase 6.5 embeds, and the row still was not written. The 435 at `complete` are the sharpest case: the status claims the whole pipeline finished.

This is the `pipeline-status-truth.md` shape — a status that runs ahead of its output — applied to a phase whose output lives in another database, where nothing downstream fails loudly when it is missing.

## Wanted

1. Why Phase 6.5 does not write for these (silent catch? projection? a path that advances status on the failure branch?).
2. A backfill for the 2,194 (`scripts/migration/backfill-book-embeddings.mjs --incremental` skips existing rows).
3. If the write can fail silently, it needs the "output OR a recorded skip" treatment the pipeline-status invariant already requires.

Related: #4865 (embedders locked out of the envelope lane), #4866 (untranslated books never embedded).

JDerekLomas:
**Done 2026-09-15.** `backfill-book-embeddings.mjs --incremental` on Hetzner: **30,246 embedded, 0 skipped, 0 errors, 749s** (~40/s).

Re-measured after, joining on both id forms:

```
live books: 41,891   with book vector: 41,891 (100.0%)   still missing: 0
```

Book-level coverage went **39.3% → 100%**, which closes this issue's 2,194 and also gives the 23,242 untranslated books of #4866 book-level retrieval (title + author + description + categories via the fallback path), while their page vectors catch up separately.

Two things left deliberately open, not part of this issue:

[...]

Issue #4866, opened 2026-09-15
23,200 live books are invisible to semantic search: OCR'd, never translated, so never embedded


JDerekLomas:
## Measured 2026-09-15

Of 41,891 live books (`visible: true && pages_count > 0`), **23,200 have OCR text and zero translated pages** — 651,412 pages. Sampling 120 of them, **3% have any page vector** (n=120). They are therefore absent from every semantic surface: `/api/search`, `search_concept`, the librarian, `match_semantic`.

By language, the books with no book-level vector at all: **Chinese 12,059, Latin 10,599**, German 1,127, English 848, then a long tail.

This is the generalised form of a lesson already recorded: *"not found" is never evidence we do not hold it* — semantic search is blind to untranslated books, so a query that should surface a Chinese or Latin holding returns nothing and reads as an absence.

## It does not need translation to be fixed

`pageEmbeddingInput` (`scripts/lib/page-embedding-text.mjs`) **falls back to the original text** when there is no translation — deliberately, unlike `pageTextForLang`, which must not. So these pages can be embedded as they are, in their own language, and multilingual retrieval will reach them. No translation spend required.

Cost at the documented rate (pages × min(chars, 8000) ÷ 4.29 ÷ 1e6 × $0.20): **≈ $18 for one full pass** over the 651,412 pages.

## Blocked on

#4865 — `embed-gemini` uses the unscoped spend guard, so an envelope cannot fund this and the nightly incremental is refused every run.

## Order of work

1. #4865 (scoped guard) — otherwise this backlog cannot be funded without opening the global dial to every lane.
2. Backfill by `--books-file`, confined to an envelope. Note `--missing-only` **cannot reach a book with zero rows** (it only scans books already present) — that is exactly this population, so it must be `--books-file`.
3. Re-measure coverage and record the number.

JDerekLomas:
**Progress 2026-09-15/16.**

**Batch 1 done** (the 3,000 largest untranslated books, 236,861 pages): **105,320 pages embedded, 129,893 already present, 50 errors, 11,127s**. The 50 errors are all `The operation was aborted due to timeout` on the embed call; the pages are simply left unembedded and a re-run picks them up, since selection is by absence of a row.

**Cost, measured rather than estimated** — 22 `gemini_usage` rows covering 105,320 texts and 31.1M input tokens:

```
$0.0000591 per page   (≈295 tokens/page)
batch 1 .................. $6.22
[...]

Issue #4540, opened 2026-09-01
The allowlist bypasses the pause but not the spend dial: a named book has no lane


JDerekLomas:
## The need

There is no way to say "run *these* books, whatever else is going on." Today the only two dials are global: the pause flag, and the $5/day spend ceiling. A named book that a human wants finished — a partner's holdings, a book someone is waiting on, a repair after an incident — has no lane. It queues behind whatever unrelated sweep is spending the budget.

Concretely, today (2026-09-01): measured Gemini spend hit **$21.25 against the $5/day dial**, $13.04 of it a single `ft_ladder_skeptic` sweep (4,699 calls, 11:00–14:00 UTC). The spend guard did exactly its job and closed the gate — `pipeline-orchestrator-worker` has logged `ocr_submitted: 0, images_submitted: 0` on every run since. A Wellcome book that is fully archived and needs 186 pages of OCR (~$0.15) sits at `archive_complete` and will not move until the day rolls over, because a sweep it has nothing to do with spent the ceiling.

That is the wrong failure. The ceiling should stop *unbounded background work*; it should not stop *a person choosing a book*.

## What exists, precisely

**The allowlist already exists — but only for the pause.** `system_config.processing_control.allow_book_ids[]` / `allow_collections[]` / `allow_scopes{}` (#2610, helpers in `scripts/workers/lib/selective-unpause.mjs`). `shouldBypassPause(control, {bookOverride})` returns true when a scope is configured, and every phase then post-filters its candidates to that scope via `applyBookOverride`. #2713 fixed the phase-local limits that used to strand scoped books.

**The spend dial has no such thing.** In `scripts/lib/spend-guard.mjs`, `budgetAllowsDispatch(db, label, {bypass})` accepts a bypass, and in `pipeline-orchestrator.mjs` every one of the seven call sites passes the same thing:

```
{ bypass: !!BOOK_OVERRIDE }
```

`BOOK_OVERRIDE` is the CLI flag `--book=ID` (line 237). **The config-driven allowlist is not consulted by the spend guard at all.** So a book can be allowlisted past a global pause, and still be refused by the dial.

[...]

PR #4869, opened 2026-09-15
fix(embeddings): both embedders were locked out of the envelope lane (#4865)

### LINKED ISSUE #4865: Embeddings are starved by the shared daily dial: both embedders use the UNSCOPED spend guard and are refused on every run
## Measured 2026-09-15

Both embedding workers are scheduled and running — and **every run is refused**:

```
embed-gemini (every 4h):      CEILING REACHED — no new dispatch   ($36.67 / $5.00 today)
image-embeddings-cron (02:00): CEILING REACHED — no new dispatch   ($19.91 / $5.00 today)
```

Logs rotate daily through today, so the crons are alive; they just never get to dispatch. OCR and translation spend the $5 dial before 02:00/04:30 every day.

**Root cause:** `embed-gemini.mjs:236` and `image-embeddings-cron.mjs:92` call **`budgetAllowsDispatch`** — the pre-#4540 unscoped guard. Every other paid worker (`enrich-worker`, `image-extract-worker`, `translate-worker`, `pipeline-orchestrator`) calls **`budgetAllowsDispatchScoped`**, which falls through to the scope-envelope lane when the global dial is closed. The two embedders cannot reach that lane at all, so an envelope cannot be used to fund them and they are locked out whenever the dial is spent — which is every day.

## What it cost, in coverage (all measured today, not recalled)

Denominator: 41,891 live books (`visible: true && pages_count > 0`).

| store | coverage |
|---|---|
| `book_embeddings` | **16,455 / 41,891 = 39.3%** (exact — pulled all 35,974 rows and joined on both id forms) |
| page vectors, translated books | 90% have ≥1 (n=200, Wilson CI 85–93); only **67% are ≥90% covered**; 80% of pages |
| page vectors, untranslated books | **3%** (n=120) |
| `gallery_text_embeddings` | newest row **9 Aug**, while **10,051** gallery images have been created since 10 Aug |
| `clip_embeddings` | newest row **27 Aug** |
| `artwork_embeddings` | 21,925 vs 21,989 enriched artworks — fine (only 2 artworks created since July) |

Spot check on 17 books whose gallery images were extracted after 10 Aug: 5 have no `gallery_text_embeddings` rows, 4 have no `clip_embeddings` rows.

## Fix

Switch both to `budgetAllowsDispatchScoped` and honour `envelopeIds` (confine the run to those books, exactly as `enrich-worker` does). Then an embeddings envelope can fund the catch-up without touching the global dial, and the nightly incremental can keep up on a few dollars a month.

## Why nobody noticed

An unembedded book and a book with no match return the same empty list, and the only tell was the newest-row timestamp per table. This is the same class the doc already warns about — "a step outside the pipeline can be switched off without anything downstream noticing" — in a new form: not switched off, **starved**. A daily coverage check is filed separately.

## Instrument note

My first probe of page-vector coverage reported **0%**. It was wrong: `page_translations` has no `id` column, every request 404'd on the column name, and my code read a missing `Content-Range` header as zero. A positive control on a known-embedded book caught it. Any script written against these tables should assert the probe fires before reporting a gap.

JDerekLomas:
Closes #4865.

## What was wrong

`embed-gemini.mjs` and `image-embeddings-cron.mjs` called the pre-#4540 **unscoped** `budgetAllowsDispatch`, which can only see the global daily dial. OCR and translation spend that dial before either cron runs, so from August to 15 September **every scheduled run was refused** — crons alive, logs rotating nightly, nothing written:

```
embed-gemini (4-hourly):  CEILING REACHED — no new dispatch  ($11.75, $25.23, $29.60, $36.67 / $5.00)
image-embeddings-cron:    CEILING REACHED — no new dispatch  ($19.91 / $5.00)
```

An embeddings envelope could not fund them either: the envelope lane is unreachable from that function. Every other paid worker (`enrich-worker`, `image-extract-worker`, `translate-worker`, `pipeline-orchestrator`) already uses the scoped guard.

## What it cost, measured 2026-09-15

Against 41,891 live books (`visible && pages_count > 0`):

| store | coverage |
|---|---|
| `book_embeddings` | **16,455 (39.3%)** — exact; all 35,974 rows pulled and joined on both id forms |
| page vectors, translated books | 90% have ≥1 (n=200, Wilson CI 85–93); **67% are ≥90% covered** |
| page vectors, untranslated books | **3%** (n=120) — 23,200 books, 651,412 pages |
| `gallery_text_embeddings` | newest row **9 Aug**; 10,051 gallery images created since |
| `clip_embeddings` | newest row **27 Aug** |

## The change

- Both workers call `budgetAllowsDispatchScoped`.
- `embed-gemini` narrows its book set to `envelopeIds` **after** the mode branch has chosen it, so an envelope *confines* and can never widen: a `--book` run for a book outside the envelope exits rather than turning into a corpus run, and `--books-file` / `--missing-only` / `--restale` are intersected with it.
- `image-embeddings-cron` runs whole when an envelope is open — its three child backfills take no book filter, so "an open envelope" means "image embeddings may run tonight". Their spend is attributed to the envelope where a `book_id` exists and to the global dial otherwise, exactly as the envelope contract already states.
- Guard test pins both call sites (including that the unscoped name is gone from the file — importing it and calling it on one path is how this shape survives) and the narrowing form.
- `.claude/docs/embeddings.md` records the mechanism and the tell: **a store's newest `created_at` is weeks old while its cron log is dated today.**

## Out of scope

- The actual catch-up runs — funded by a scope envelope and tracked in #4866.
- #4867 (2,194 translated books with no book-level vector, which the dial does not explain) and #4868 (daily coverage detector) are separate.
- The dial itself is untouched.

`npx tsc --noEmit` clean; new test passes.


JDerekLomas:
Added a second commit: `scripts/migration/backfill-gallery-text-embeddings.mjs` (#4873). It was untracked in the Hetzner working tree since April — the sole writer of `gallery_text_embeddings` (212,433 rows), shelled out to by `image-embeddings-cron.mjs` every night, and missing from every other checkout. Found because running the cron from this branch failed that phase with MODULE_NOT_FOUND. Bytes are verbatim; the header records provenance, idempotency and that it is billed.

Issue #4851, opened 2026-09-15
Cuneiform 'translations' are recitation, not reading: a CDLI tablet catalogued as medical prescriptions (BAM 6, 555) carries an Enūma Eliš Tablet VII translation


JDerekLomas:
**Where:** https://sourcelibrary.org/book/neo-assyrian-medical-prescriptions-bam-6-555-p394421/page/69aafe226ccc25ab9ea1c343 (book `69aafe226ccc25ab9ea1c342`, contributing library CDLI, 1 page).

**Reader question (Derek, 2026-08-29):** "Was this really translated from the image or from another source in, e.g., German?"

**Verified against the record:** the OCR text opens with tag soup describing the photograph (fragments K.2421, K.2511, K.16765, Library of Ashurbanipal) and the translation is a summary of **Enūma Eliš Tablet VII — the fifty names of Marduk — with a didactic epilogue and royal colophon**. The book title says medical prescriptions. Either the catalogue title is wrong or the model recognised (or guessed) a famous tablet and reproduced it from memory; either way the "translation" was not produced by reading the signs on this photograph. A vision model does not read cuneiform from a tablet photo; what it emits is recall keyed off the museum number.

**Scope:** 4 CDLI books, all 4 carry translations. Small, but the pattern is a fabrication class we badge nothing for.

**Proposed:**
- For cuneiform (and any script the OCR route cannot actually read), the source text should be the publisher's transliteration (CDLI ships ATF for every P-number) and the "OCR" lane skipped; translate the ATF, not the photo.
- Until then, hide or badge the 4 as unverified, and reconcile this record's title against the P-number.

Related: #1285 (ANE collections, Akkadian OCR fixes), #4777 (verifier on fabricated citations).

Feedback-ID: 6a9286982dc98a60c3b38ab4

JDerekLomas:
Taking this (pipeline follow-up session, 2026-09-15): reconcile the P394421 title, then ATF-as-source plan; hide/badge decision goes to Derek.

JDerekLomas:
## Reconciled against CDLI: the catalogue is right, the reading is invented

Checked all four CDLI books against CDLI's own record (`https://cdli.earth/artifacts/<P>`, `Accept: application/json`).

[...]

Issue #4850, opened 2026-09-15
OCR degenerates into repeated Thai-script runs on a Balinese lontar: Kidung Panji Malat, 73 of 423 pages are 18K-char loops with no translation


JDerekLomas:
**Where:** https://sourcelibrary.org/book/kidung-panji-malat-rasmi-or-26275/page/6a20332318654bf8e19043d3 (book `6a20332318654bf8e19043ce`, Javanese/Balinese palm-leaf, 423 pages).

**What the reader saw:** "All repeating" (Derek, footer widget, 2026-09-14).

**Verified:** the flagged page's `ocr.data` is 18,000 chars of one ~14-char string in **Thai** characters repeated to the length cap; translation is empty. The OCR header says `<language>Balinese</language> <script>handwritten</script>` and then transcribes Balinese aksara with Thai code points. `pages.ocr.data ~ /(.{40,})\1\1\1/` matches **73 of 423** pages in this book. `pipeline_auto.status` is `translate_complete`, so the pipeline counts this as done.

**Two defects:**
1. No degenerate-output gate at the OCR write boundary — a repeat-ratio check (unique-lines / lines, or longest repeated substring share) would have rejected these instead of storing them and marking the book complete. Sibling of #4765 (translation-side loop gate); the OCR side needs the same guard, and the detector should sweep the corpus, not just this book.
2. Script mismatch: Balinese script emitted as Thai. Whether re-OCR on the full-flash non-Latin route (per #4762) fixes it, or lontar needs its own handling, needs a 10-page probe before a re-run.

**Not in scope here:** the translation loop gate itself (#4765).

Feedback-ID: 6aa75b98d8960dba640fbc20

JDerekLomas:
Taking this (pipeline follow-up session, 2026-09-15): OCR write-boundary loop gate + corpus detector + 10-page probe estimate.

JDerekLomas:
## Gate, detector and probe — PR #4882

### 1. Write-boundary gate (`scripts/lib/ocr-loop-guard.mjs`)

[...]

Issue #4806, opened 2026-09-13
The free-OCR gate cannot score Chinese — 1.59M pages were written off on a tokenizer artifact


JDerekLomas:
## The free-OCR gate cannot score Chinese, and we concluded Chinese was hopeless using it

**1,590,211 untranscribed pages in 12,074 Internet Archive books are Chinese** — the second largest pool in the library after Latin, and **10,468 of those books already carry ≥ 5 model-OCR'd reference pages**, so they are gate-ready today. The free-OCR lane (#4763, #4790) has never been run on any of them.

It was skipped on the evidence of the July baseline (`scripts/eval/results/ia-ocr-baseline-pilot-2026-07-24.md`), which listed Chinese as an **EXPECTED-COLLAPSE CONTROL**: 0 of 29 books aligned, every one failing as `only_0_of_5_probes_matched`. That looked conclusive. It is an artifact of the measuring instrument.

### The instrument is blind to unspaced scripts

Both the July pilot and the production gate score agreement over **word tokens**. The tokenizer was `[a-z0-9']` until 2026-09-13 and is now `[\p{L}\p{N}']+` with the `u` flag (#4783). Neither works on a script without spaces:

| tokenizer | tokens in `太上感應篇曰禍福無門惟人自召` |
|---|---|
| old, `[a-z0-9']` | **0** |
| current, `[\p{L}\p{N}']+u` | **1** (the entire run) |

So under the old tokenizer every Chinese book scored 0 no matter how good the text was — which is exactly the "collapse" the July pilot recorded. The Unicode fix did not repair it: one token per page turns the sequence ratio into a **binary whole-page exact match**. Demonstrated on a realistic single-character OCR near-miss:

```
word-token ratio (current gate): 0.000   tokens: 1 vs 1
character ratio (what CJK needs): 0.977  chars: 22 vs 21
```

A near-perfect page scores zero. **Chinese has never been measured by an instrument capable of measuring it**, so we do not know whether the Archive's Chinese OCR is usable. We only know our ruler reads nothing.

This is the third time on this lane that the largest cluster turned out to be the detector's own artifact — after the 0.10–0.20 leaf-offset cluster (#4763) and the non-Latin pages invisible to the ASCII tokenizer (#4780). Same lesson each time.

### Fix

[...]

Issue #4765, opened 2026-09-12
Translation has no gate for degenerate looping OCR — every fabrication found in the #4759 read came from one


JDerekLomas:
Surfaced by the free observational read in #4759 (`scripts/eval/results/translation-model-obs-report.md`), not by a user report.

## What the judge saw

A blind Claude judge over 30 within-book translation pairs flagged **five** translations as containing material with no basis in their source. All five were on the full-flash side, and all five share one cause: the OCR input was a **degenerate repetition loop** — a single syllable repeated for 16,000 to 24,000 characters. The model, handed that, wrote plausible connected prose.

This is not a model-quality finding. Flash-lite produced zero fabrications in the same read, but it was never handed those pages: the looping pages are disproportionately ones a later pass redid. **Either model will invent text when the input is a loop.** The fabrication is a function of the input, not the engine.

## Why it matters more than the count suggests

The output is stored in `translation.data` and read by the reader, search, embeddings, quotes, exports, IIIF and the MCP tools. A fabricated translation of a page whose transcription is garbage is indistinguishable downstream from a real one. This is the same class as #4523 and the `<note>original>` fabrication work in #3825 — invented content presented with the same confidence as real content.

## Proposed gate

Refuse to translate a page whose OCR text is degenerate, and record the refusal rather than skipping silently (absence is not failure — a skip must leave a record).

Detection is cheap and reference-free. Candidate signals, to be calibrated against the corpus before any threshold is fixed:
- longest repeated n-gram covering some large fraction of the text
- distinct-character or distinct-token ratio below a floor, length-aware
- compression ratio of the page text

Note the shape of the trap: a threshold tuned on Latin script will misfire on scripts with small syllabaries and on legitimately repetitive liturgical text, which is exactly the material we hold. **Positive-control it on real repetitive-but-genuine pages** (litanies, mantra repetition, Tibetan `mani` sequences) before enabling, or the gate will withhold good pages.

## Also worth checking

How many already-stored translations were made from looping OCR. If the count is material, that is a withholding candidate in the same shape as #4756, not a re-translation job — the OCR has to be fixed first.

JDerekLomas:
Gate shipped in PR #4882 (with #4850's write-boundary twin).

[...]

PR #4882, opened 2026-09-15
Refuse degenerate looping OCR at both boundaries — write and translate — and replace fabricated cuneiform with CDLI's ATF (#4850, #4765, #4851)

JDerekLomas:
Two reader-reported fabrication classes, from the 2026-09-15 feedback triage.

## #4850 — OCR that degenerates into a repetition loop

A reader reported "all repeating" on a Balinese lontar. **148 of its 423 pages** are loops, and **96 of those were translated** into fluent invented prose — the #4765 fabrication with its cause upstream. The pipeline stored every one and marked the book `translate_complete`.

Existing guards miss this class: `HALLUCINATION_LIMIT` is a runaway-**length** test (153 of the loops sit under 25,000 chars) and the blank-page guard needs an inkless image, which a loop does not have.

- `scripts/lib/ocr-loop-guard.mjs` — `loopVerdict()` measures the union of exact periodic runs as a share of the transcription body, in code points (a word-token ratio reads spaceless scripts as one token, #4806). Refuses at ≥50% with ≥6 repeats; records the refused text in `page_revisions`; `OCR_LOOP_GUARD=off` kill switch.
- Wired into **every live OCR write path**, not one helper (#3362/#3365 shipped the same bug twice that way): batch collector, `realtime-ocr`, `realtime-reocr-efm`, both legacy collectors, and both `batch-ocr-async` routes via the TS twin `src/lib/ocr-loop-guard.ts`.
- `scripts/audit/ocr-loop-corpus.mjs` — read-only detector over Atlas (checkpointed) or the local mirror, reporting the loops **and** the translations built on them.

**Measured.** Calibrated on 53,192 judged pages from 2,000 books; whole corpus: **69,223 loop pages of 6,091,450 (1.14%), 53,628 already translated.** Three false-positive classes were found during calibration and each is now a test that must pass: index leader dots, `&nbsp;` layout padding, `[illegible]` markers. The residual risk — genuinely repetitive text (a litany, a dhāraṇī folio) — is why refusals are recorded rather than dropped: 74% of refusals repeat ≥100 times, but ~17% sit in a 6–11 repeat band where a real mantra page can be refused.

**Probe ($0.057, approved).** 10 looping pages re-read on full flash with the current prompt: the Thai code points are gone and 9 of 10 no longer loop — but mostly because the model now *declines* (`<unclear>… not transcribed</unclear>`). It still cannot read the manuscript. Lontar needs its own lane; a blanket re-OCR would buy ~$2.40 of "unclear".

## #4851 — cuneiform "translations" are recitation

CDLI publishes P102318 as "2 grain-fed sheep … from Abbasaga"; our page claimed "fresh apples … to the palace" at `<confidence>0.95</confidence>`. The BAM 6, 555 *medical* tablet carried a translation of Enūma Eliš Tablet VII. Reconciled against CDLI: our catalogue titles are right, the readings are recall keyed off the museum number — as `/blog/cuneiform-ocr` already measured ("a cuneiform commentator, not a cuneiform reader"). Those four books are that post's exhibit tablets, published with the fabrications as transcription.

- `scripts/import/cdli-atf-source.mjs` — CDLI's published ATF becomes the source text (`ocr.source: 'cdli-atf'`, `ocr.model: null`). Refuses anything that is not one artifact to one page: the Hammurabi **composite** and the four-sided obelisk are held.
- `scripts/maintenance/hide-cdli-fabricated.mjs` — withhold by name (`visible:false` + `hidden:true` + `hidden_reason` flip-guard), `--withdraw-text` for the two that cannot be repaired.

**Already applied to production** (Derek's call, this session): 2 repaired from ATF, 2 text-withdrawn, all 4 hidden, Supabase catalogue synced — verified `visible:false` on both sides. Nothing deleted; every fabrication is in `page_revisions`.

## Out of scope

- The **53,628 translated loop pages** corpus-wide: a withholding decision, in #4765's shape, not a re-translation job.
- This lontar's own 148 loops + 96 translations: same decision, awaiting Derek.
- The obelisk and the Hammurabi composite: need a surface→page mapping before ATF can land.
- Two `/blog/cuneiform-ocr` "Read in Source Library" links now point at hidden books (the obelisk and Hammurabi). Authored public copy — proposed change, not made.

## Checks

`npx tsc --noEmit` clean · eslint clean · `tests/unit/ocr-loop-guard.test.ts` 15/15 (8 of them "must NOT refuse" controls).


JDerekLomas:
Two follow-ups since the description was written, both now green:

- **Seven more OCR writers guarded.** Wiring the collectors left `src/workers/write-processor-logic.ts` (the Lambda/SQS path that stored 125K pages before going idle), `batch-save`, `import-batch`, `process/batch`, `contribute/process`, `ocr-correct-grounded`, `apply-reocr-verdicts` and `mineru-ocr-worker` unguarded. `tests/unit/ocr-write-paths-guarded.test.ts` now sweeps for writes of OCR text and fails on any file importing neither guard twin, with a reasoned allowlist for text that is not a model reading (IA's delivered OCR, CDLI's ATF, ORAEC, and the sweeps that only MOVE stored text).
- **`page-counter-writers` caught both new CDLI scripts** writing `pages_ocr`/`pages_translated` by hand — the #4499 ratchet doing its job. Both now derive from `buildVisiblePageCountPipeline`, and production was re-derived the same way (the four books still claimed 1–2 translated pages after the fabrications were withdrawn; now 1/0, 1/0, 0/0, 0/0).

JDerekLomas:
**#4765 added** — the translation half, which is the one that reaches the reader.

A looping source is now refused **pre-flight**, so the call is never billed: `isDegenerateSource()` + an `'ocr-loop'` reason on `isTranslatablePage()` (THE translatability check, so audits and eval samplers inherit it), the worker stamping `translation.health_blocked: 'source_loop'` (the existing field, already excluded by the selection query — no new field), and the same refusal in `realtime-translate`, `retranslate-stale`, `retranslate-pages` and both `batch-translate-async` routes.

Of the 69,223 loop pages in the corpus, 53,628 are already translated — a withholding question, not this PR's — and **15,595 across 1,840 books are not yet**. Those are the ones this stops from becoming fabricated prose.

`tests/unit/translate-loop-source.test.ts` pins the refusal and the controls it must not touch (litany, spaceless script, ordinary prose). Full unit suite green locally: 299 files, 3,963 tests.

PR #4888, opened 2026-09-16
dehyphenate: the NOT SIGN (¬) line-end marker is a hyphen too (#4780)

JDerekLomas:
From the #4780 day-3 spot check: [The Voice of Africa p491](https://sourcelibrary.org/book/6a09e9bd3ca4edcd1ef0189d?page=491) reads `non¬\ncommissioned` — on some Archive items ABBYY writes its soft-hyphen marker as U+00AC NOT SIGN, which `dehyphenateLineBreaks` did not treat as a hyphen, so those words stayed split after the backfill. One character added to the class; the lowercase-continuation rule is unchanged; test added. Prevalence measured and posted on #4780; if material, `dehyphenate-ia-ocr.mjs` re-run (idempotent) joins them.


JDerekLomas:
Prevalence measured from `book_events` (Hetzner): **33 books, 3,245 pages** of the 1,114 ingested books carry a `¬` line-end hyphen. After merge: `dehyphenate-ia-ocr.mjs --apply` (idempotent; only those pages change, revisions saved as before).

PR #4904, opened 2026-09-17
docs(claude): re-measure the corpus stat line — it was 24% low on the reader-facing count

JDerekLomas:
## Why

The `## Stack` line in `CLAUDE.md` carries the corpus figures every session quotes from. It was measured **2026-08-30** and presented as current. I quoted the live-book count off it earlier this session before checking — which is precisely the failure its own "re-measure before quoting" clause warns about.

Re-measured today against `bookstore`. Every figure had moved, and the one most likely to be repeated out loud had moved most:

| figure | line said | actual (2026-09-17) |
|---|---|---|
| **books you can read** (`visible && pages_count > 0`) | 31,731 | **41,919** (+32%) |
| total docs | 109,567 | 114,695 |
| `visible: true` | 47,483 | 57,670 |
| artwork records at `pages_count: 0` | 15,752 | 15,751 |
| `pages_count > 0` | 84,574 | 89,700 |
| `pages_ocr > 0` | 61,829 | 64,444 |
| pages | 20.2M / 6.45M / 5.05M | 21.6M / 6.77M / 5.07M |

18 days of drift put the reader-facing number **24% low**. The corpus grew; the doc didn't.

## What changed

One line. The numbers are refreshed and the "re-measure before quoting" warning now cites the *fresh* drift (three weeks, 24% low) alongside the older 5× miss — a recent miss argues better than an archived one, and it costs the same words.

**Budget-neutral by construction:** 128 words before, 128 after. `wc -w` unchanged at 5,595. No new rule, so nothing had to be demoted.

## Scope

- **In:** the stat line only.
- **Out:** the 95-word overage against the ~5,500 budget. It predates this change and closing it means demoting something — the best candidate is the R2 book-scoped-key bullet under Data Protection (it names its own subsystem, its own guards, and its own standing detector, which is the stated test for demotion). That is a judgement call about a CRITICAL section and belongs in a PR of its own, with your eyes on it, not in a wrap-up.
- Checked the twin `~/.claude/CLAUDE.md` — these figures appear only here, so nothing diverged.
- Swept every `*.md` pointer in `CLAUDE.md`: all resolve.


JDerekLomas:
Superseded by #4469 (815215d7), which demoted the corpus stat line off CLAUDE.md entirely to `.claude/docs/invariants/visibility-and-stats.md` to hold the word budget — closing rather than re-adding dated numbers this PR's diff would reintroduce.

Issue #4883, opened 2026-09-15
Withdraw Syriac (and, until the NDL lane runs, kuzushiji) translations from "readable": nothing reads the script — takedown-surface list (#4746, #4745)


JDerekLomas:
## Why

The specialist OCR benchmark (#4746, registry #4735) found that **nothing we run reads Syriac**: on 20 sealed pages the two Gemini arms disagree with each other on 16 (distance > 0.5), Surya 2 emits placeholders or switches to Arabic, Tesseract `syr` and Kraken Sophro Mhiro are in the #4746 table, and by eye the Gemini output on a Serto manuscript page is either **recitation** (lite: one line of John 6:32 not on the page) or a **loop** (flash-preview: 448 lines for a 35-line page). The failure is fluent Syriac, so a reader cannot tell it from a transcription. This reproduces the GlotOCR result (0 % Acc@5 for every VLM on Syriac) on our own images.

We currently serve translations built on that OCR. Measured on the local corpus mirror (2026-09-15):

| | |
|---|---|
| books with `language = Syriac` | 105 (84 visible) |
| pages | 50,304 |
| pages with a served translation | 41,444 |

The #4746 decision rule says: *nothing reads it → say so and stop serving the translations as readable.* This issue is the withdrawal. It is a **hold-list change** (public-facing, data-shaped, reversible only by design), so nothing here runs without Derek's sign-off.

## What "withdraw" means — not deletion

- Keep every page image, every OCR row and every translation row. Nothing is deleted (`preservation-policy.md`).
- Mark the affected books so that the reader shows the image with a "no reliable transcription yet" state instead of the OCR/translation, and so that no derived surface presents the text as a reading.
- Reversible by one flag flip when an engine passes the #4746 bar (≥ 0.6 alignment to a reference where lite < 0.3).

## Surfaces to sweep (the takedown lesson: a hidden text stops rendering; a sentence about it does not)

A withdrawal touches every surface the Kloss takedown touched (`~/.claude/.../lesson_kloss_takedown_hidden_reason_sweeps.md`, ten surfaces), restricted to *text* rather than the book:

1. **Reader** (`/book/[id]`): OCR and translation panels for these pages → replaced by the withdrawal notice; image stays.
[...]

Issue #4681, opened 2026-09-08
Translation runs realtime, not on the Batch API — ~$978 on the current backlog, and half off every page after


JDerekLomas:
## Summary

OCR runs on the Gemini **Batch API** (50% discount). Translation does not — it runs realtime, at full
price, for a structural reason that looks avoidable. The meter in `scripts/audit/scope-progress.mjs`
states it plainly:

```
Estimated cost to finish: ~$37.5  (OCR $0.00222/pg batch + translation $0.00241/pg realtime)
```

Translation costs *more per page than OCR* despite shorter prompts, purely because it forgoes the batch
discount.

**The prize:** 811,864 pages are OCR'd but untranslated across visible books (5,687,815 `pages_ocr` vs
4,875,951 `pages_translated`, measured 2026-09-08). At the measured realtime rate that backlog is
**~$1,957**; on batch it is **~$978**. A **~$978 one-time saving**, plus roughly half off every page
translated thereafter. For comparison, the entire #3391 corruption repair is $37.

## Why it isn't on batch today — the actual blocker

Not an oversight. `scripts/workers/translate-worker.mjs` builds a **sequential chain**: each call's
prompt contains the *previous call's output*.

```
line 629  // Get previous page's translation for context continuity
line 695  prevTranslation = result.text;
line 793  prevTranslation = translatedText;
line 801  prevTranslation = singleResult.text;
line 877  prevTranslation = singleResult.text;
```

and it is injected at lines 265–268 / 333–336 as
`**Previous page translation for continuity:**\n${prevTranslation.slice(0, 2000)}`.

The Batch API requires every request to be submitted up front, so **no request can contain a prior
request's output**. That is the whole blocker.

Note the naming trap that hides this: `--batch-size=8` in this worker means *8 pages bundled into one
realtime call*, not the Batch API. Anyone grepping for "batch" here concludes translation is already
batched. It is not.

## Proposed fix — swap the continuity anchor for one that is knowable in advance

The continuity block exists so the model keeps terminology and register consistent across a page break.
It currently anchors on the previous page's **translation** (only knowable after that call). Anchor it
instead on the previous page's **OCR** — the source text — which is fully present before any
translation runs. Every page then becomes independent, the whole book submits as one batch job, and the
discount applies.

Something is genuinely lost: "how did I render this term last page" becomes "here is what the previous
page said", so cross-page *term* consistency weakens even though *contextual* continuity is preserved.
[...]

PR #4912, opened 2026-09-17
eval(batch-continuity): does translation survive the Batch API? Result: the seed matters; a seam-repair pass ties production (#4681)

JDerekLomas:
## What this is

The harness, the pinned sample, the paid run and the result for the preregistered question in #4905: **does translation lose anything at the block boundary when it moves to the Batch API?** Issue: #4681.

**Stacked on #4905.** This branch starts from the preregistration commit so that it precedes every paid call. Merge #4905 first and this diff shrinks to the harness and results; or merge this alone and #4905 becomes redundant.

## Result

By the decision rule as written, **rung 5 fired: nothing passes, do not migrate.** That is reported as written, and it is not the whole story, because one leg of the rule turned out to be unreachable.

| arm | blind judge prefers A / arm / tie | A's share (limit 60%) | terminology kept, seam page |
|---|---|---|---|
| A production (chained) | — | — | 72% |
| B plain batch (no seed) | 41 / 11 / 5 | **76.3% fail** | 40% |
| C seeded with source text | 32 / 16 / 9 | **64.0% fail** | 48% |
| D one-page overlap | 30 / 21 / 6 | 57.9% pass | 32% |
| E second pass repairing the seam page | 27 / 27 / 3 | **50.0% pass** | 72% |
| *A run a second time (noise floor)* | not judged | — | 80% |

- **The continuity seed is not buying nothing.** A blind judge prefers production over plain batch 41 to 11 at the seam, with concrete reasons: a dangling clause picked up or dropped, "Allah" kept versus switched to "God".
- **H1's whole-block bound cannot be met at this sample size.** Only 17 of 57 boundaries carry an eligible term, and production run twice against itself fails the bound by more than any arm did. So every arm "fails H1", including E.
- **On the two legs that can discriminate, E and D pass and B and C fail.** E ties production, never sees pages outside the seam, left the seam page a median 96% similar to B's, and cost 15% on top of B.
- Seeing the previous page's *source* (C, D) does not carry renderings across. Seeing its *translation* (A, E) does. That is why D measured below E although it was ranked above it beforehand.

The migration call is Derek's. The evidence points at batch plus a seam-repair second stage, confirmed by a re-run that drops whole-block H1.

## Controls

- **The probe fires:** real cross-boundary consistency 73.8% against 3.8% on another book's terms.
- **The harness runs production's configuration:** arm A reproduces production's stored translations as closely as two harness runs reproduce each other.
- **Nothing was written to `pages`:** sha256 over all 928 sample page documents is identical before and after.
- **The draw is explained:** 209 candidate seams rejected, by reason, in the pinned sample; 40 of 58 seams end mid-sentence.
- Positive-control unit tests for the term scorer, the seam filter and the prompt builder.

## Spend

**$2.72** metered under `eval/translation-batch-continuity`, against a $3.00 estimate and a $5 ceiling. Run on Hetzner. It counted against the 2026-09-17 daily dial.

## In scope

- `scripts/eval/translation-batch-continuity-ab.mjs` and its tests
- `gemini` and `pool` exported from `translation-prompt-ab.mjs` so they are imported, not copied
- eight dated preregistration amendments, each committed before the output it governs
- results under `scripts/eval/results/`, the `EXPERIMENTS.md` entry, regenerated `INDEX.md`

## Out of scope, deliberately

- **No change to the translate worker, the batch collector or the spend detector.** This PR measures; it migrates nothing.
- **Building the two-stage batch lane for E.** That follows Derek's decision, and should be preceded by a confirmation run.
- **An H2 noise floor** (judging A against its own replicate). Worth adding to the confirmation run; not paid for here.
- Scoring against published translations: a different estimand, covered by #4883.


JDerekLomas:
**Re-run with the seam in view (2026-09-22, branch `eval/batch-continuity-tail`, $0.50).** #4968 found that arm A *and* arm E's repair prompt seeded the FIRST 2,000 chars of the previous page. Arms **At** and **Et** are A and E with the tail (editorial blocks stripped, `...` in front), prompts pinned to Standard Translation v13 / English Modernization v1 as in the original run; prev and B reused. Blind Sonnet judge, same question, 57 junctions per pair.

| pair | first / second / tie | first's share (limit 60%) |
|---|---|---|
| **At / Et** (the gate) | 15 / 18 / 24 | **47.4%** — Et holds |
| A / At (does the tail fix help?) | 24 / 13 / 20 | 59.6% — old A preferred |

Splits: only **15 of 58** previous pages exceed 2,000 stripped chars in this sample (median 1,659), so for 43 seams At differs from A only by the stripped `<summary>/<keywords>`, the moved `...` and the label — and A still leads there 17/9/16; among the 15 where the tail really differed, 7/4/4. Mid-sentence seams (40): A/At 19/11/10; At/Et 10/12/18. H1 seam-page terminology: A 72%, At 68%, E 72%, Et 72% (n=25, within the A-vs-A2 noise).

Reading of the A-wins: mostly format conventions the head of a page carries (running header, page-number placement, quote style, capitalisation) and name/term consistency — plausibly the `<summary>` I strip was doing work. So: the migration case (batch + seam repair) survives a fixed baseline; the tail-only seed as merged in #4970 is not vindicated and needs a hybrid arm (summary kept + head + tail) before it is trusted. Verdicts, key and packets are on the branch.

JDerekLomas:
Superseded by #4973, which contains every commit here plus the tail/hybrid arms that #4968 made necessary. #4973 is rebased on main (2026-09-22) and carries the resolved EXPERIMENTS/INDEX/PREREGISTRATION merges.

PR #4919, opened 2026-09-18
feat(ops): GPU lease watchdog on Hetzner — expired Scaleway leases are stopped via the provider API (#4909)

### LINKED ISSUE #4909: Rented GPUs need a lease enforced from Hetzner: ≈€40 of a $52 benchmark was an idle L4 (#4735)
## Why
The specialist OCR benchmark (#4735, #4885) cost ≈ $52, of which ≈ $4 was model calls and ≈ €40 was a rented Scaleway L4 sitting idle: 43 h after a hung step (the owning background session had hit its usage limit and could not wake), then 5 h more because a guest-side `shutdown -h` leaves a Scaleway instance "stopped in place" and still billed. The guard was the session's own attention, and the session was the thing that failed. A rule in a doc did not prevent the second occurrence two days after the first.

## The fix: a lease, enforced from a host that stays up
1. **Every rented GPU instance carries a lease tag** at creation: `lease-until=<ISO time>` and `owner=<issue number>`. Extending the lease is an explicit act (`scw instance server update <id> tags.0=...`), so continuing to spend is a decision, never a default.
2. **A watchdog on Hetzner, every 10 min** (`scripts/maintenance/gpu-lease-watchdog.mjs`, cron): lists Scaleway instances of GPU types in all zones via the provider API.
   - running + lease expired → `server stop` through the **API** (archives the instance; volumes and data are kept, nothing is deleted), then re-reads state until it is `stopped`/`archived`, and sends a push notification.
   - running + no lease tag → notify on every pass (does not stop: another project's training run may be legitimate; the nag is the cost of not tagging).
   - running + GPU utilisation 0 % for 30 min (optional ssh probe) → notify.
   - stopped/archived instances with attached volumes older than 7 days → weekly line in the report with the monthly storage cost, so leftovers are deleted deliberately.
3. **One launcher** (`scripts/eval/lib/gpu-lease.mjs`): `rent({type, hours, issue})` creates the instance with the tags already set; eval scripts use it instead of raw `scw`. Unattended steps keep running under `timeout` as adopted in #4735.
4. The watchdog's own failure must look different from "nothing running": it writes a heartbeat row, and the daily health snapshot flags a heartbeat older than 30 min.

## Needs Derek (credential)
A Scaleway API key scoped to `InstancesFullAccess` only, stored on Hetzner by name. No other step needs him.

## State on 2026-09-17 04:55 UTC (checked with `scw instance server list zone=all`)
- `sl-ocr-gpu-test` (L4, fr-par-1): archived, not billing compute; 80 GB volume + IP still billed.
- `sl-mitra-1` (L4, fr-par-2): archived; 80 GB + 300 GB volumes + IP still billed.
- `earthai-gpu` (L40S, fr-par-2): RUNNING, GPU 100 %, a 6.5 h training job — in use, but it has no dead-man of any kind.

## Out of scope
Gemini/API spend (already gated by the dial and envelopes); deleting the leftover volumes (needs a look at what is on them first).

JDerekLomas:
## Why
Closes #4909. ≈ €40 of the $52 specialist OCR benchmark was a rented Scaleway L4 sitting idle: 43 h after a hung step whose owning session could not wake, then 5 h more because a guest-side `shutdown -h` leaves a Scaleway instance "stopped in place" and still billed. The guard was a session's attention, and a rule in a doc did not stop the second occurrence two days after the first. This puts the guard on a host that stays up.

## What
- `scripts/maintenance/gpu-lease-watchdog.mjs` — every 10 minutes on Hetzner:
  - lists instances in all nine Scaleway zones through the REST API (no `scw` on the box);
  - watches every GPU commercial type (L4, L40S, H100, A100, B200, GPU-*, RENDER-*) and anything carrying a lease tag;
  - **lease expired** (`lease-until=<ISO>` in the past) → `poweroff` via the API, which archives the instance and keeps its volumes; re-reads state until `stopped` (4 min cap); emails. An unconfirmed stop exits 2 and emails "STOP NOT CONFIRMED";
  - **no lease tag** → emails once per 6 h per server and never stops it (another project's training run is legitimate);
  - `--weekly` (Mondays 08:10) → volumes still billing on stopped/archived instances, with approximate €/month;
  - heartbeat file written only after the API answered; API failure exits 1 and emails (6 h cooldown);
  - `--lease <id> --zone <z> --hours <n> --owner <issue>` sets or extends a lease, so continuing to spend is an explicit act.
- `scripts/workers/crontab.production` — the two lines, re-snapshotted from the live crontab.

## Installed live (2026-09-18 07:05 UTC)
- Key: the existing `SCALEWAY_SECRET_KEY` from secret-lover, copied to `/root/.scaleway.env` (mode 600) and sourced by these two cron lines only — not added to `.env.production.local`, which every worker sources.
- Dry run on Hetzner: 5 instances, 3 watched (all stopped), heartbeat written. Dry run from the laptop with `--weekly`: three idle volumes reported (80 GB on `sl-ocr-gpu-test`, 80 GB on `sl-mitra-1`, 300 GB on `earthai-gpu`), ≈ €36/month between them.
- Not exercised live: the expired-lease stop path and the untagged nag — nothing is running right now and I did not rent a GPU to test a guard against renting GPUs. The first real lease will be the test; the confirm-or-alert shape means a failure of that path emails rather than stays silent.

## Out of scope
- A launcher library that creates instances with the tag already set (the issue's step 3) — until then, tag at creation with `scw ... tags.0=lease-until=... tags.1=owner=...` or run `--lease` right after boot.
- Wiring the heartbeat into `daily-health-snapshot` (issue step 4).
- A narrower Scaleway key: the copied key is the account's existing one; an `InstancesFullAccess`-only key would be strictly safer and is a one-line swap in `/root/.scaleway.env`.
- Deleting the three leftover volumes — the weekly report nudges; deletion stays a human act.


JDerekLomas:
Also in this PR (the gnite ratchet): one routing line in CLAUDE.md under *Writing a sweep, an import, or a new field* — renting a GPU → lease tag + this watchdog — paid for by trimming the dated 2026-05-26 archaeology from the Stack stats sentence. Word count 5,595 → 5,619 (budget ~5,500; it was already over before this PR).

PR #4929, opened 2026-09-18
Translation staleness is the timestamp rule, stored and indexed; withholding is per lane (#4927)

### LINKED ISSUE #4927: A translation should know which transcription it was made from — hash, flag, index, consumer
## A translation should know which transcription it was made from — hash, flag, index, consumer

Derek, 2026-09-18: "is it obvious across the database when the transcription is newer than the
translation (and therefore the translation needs updating)? can it be?" — measured answer: **no, and
the rule we have catches none of the live cases.** This issue is the fix.

### Measured 2026-09-18 (full scan, Hetzner)

| | |
|---|---|
| pages with a stored translation | **5,401,385** |
| of those, carrying BOTH `ocr.updated_at` and `translation.updated_at` | **5,358,661 (99.2%)** |
| pages where the transcription is strictly newer than the translation | **16,026** |
| …of which the current rule catches | **0** |
| pages carrying `ocr.pipeline` at all | 3,231 |

Gap distribution for those 16,026: 55 under a minute, 933 under an hour, 4,836 one hour to a day
(much of that is ordinary within-run ordering), **1,567 one day to a week, 3,730 one week to a month,
4,905 over a month.** So roughly **10,200 pages are serving English translated from text the page no
longer holds**, and the tail is a year deep.

**Why the rule misses them.** `staleTranslationReason()` (`scripts/lib/stale-translation.mjs`) only
evaluates the date comparison when `ocr.pipeline` is set — and the only script in the repo that sets
that field is `scripts/maintenance/apply-reocr-verdicts.mjs` (#4523). The ordinary OCR writers
(`batch-collector.mjs`, `realtime-ocr.mjs`, the Lambda path) never stamp it, so every re-read outside
that one lane is invisible to the detector. The rule worked for the lane it was written for; it was
never general.

**Why it cannot simply be queried.** `pages_ocr_updated_idx` and `pages_translation_updated_idx` both
exist, but MongoDB cannot use an index to compare two fields to each other — the `$expr` scan above
took long enough to blow a 30s socket timeout on the first attempt. Staleness is unanswerable at
speed today because no verdict is stored.

### The fix

**1. A translation records the transcription it was made from.** On every write of
`translation.data`, also write `translation.source_hash` — sha256 (hex, truncated to 16 chars) of the
exact `ocr.data` string handed to the model — plus `translation.source_updated_at`. A page is stale
**iff** `source_hash !== hash(ocr.data)`. Exact, not inferred: immune to clock ordering, to idempotent
re-runs that rewrite identical text, and to a translation and transcription written seconds apart in
the same pass.

Writers to cover (find them all; this list is a start, not a guarantee):
`scripts/workers/translate-worker.mjs` (including the blank-page, recitation-blocked and
safety-blocked branches), `scripts/batch/realtime-translate.mjs`, `scripts/workers/batch-collector.mjs`,
`scripts/workers/lib/supabase-page-writer.mjs`, `src/app/api/[tenant]/pages/[id]/route.ts`.
Note `page_translations` (Supabase) holds the only copy of translation text — this is additive
metadata on the Mongo page; **nothing here deletes translation text**.

**2. Keep the timestamp comparison as the fallback, minus the gate.** For pages predating the hash,
drop the `ocr.pipeline` precondition and compare dates with a small margin (~60s) so that same-run
ordering is not reported as staleness. Missing translation date still counts as old, not fresh — that
existing choice is right and the reasoning in the module should stay.

**3. Materialise the verdict so it is queryable.** A partial-indexed marker (`translation_stale` with
a reason, or a `needs_retranslate` marker) written by the OCR writers at the moment they change text
under an existing translation, and by a sweep that re-derives the set for anything missed. **Do not
build a frozen id list** — the Kloss takedown leaked for six weeks that way. Self-healing: when the
page is retranslated the hash matches and the marker clears with nothing to remember.

**4. A consumer, which is the piece that does not exist at all.** Flagged pages need a re-translate
queue that actually drains, using the production translation path and the current routing policy.
Without it this becomes the fourth queue in the repo that nothing reads (#3958, #3635, and the review
queues in #4920 are the others).

### Guards

- A blank page carries a `[Blank page]` placeholder in `translation.data` and is not a translation —
  do not sweep those into a paid re-translate loop (`lesson-blank-page-accounting`). Same for the
  recitation- and safety-blocked placeholders.
- **Writing a marker that a worker consumes is actuation, not recording** — say in the PR what reads
  the flag and when it next runs.
- Re-translation costs money: report the page count and a price before anything drains, and ask.
- Verify by OUTPUT: `modifiedCount` on the backfill, and the stale count falling after the consumer
  runs — not the absence of errors.

### Why now

The Syriac Kraken lane (#4883) is about to re-transcribe up to 42,794 pages. Every one of them
orphans its English translation, and as things stand **nothing would notice**.

JDerekLomas:
Closes #4927. Reworked 2026-09-19 to the re-scope on #4927 and the hold comment below: **no content hash**.

Derek, 2026-09-18: *"I just don't see a situation where the timestamp wouldn't be enough… and if it doesn't leave one, that's a bigger issue."*

## What this makes true

**Staleness is the timestamp rule, stored and indexed; withholding is a per-lane opt-in that shares no predicate with it.**

- `translationStaleness(page)` (`scripts/lib/stale-translation.mjs`): a page's translation is stale when `ocr.updated_at` is newer than `translation.updated_at` by more than `STALE_MARGIN_MS`, or the translation has no date at all (missing counts as old). **The `ocr.pipeline` gate is gone from the rule.** A placeholder (`[Blank page]`, recitation, safety) is never stale, so nothing sweeps it into a paid loop.
- `translation_stale: { reason, since, lane? }` is the materialised verdict, behind the partial index `pages_translation_stale_partial` (`translation_stale.reason` + `book_id`). Written by the daily sweep `scripts/maintenance/mark-stale-translations.mjs` (crontab 07:30, re-derived every run — no frozen list) and, if a re-OCR lane wants its pages found at once, by that lane (the Syriac Kraken lane in #4928 already stamps it). **Cleared by every translation writer** with `$unset` (`CLEAR_STALE_UNSET`, 24 update statements across scripts + app, placeholder branches included): a new translation is the exit.
- **The clock is guarded at the write boundary, not tolerated downstream:** `tests/unit/ocr-write-stamps-updated-at.test.ts` sweeps every writer of `pages.ocr.data` and fails when a write does not stamp `ocr.updated_at` in the same update. It found one live writer that did not — `src/app/api/contribute/process/route.ts` wrote a contributor's reading with `ocr.processed_at` only — fixed in this PR. The probe has a positive control and asserts it still matches `batch-collector.mjs` and `write-processor-logic.ts`.
- **Consumer:** `scripts/batch/realtime-translate.mjs --stale` — books from the partial index, held books excluded (`NOT_HELD`), every page re-verified by the rule before a paid call, a page that healed in the meantime has its marker cleared instead of being billed, spend-guard consulted (`budgetAllowsDispatch`). Not scheduled; a priced dry run by default. `translate-worker.mjs` page selection also matches the marker for books it is already dispatching (it already matched `translation.updated_at < ocr.updated_at`, so this changes nothing about what it spends).
- Health snapshot line: "Stale translations: N" — a number that should fall as the consumer drains it.

## Which sweep withholds, what it keys on now, and what it would do next run

**Only `scripts/maintenance/withhold-stale-translations.mjs` withholds** (hourly, Hetzner crontab `20 * * * *`). Flagging (`translation_stale`) never takes text off a page. The two verbs have two predicates:

| verb | predicate | scope |
|---|---|---|
| flag for re-translation | `translationStaleness()` | whole corpus |
| withhold | `staleTranslationReason()` = stale **and** `ocr.pipeline ∈ WITHHOLD_LANES`, or `ocr.unreadable`, or a looping source | `WITHHOLD_LANES = ['reocr_bdrc_4523']` |

The lane check lives **in the verdict itself**, not only in the candidate query, so `--loop-arm`, the restore script and book-scoped runs cannot widen it (pinned by test). Before this PR the hourly sweep keyed on `ocr.pipeline` *existing*, which would have withheld the Syriac lane page by page once #4928 stamped that field — Derek, same day: *"don't withhold individual pages though."* The Syriac lane (`syriac-kraken-2026-09`) is deliberately not in the list; its pages get flagged and re-translated, never hidden.

**Measured on production 2026-09-19 (read-only, Hetzner, `/root/stale-bands-4927.json`):**

| | |
|---|---|
| pages carrying `ocr.pipeline` | 69,209 `reocr_bdrc_4523` + 3 `syriac-kraken-2026-09` (probe pages, no translation) |
| BDRC pages still holding a translation | 3,231 — all retranslated after the re-OCR, **0 stale** |
| `ocr.unreadable` pages holding a translation | 0 |
| **withhold sweep, next run** | **0 pages** |
| pages already withheld (`translation_withheld`) | 140,892 |
| `translation_stale` markers today | 0 (nothing from the earlier PR version ever landed) |

## The margin, from reading pages rather than guessing

15,951 pages have an OCR clock newer than their translation's (16,026 on 09-18). Gap distribution and what each band turned out to be, from `/root/stale-samples-4927.txt` (20 read per band):

| gap | pages | what the sample shows |
|---|---|---|
| < 1 s | 20 | one collector tick writing a translation job's result and an OCR job's result for the same book 96 ms apart (a Japanese mushroom manual, 2026-07-10); every translation fits its page. **The only band where both fields come from one writer pass.** |
| 1 s – 1 min | 27 | a realtime read translated, then a batch read of the same pages collected 3–51 s later — the same race as the next band |
| 1 – 60 min | 937 | the April–May pipeline race: realtime translation at :38–:46, batch OCR landing at :54. 19/20 translations still fit the current transcription; **1/20 does not** (a Latin manuscript re-read by full flash 22 min after a lite translation: a 633-char reading now sits under a 5,261-char translation) |
| 1 – 24 h | 4,830 | 12/20 are the IA-OCR ingest (#4780) replacing a Gemini read of an English printed book with IA's own text ~11 h after a normalising "translation" — the English still matches the page, and in one case is *better* than the new OCR ("LIFE OF JOHN JAY" vs IA's "JOHN J.VY"); 2/20 are genuinely stale (a Yoruba hymnal where a 134-char heading-only reading now sits under a full hymn translation); the rest are full-model re-reads 7–12 h after a lite translation, translations fit |
| 1 d – 1 w | 1,560 | re-OCR lanes |
| 1 w – 30 d | 3,730 | re-OCR lanes |
| > 30 d | 4,847 | March full-flash re-reads, July batch re-OCR; most translations still fit the page's content, some do not (a page now read as *blank* under a chapter continuation; a 776-char Chinese reading under a 5,960-char translation) |

What that says: from three seconds upward, every sampled page really did have its transcription **replaced by a different read** after the translation was made. The translation usually still fits, sometimes not, and a clock cannot rank that — so the margin does not try. It is a **write-ordering tolerance**: `STALE_MARGIN_MS = 60_000` covers the one-tick case plus clock skew between the hosts that write these fields (Lambda, Hetzner, Vercel). **It excludes 47 of 15,951 pages (0.3%).** The test pins the constant's edge, not its value.

Two findings for whoever drains the set, stated here rather than built: (1) 2,630 of the sub-day pages are the IA-ingest cohort (`ocr.source: ia_djvu`) on English books, the lowest-value re-translation in the set — the consumer's dry run prints per-book counts, and the operator can take the oldest bands first; (2) the pipeline race that produced the 1–60 min band (batch OCR landing after a realtime translation) is mostly April–May 2026 — 751 of its 937 pages carry the retired `-preview` model tag — with a tail into September.

## ACTUATION (CLAUDE.md: writing to a store an automated job reads)

- The daily sweep (07:30) writes `translation_stale`. Readers: `translate-worker.mjs` page selection (books it is already dispatching; no change in spend, see above) and `realtime-translate.mjs --stale` (hand-run, priced dry run by default, spend-guard). Nothing in this PR spends money on its own.
- **The first `--full` flagging pass has NOT been run.** It is an operator step on Hetzner after merge: `node --env-file=.env.production.local scripts/maintenance/mark-stale-translations.mjs --full` (dry run; add `--apply` to write). Expected to mark ≈15,904 dated pages plus the undated cohort (55 real translations with no date under a dated OCR). Report `modifiedCount`.
- The sweep never bumps `pages.updated_at` (`embed-gemini --incremental` selects on it).
- Nothing here deletes translation text; `page_translations` is untouched.

## Tests

- `stale-translation.test.ts`: translated-then-re-OCR'd is stale; within-margin is fresh, one ms past it is stale; missing date = old; placeholders never stale; **the rule is not gated on `ocr.pipeline`; withholding is** (a stale page with no lane, or the Syriac lane, returns `null` from the withhold verdict).
- `ocr-write-stamps-updated-at.test.ts`: the write-boundary sweep, with positive control.
- `npx tsc --noEmit` clean; 4,024 unit tests green.

## Out of scope, deliberately

- `retranslate-stale.mjs` (model-vintage staleness, Batch API) is untouched.
- A `--ocr-source` filter on the consumer (to skip the IA-ingest cohort) — one-liner, wanted only once the operator has seen the priced dry run.
- The Syriac lane's own `translation_stale` stamp (#4928) still filters on `translation.source_hash` when present; after this merges that field never exists, so the filter is inert and can be dropped there.

## Coordination

#4928 holds its apply pass (`NO_APPLY`) until this merges and the hourly auto-pull lands the pinned `WITHHOLD_LANES` filter on Hetzner.


JDerekLomas:
**Hold — this PR predates Derek's decision and needs two changes before it can be merged.**

**1. Drop the content hash.** Derek, 2026-09-18: *"I just don't see a situation where the timestamp
wouldn't be enough… and if it doesn't leave one, that's a bigger issue."* Verified before conceding:
every live writer of `ocr.data` already stamps `ocr.updated_at` (the ones a dotted-key grep misses
assign the whole `ocr` object with `updated_at` inside it). A writer that changed text without
stamping would be **the bug to fix at the write boundary** — the same lesson as the R2 book-scoped
keys — not a reason to run a parallel mechanism that tolerates it. This PR still carries 46
occurrences of `source_hash`, `scripts/lib/translation-source.mjs` + its `src/lib` twin, a parity test
and `backfill-translation-source-hash.mjs`. Those come out, along with the hash writes threaded
through ~20 write sites.

**What stays** (the re-scope comment on #4927 is the spec): the `ocr.pipeline` gate removed from
`staleTranslationReason()`; the margin **set by reading sample pages, not by guessing a number**, with
the PR saying what the sample showed and how many pages the margin excludes; a test asserting that any
update touching `ocr.data` also sets `ocr.updated_at`; the stored, indexed verdict; and
`retranslate-stale.mjs`, the consumer that has never existed.

**2. A conflict with an explicit instruction, and it is the urgent half.** The Syriac lane session
reports that the **hourly withhold sweep keys on `ocr.pipeline` existing**, so once the Kraken lane
(#4928) stamps that field it would withhold the stale English **page by page**. Derek, same day:
*"don't withhold individual pages though."* In this design the answer to a stale translation is
**flag and re-translate — never hide**. So:

- removing the gate must not widen what the **withhold** sweep acts on. Flagging and withholding are
  different verbs and must not share a predicate;
- the PR must state plainly which sweep withholds, what it keys on after this change, and how many
  pages it would touch on its next run;
- if the hourly sweep would withhold Syriac pages once #4928's apply pass runs, that is a **blocker
  for the Syriac lane**, not a footnote.

#4928 is holding its apply pass (`NO_APPLY`) until this merges, so this PR is on the critical path.

JDerekLomas:
**Reworked to the hold comment and the #4927 re-scope — both points addressed, `blocked` removed.**

1. **Hash dropped.** `translation-source.mjs`, the `src/lib` twin, the parity test, the backfill and every `source_hash` / `page_source_hashes` / `markStaleAfterOcrWrite` write are gone (0 occurrences left). The rule is the timestamps, ungated, in `stale-translation.mjs`; the margin was set by reading 20 pages per gap band (table in the body) and excludes 47 of 15,951 pages; the write-boundary sweeping test is in and **found one live writer that did not stamp `ocr.updated_at`** (the contributor route) — fixed here.

2. **Withholding cannot widen.** Only the hourly `withhold-stale-translations.mjs` withholds; its verdict now requires `ocr.pipeline ∈ WITHHOLD_LANES` (`['reocr_bdrc_4523']`) *in the verdict itself*, not just the candidate query, so `--loop-arm` / restore / book-scoped runs can't reach further either. Measured on prod today: **next run would withhold 0 pages** (3,231 BDRC pages still carry a translation, all retranslated after the re-OCR; 0 unreadable-with-translation). The Syriac lane stays out of the list — flagged and re-translated, never hidden — so #4928's apply pass is safe once this lands on Hetzner.

Not run: the first `--full` flagging pass (operator step after merge, command in the body). CI green; `tsc` clean; 4,024 unit tests pass.

